PR #28, opened 2017-07-27
Proof of concept: Stories

e5150:
I threw together a method for getting stories through the API. This only fetches the stories of followed accounts that have not been marked as seen, to get all stories of a user https://i.instagram.com/api/v1/feed/user/<user>/reel_media/ needs to be queried. I guess it doesn't really conform to the structure of the existing code, but might serve as a starting point if story downloading is a feature of interest…

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ python:
 install:
  - pip install pylint requests
 script:
- - python3 -m pylint -r n -d bad-whitespace,missing-docstring,too-many-arguments,locally-disabled,line-too-long,too-many-public-methods,too-many-lines,too-many-instance-attributes instaloader
+ - python3 -m pylint -r n -d bad-whitespace,missing-docstring,too-many-arguments,locally-disabled,line-too-long,too-many-public-methods,too-many-lines,too-many-instance-attributes,too-many-locals,too-many-branches,too-many-statements instaloader
--- a/README.rst
+++ b/README.rst
@@ -247,6 +247,12 @@ To download the last 20 pictures with hashtag #cat, do
 
     loader.download_hashtag('cat', max_count=20)
 
+If logged in, Instaloader is also able to download user stories:
+
+.. code:: python
+
+    loader.download_stories()
+
 Each Instagram profile has its own unique ID which stays unmodified even
 if a user changes his/her username. To get said ID, given the profile's
 name, you may call
--- a/instaloader.py
+++ b/instaloader.py
@@ -627,10 +627,12 @@ class Instaloader:
         return downloaded
 
     def download_stories(self,
+                         userids: List[int] = [],
                          download_videos: bool = True,
                          fast_update: bool = False) -> None:
         """
-        Download 'unseen' stories from user followees. Does not mark stories as seen.
+        Download 'unseen' stories from user followees. Or all stories of users whose ID are give.
+        Does not mark stories as seen.
 
         :param fast_update: If true, abort when first already-downloaded picture is encountered
         :param download_videos: True, if videos should be downloaded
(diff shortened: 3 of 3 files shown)


Thammus:
Thanks for this great enhancement 😃! We merged your commit into our v2-dev branch. It actually conformed well to the structure of the existing code. We made only minor adaptions in commit 7cac6d5.

e5150:
Great, I suppose the dev branch is appropriate, there probably needs to be a fair amount of testing to figure out all the potential variations in the json responses.
In commit [`14b02fa`](https://github.com/Thammus/instaloader/pull/28/commits/14b02fada1cf8f774eb0b3d53d4252f249570407) I've added support for downloading all active stories from specified user. Maybe this should be refactored in another way, possibly hocked into `download_profiles` / `download_feed`.

aandergr:
Thanks for adding the functionality of downloading user stories. I think this is a long-awaited feature and a major enhancement of Instaloader.

Besides c355338, which already has been merged into our v2-dev branch, this Pull Request currently adds the commits
- 14b02fa, adding a `userids` parameter to the new `download_stories()` method, which, when the list is not empty, selects given profiles to download stories from, instead of downloding from all profiles. Also, it does a fix regarding caption saving (can user stories have captions?), and regarding sleeping after not-downloading a story video.
- 9a274d2, allowing to mix id numbers and profile names in the `userids` parameter.

The remaining commits, f131847, 0ade7ef and c910bfc, do not add anything and thus should not be merged into our main development branch. So for incubating, I would recommend cherrypicking aforementioned commits.

I made a few tests and it seems to work fine.

The code is of overall high quality. However, I propose the following changes:
- `userids` having `[]` as a default value (_mutable_) might very likely lead to bugs or unexpected behavior. `userids` should be of type `Optional[List[Union[int,str]]]` with `None` as default value. Consequently, rather than testing whether it is empty, it should be tested whether it is `None`. Also, the parameter should be mentioned in the method's docstring.
- I somehow feel like it should be the callers job to lookup username ids, just like in the other Instaloader's method (at least I think so). So I am uncertain of whether we should merge 9a274d2.
- In `_get()`, the server's response is dumped to console. This probably remains from debugging and should be removed.
- In `_user_stories()`, `get_id_by_username()` is called to convert the profile name string to a user id. This must be `self.get_id_by_username()`.
- pylint considers it a bad practice to use `id` as variable name. Another name should be choosen, or pylint's warning should be deactivated for this case (`pylint:disable=redefined-builtin,invalid-name`). Further, it complains about using `type()` for a typecheck.

Further, I propose adding a target for the command line similar to `:feed-all` and `:feed-liked` and mentioning it in `README.rst` with an example, like
```
instaloader --login=USER --filename-pattern={date}_{profile} :stories
```
Maybe, we could also add a way to specify a list of users to download stories from.

e5150:
Yeah, sorry for the redundant commits, I was fumbling with git and github.
The issue of numeric id vs usernames was mostly for convenience in testing, since I didn't want to mess with the argument parsing or other methods. My idea was that there could be a command line switch to be able to get feed + stories, or particular users + their stories.  But I don't know what would fit best with the rest of the code. (Disregarding `9a274d2` also voids the issue of `type()`.)

And yes, stories can have a sort of caption, which tend to be stylishly printed onto the image itself. But a plain text version is included in the json, either as a simple string, or as a dict `{ pk, user_id, text, type, created_at, created_at_utc, content_type, status, bit_flags, user { cture, is_unpublished, is_favorite }, did_report_as_spam, media_id}` (maybe the same as `edge_media_to_caption`, which I've yet to stumble upon). I figured it might as well be saved, since it doesn't require additional requests. Similarly "free" location metadata gets stored, without consulting a `geotags` setting. Other metadata might also be of interest, such as `"story_hashtags"`.

PR #35, opened 2017-08-13
Simplify profpic regex.

e5150:
The original method of substituting 2048x2048 for whatever resolution was given seemed somewhat convoluted. This accomplishes the same thing, except raising an exception if the given url is not on the right domain.


aandergr:
Thank you!

PR #37, opened 2017-08-19
Use anonymous session for HEAD request.

e5150:
The default Instaloader headers aren't passed with a simple `requests.head()`, so leakage of user agents such as `"python-requests/2.18.1"` will occur.


aandergr:
Thanks for fixing this! :)

Issue #73, opened 2018-01-14
Another Profile's Interactions


anseml:
I have a feature suggestion that would helpful, I was wondering if it was possible to add the ability to save posts of a user you're following interacts with. I.e. posts they like or comment on.

In the Instagram app you can view these under the following section.

I was trying around the with  `--only-if ` and `viewer_has_liked` but didn't  find a way to have this work with a specified user other then it attempting to download images of theirs I had liked.  Would there be a way to implement specify saving posts and/or comments by a specific profile?

A possible extension is to output a list of followers and/or followees of a profile similar but not download the posts of them. 

A final idea is there a way to implement the downloading of posts a profile is tagged in (like  the photos of tab)?

PR #154, opened 2018-07-26
Add meth:get_tagged_posts to Profile

e5150:
Initial support for getting posts where a user has been tagged, which was recently added to the official website. (Suggested in #73.) 
Not sure how to best incorporate into the CLI, prepending "@" to the username would seem intuitive as a target, but that's taken. Maybe `--tagged` and `--tagged-only` like what is done for stories. And I don't know what default dirname makes the most sense, `tagged-user`, `post-owner` or possibly `@tagged-user`. With `@tagged-user` there will at least not be conflicts if one wants to `--fast-update --tagged-only username` and some post has been downloaded from `:feed` or such.

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -62,6 +62,7 @@ def _main(instaloader: Instaloader, targetlist: List[str],
           profile_pic: bool = True, profile_pic_only: bool = False,
           fast_update: bool = False,
           stories: bool = False, stories_only: bool = False,
+          tagged: bool = False, tagged_only: bool = False,
           post_filter_str: Optional[str] = None,
           storyitem_filter_str: Optional[str] = None) -> None:
     """Download set of profiles, hashtags etc. and handle logging in and session files if desired."""
@@ -146,7 +147,8 @@ def _main(instaloader: Instaloader, targetlist: List[str],
                 with instaloader.context.error_catcher(target):
                     try:
                         instaloader.download_profile(target, profile_pic, profile_pic_only,
-                                                     fast_update, post_filter=post_filter)
+                                                     fast_update, download_tagged=tagged,
+                                                     download_tagged_only=tagged_only, post_filter=post_filter)
                     except ProfileNotExistsException as err:
                         if instaloader.context.is_logged_in and not stories_only:
                             instaloader.context.log(err)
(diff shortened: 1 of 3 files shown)


Thammus:
Thank you for your code! We merged it.

e5150:
Actually, there is an issue with my code, `self` is passed as third argument to the `Post` constructor, which is inaccurate for posts where the owner is not the tagged user themself. This ought to be `self if int(node['owner']['id']) == self.userid else None`, sorry.

aandergr:
> Fix owner_profile for Profile.get_tagged_posts()

Thanks! I pushed it to our repository as b5aa268.

> Not sure how to best incorporate into the CLI

Not sure, either. Maybe `USER:tagged` as target name. So we could check whether target.endswith(':tagged'); and Instaloader already does colons in dirnames (e.g. target `:saved`). A `--tagged`/`--tagged-only` flag would be an option, too, but if it downloads them with target `USER` the file's path would not indicate whether it was posted by the user or the user only has been tagged in. I tend to be against `@tagged-user` or similar, because target.startswith('@') is already taken for followees and I'm unsure whether we'd run into problems with Windows again when having `@` in path names.

Maybe we could go with `USER:tagged` as target name and add flags `--tagged`/`--tagged-only` that add/substitute each `USER` target with `USER:tagged`?

e5150:
[`700b3a8`](https://github.com/instaloader/instaloader/pull/154/commits/700b3a8d094f552caa638a9d91f9221392a8e3f0) adds support for `--tagged`/`--tagged-only`, which is passed along to `download_profile()` rather than manipulating the target list. I wanted to go via `download_profile` to not have duplicate code for the checks for id, private, blocked and such. Support for `USER:tagged` could easily be added to imply `download_tagged_only` for that profile.

aandergr:
Great, thanks! Also thanks for already having it merged with my recent change to `__main__.py`. The help texts for the flags should be copied into `docs/cli-options.rst`, but I can also do that (actually I feel it's my fault to still haven't found a proper solution for that).

I'm going to push these commits into our `v4.1-dev` branch which I'll soon start, since it touches existing code without fixing bugs.

aandergr:
Thanks, it's in `v4.1-dev` now. I squashed it together with the merge commits into one commit to linearize the tree, I hope that's okay.

aandergr:
I changed the tagged target from `user:tagged` to `user/:tagged`. Since `user:tagged` is not supported as a target specification for the CLI, one couldn't invoke Instaloader with `instaloader [flags] */` when there were `ùser:tagged` directories present. I hope this solution is ok.

PR #161, opened 2018-08-06
Allow reading arguments from file

AndCycle:
function referenced from instagram-scraper,
which provide a convenient way to hide sensitive information.

as instagram is always asking me password even if I provide the sessionfile.

https://github.com/rarcega/instagram-scraper/pull/50

still need to write some help info to explain this function.

### DIFF
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -203,11 +203,9 @@ Miscellaneous Options
 
         text file should separate arg with line break.
 
-        args.txt example
+        args.txt example::
 
-        ```
             --login MYUSENAME
             --password MYPASSWORD
             --fast-update
-         ```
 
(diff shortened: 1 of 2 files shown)


aandergr:
Thanks for your suggestion!

I think `+` is a sufficiently esoteric fromfile prefix char, it doesn't clash with any of our target prefixes, nor with what Instagram allows in usernames, nor with the characters that are reserved in the shell. Also we should mention that feature in the documentation, maybe the introduction section in `docs/cli-options.rst` would be a good place for that.

AndCycle:
@e5150 thanks for the head up, I forget to lookup to make sure it doesn't conflict with any current function

@aandergr thanks for the review, I think I will close this one as `@` is a conflict prefix char

aandergr:
I would like to have this change included in v4.1, but we cannot use it with `@` as fromfile prefix char. In your commit, the `@` would just need to be changed to `+`.

aandergr:
Thanks, it's in `v4.1-dev` now.

PR #174, opened 2018-09-12
Add proxy server support

modgahead:
- Add proxy argument to `instaloader.Instaloader` class
- Add proxy option to `__main__.py`

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -303,6 +303,10 @@ def main():
                               'there is not yet a valid session file.')
 
     g_how = parser.add_argument_group('How to Download')
+    g_how.add_argument('--proxy', metavar='PROXY-SERVER',
+                       help='Set an SSL Proxy for requests. Without this option, '
+                            'requests will be done from your IP address. '
+                            'Valid formats are: "https://IP_ADDRESS:PORT" or "IP_ADDRESS:PORT"')
     g_how.add_argument('--dirname-pattern',
                        help='Name of directory where to store posts. {profile} is replaced by the profile name, '
                             '{target} is replaced by the target you specified, i.e. either :feed, #hashtag or the '
@@ -371,7 +375,8 @@ def main():
                              post_metadata_txt_pattern=post_metadata_txt_pattern,
                              storyitem_metadata_txt_pattern=storyitem_metadata_txt_pattern,
                              graphql_rate_limit=args.graphql_rate_limit,
-                             max_connection_attempts=args.max_connection_attempts)
+                             max_connection_attempts=args.max_connection_attempts,
+                             proxy=args.proxy)
         _main(loader,
               args.profile,
               username=args.login.lower() if args.login is not None else None,
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -116,9 +116,18 @@ class Instaloader:
                  post_metadata_txt_pattern: str = None,
                  storyitem_metadata_txt_pattern: str = None,
                  graphql_rate_limit: Optional[int] = None,
-                 max_connection_attempts: int = 3):
+                 max_connection_attempts: int = 3,
+                 proxy: Optional[str] = None,
+                 ):
 
-        self.context = InstaloaderContext(sleep, quiet, user_agent, graphql_rate_limit, max_connection_attempts)
+        self.context = InstaloaderContext(
+            sleep,
+            quiet,
+            user_agent,
+            graphql_rate_limit,
+            max_connection_attempts,
+            proxy,
+        )
 
         # configuration parameters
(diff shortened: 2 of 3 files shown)


ilmetu:
why not insert the possibility ti use also username and password proxy authentication ?

thanks

gnu-like:
if you will add proxy please atention sock and http such as tor

instaloader/__main__.py
```diff
@@ -303,6 +303,10 @@ def main():
                               'there is not yet a valid session file.')
 
     g_how = parser.add_argument_group('How to Download')
+    g_how.add_argument('--proxy', metavar='PROXY-SERVER',
```
aandergr: Also add description to `docs/cli-options.rst`

instaloader/instaloader.py
```diff
@@ -116,9 +116,18 @@ def __init__(self,
                  post_metadata_txt_pattern: str = None,
                  storyitem_metadata_txt_pattern: str = None,
                  graphql_rate_limit: Optional[int] = None,
-                 max_connection_attempts: int = 3):
-
-        self.context = InstaloaderContext(sleep, quiet, user_agent, graphql_rate_limit, max_connection_attempts)
+                 max_connection_attempts: int = 3,
+                 proxy: Optional[str] = None,
+                 ):
```
aandergr: The code formatting is not consistent with the rest of Instaoader (especially lines that only contain `)` or `):`, and lines ending with `(`)

(affects several changes in this pull request)

instaloader/instaloadercontext.py
```diff
@@ -47,15 +47,30 @@ class InstaloaderContext:
...
+            sleep: bool = True,
+            quiet: bool = False,
+            user_agent: Optional[str] = None,
+            graphql_count_per_slidingwindow: Optional[int] = None,
+            max_connection_attempts: int = 3,
+            proxy: Optional[str] = None,
+    ):
 
         self.user_agent = user_agent if user_agent is not None else default_user_agent()
         self._session = self.get_anonymous_session()
```
aandergr: Did you consider setting the `proxies` dictionary in the requests.Session object, rather than passing it to each Session.get call? I think it might be the better option since there are fewer placer where sessions are created than where they are used to make a request.

http://docs.python-requests.org/en/master/api/#requests.Session.proxies

instaloader/instaloadercontext.py
```diff
@@ -47,15 +47,30 @@ class InstaloaderContext:
...
         self.username = None
         self.sleep = sleep
         self.quiet = quiet
         self.max_connection_attempts = max_connection_attempts
+        if proxy:
+            if 'https' not in proxy:
```
aandergr:
```suggestion
            if not proxy.startswith('https'):
```

because `not in` searchs anywhere in the string

aandergr:
I close this pull request due to its long inactivity despite its pending issues. Please reopen when there are any updates.

nOOBIE-nOOBIE:
It would be awesome if proxy support is added on the next release.

jhjjj12:
Please can anyone help me with adding a proxy to my program! I've been looking everywhere to find help but can't find anything! Please respond!

amitupreti:
@jhjjj12  if you want to use proxy with instaloader. You can set the proxy in the environment variable(`https_proxy`).

Instaloader uses request library under the hood and to set up proxy with instaloader. you can simply do something like this

```py
import os
proxy = 'http://username:password@host:port'
# proxy = 'host:port'
os.environ['https_proxy'] = proxy
```

Now, instaloader will fetch data using that proxy.

I have been using this method and switch proxy after 50 requests and have never got 429 errors.

source: https://stackoverflow.com/questions/8287628/proxies-with-python-requests-module

jhjjj12:
> @jhjjj12 if you want to use proxy with instaloader. You can set the proxy in the environment variable(`https_proxy`).
> 
> Instaloader uses request library under the hood and to set up proxy with instaloader. you can simply do something like this
> 
> ```python
> import os
> 
> def switch_proxy(proxy):
>     os.environ['https_proxy'] = proxy
> 
> 
> # your instaloader code
> # whenever you need to switch proxy
> proxy = 'http://username:password@host:port'
> # proxy = 'host:port'
> switch_proxy(proxy)
> # your instaloader code cotinued
> ```
> 
> Now, instaloader will fetch data using that proxy.
> 
> I have been using this method and switch proxy after 50 requests and have never got 429 errors.
> 
> source: https://stackoverflow.com/questions/8287628/proxies-with-python-requests-module

youre a god sir thank you!

thebeautyofcoding:
This doesn not seem to work for me , still getting:
`    raise LoginRequiredException("Redirected to login page. Use --login.")
instaloader.exceptions.LoginRequiredException: Redirected to login page. Use --login.`

PR #245, opened 2019-01-28
Add Story Folder command line option

oneup7:
Added Story Folder command line option


aandergr:
I like the idea of giving the user the choice to have stories saved in another directory than regular posts for a profile without having to use separate Instaloader invocations. However, your implementation also creates a 'stories' folder when downloading with `:stories` and when downloading highlights, and adds a special case to the file naming logic which is handled with dirname-pattern and filename-pattern. We could give the user the option to store posts and stories in different directories by splitting the filename-pattern into a post-pattern and storyitem-pattern, which would be a more elegant and flexible solution.

Issue #369, opened 2019-08-03
How can I speed this up ?


ishandutta2007:
**Provide us a use case of the feature**
I am trying to download last few days post of a user, so my call is like this:

```
instaloader --post-filter="date_utc >= datetime(2019, 7, 1)" profileX
```
The thing is for an account having 10k-20k posts it takes hours to download a couple of pics.

**Describe the solution you'd like**
I was expecting to have something like this:

```
instaloader --count=10 profileX
```
or
```
instaloader --post-filter="date_utc >= datetime(2019, 7, 1)" --count=10 profileX
```

So that it can exit before having to check all timestamps.
Why is `count` only available for #hashtag and :feed ? Why  isn't ` count ` available for other usecases ?

ishandutta2007:
I have sent a PR as per above, please review and let me know https://github.com/instaloader/instaloader/pull/370

ishandutta2007:
@aandergr It has been more than a month since I posted this PR, what's the procedure to review ?

aandergr:
(copied from https://github.com/instaloader/instaloader/pull/370#issuecomment-541441960)

> [...], this would unfortunately change Instaloader's behavior in a backward-incompatible manner, since an invocation with `--count` and a mix of hashtags/feed/location and a profile would now behave differently in that the max count now also affects the profile and not only hashtags/feed/location. I hate that we currently have this inconsistency, but I fear that we confuse users when changing this behavior.
> 
> The next major version, Instaloader 5, would definitely have to have this included.
> 
> Also, a way we could "fix" this might be to support a per-target configuration, e.g. a YAML file where all the targets are defined with their individual options. There, a max-count option should be supported in all kinds of targets. This is something we could already do in Instaloader 4.

PR #370, opened 2019-08-04
Feature add max count for profile posts

ishandutta2007:
Modified as per this https://github.com/instaloader/instaloader/issues/369

### DIFF
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -171,7 +171,7 @@ Which Posts to Download
 .. option:: --count COUNT, -c
 
    Do not attempt to download more than COUNT posts.  Applies only to
-   ``#hashtag``, ``%location id``, and ``:feed``.
+   ``#hashtag``, ``%location id``, ``:feed`` and ``profile(s)``.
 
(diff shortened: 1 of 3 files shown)


aandergr:
Even tough we appreciate your effort and the PR looks nice and I like the idea, this would unfortunately change Instaloader's behavior in a backward-incompatible manner, since an invocation with `--count` and a mix of hashtags/feed/location and a profile would now behave differently in that the max count now also affects the profile and not only hashtags/feed/location. I hate that we currently have this inconsistency, but I fear that we confuse users when changing this behavior.

The next major version, Instaloader 5, would definitely have to have this included.

Also, a way we could "fix" this might be to support a per-target configuration, e.g. a YAML file where all the targets are defined with their individual options. There, a max-count option should be supported in all kinds of targets. This is something we could already do in Instaloader 4.

ishandutta2007:
Didn't get, how do you think it will affect other cases? Can you explain with an example?
I have made `max_count` optional such that none of the other use-cases get affected.

Issue #388, opened 2019-08-23
Incorrect ProfileNotExistsException when scan is rate limited: redirect to https://www.instagram.com/accounts/login/


corpone:
**Describe the bug**
IG now begins to redirect to login page when scan limit is exhausted.

This shows up as:

`HTTP redirect from https://www.instagram.com/<profile>/ to https://www.instagram.com/accounts/login/`

**To Reproduce**
Scan too fast.

**Instaloader version**
Latest.

**Additional context**

Solution:
1. Special case redirect to `https://www.instagram.com/accounts/login/` on scan in https://github.com/instaloader/instaloader/blob/a88316224bab83603e0652265ef1558f778905c8/instaloader/instaloadercontext.py#L359

goker67:
I didn't understand clearly that solution. When I attempt to scrape, I still take the HTTP redirect message. How can I overcome to it?

marty90:
+1

marty90:
it appears like there is a new rate limit in scraping `https://www.instagram.com/<profile>/` pages :(

hgkmaxymus:
> it appears like there is a new rate limit in scraping `https://www.instagram.com/<profile>/` pages :(

Yeah. When logged in, the rate is greater than anonymous.

Weirdly enough, I didn't scrap anything yesterday and today I still had the same issue. Trying to check some profiles while logged out redirects me to the login webpage (using Chrome Private Browsing).

I think Instagram is having some massive server issues and is limiting everything now. I hope it is temporary, but who knows...

psylenced:
I tried the pull request last night and doesn't seem to be working.

aandergr:
Thanks for reporting and for your pull request!

I could not reproduce this behavior, in my experiments I always got the regular 429 response when exceeding the rate limit. I merged your pull request anyway, since it looks very good and doesn't introduce any side effects. However, for now we defer releasing 4.2.10 just with this fix included, since we do not see enough evidence yet that this bug still exists.

marty90:
Thank You.
It would be interesting to understand some further details about the new policies of Instagram.

corpone:
@aandergr not ideal to share research on this in a public forum, but I have a good idea of the factors that cause it. 

While we're on this topic, one suggestion I'd make is to add variable or exponential backoff and acceleration to the rate limiter. It's not capturing a lot of variables that I'm aware of, so picking fixed numbers that work in two scenarios will not be optimal for many users.

[...]

PR #389, opened 2019-08-23
handle alternate rate limit exceeded behavior of redirecting to login

corpone:
Closes https://github.com/instaloader/instaloader/issues/388


corpone:
@aandergr or @Thammus please review when you have a chance.

PR #391, opened 2019-08-23
Basic batch script

cyb3rsalih:
My friend looks something to download instagram photos.
Alternatives have restrictions. I advice your app. It works 
awesome but my friend not good with console apps. I created
a basic interface for him :) I think this is usable for windows users like him. 
Also it can improvable... 

Have a good day!


aandergr:
There's no reason to include such a script in the Instaloader source tree.

PR #422, opened 2019-09-29
support passing timeout to requests

### LINKED ISSUE #330: Timeout for HTTP requests
Hi all,
thank you for maintaining `instaloader`! It is a really great tool!

I am downloading Instagram data using `instaloader` API, in particolare the `get_comments()` method. Sometimes the underlying **HTTP request** (made by the `requests` library) **hangs**, and my script gets **blocked** forever. Is there any simple way to put a timeout to HTTP requests? `requests` support timeout, but currently `instaloader` do not use this feature.

Thank You
Martino

corpone2:
Adds a `request_timeout` parameter to IL instantiation.

Closes #330

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -343,7 +343,7 @@ def main():
                        help='Tries to ensure downloaded images avoid corruption in case of unexpected interruption. '
                        'If the last picture is corrupted, Instaloader will fix the picture the next time it is run. '
                        'Requires the JSON metadata to be saved.')
-    g_how.add_argument('--request-timeout', action='store_true', metavar='N', type=int,
+    g_how.add_argument('--request-timeout', metavar='N', type=int,
                        help='seconds to wait before timing out a connection request')
 
     g_misc = parser.add_argument_group('Miscellaneous Options')
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -343,6 +343,8 @@ def main():
                        help='Tries to ensure downloaded images avoid corruption in case of unexpected interruption. '
                        'If the last picture is corrupted, Instaloader will fix the picture the next time it is run. '
                        'Requires the JSON metadata to be saved.')
+    g_how.add_argument('--request-timeout', action='store_true', metavar='N', type=int,
+                       help='seconds to wait before timing out a connection request')
 
     g_misc = parser.add_argument_group('Miscellaneous Options')
     g_misc.add_argument('-q', '--quiet', action='store_true',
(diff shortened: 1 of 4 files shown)


instaloader/__main__.py
```diff
@@ -343,6 +343,8 @@ def main():
                        help='Tries to ensure downloaded images avoid corruption in case of unexpected interruption. '
                        'If the last picture is corrupted, Instaloader will fix the picture the next time it is run. '
                        'Requires the JSON metadata to be saved.')
+    g_how.add_argument('--request-timeout', action='store_true', metavar='N', type=int,
+                       help='seconds to wait before timing out a connection request')
```
aandergr: Could you also document this in `docs/cli-options.rst`?

instaloader/instaloader.py
```diff
@@ -150,6 +151,7 @@ class Instaloader:
 
     def __init__(self,
                  sleep: bool = True, quiet: bool = False,
+                 request_timeout: Optional[int] = None,
```
aandergr: There is an invocation of the Instaloader constructor in Instaloader.anonymous_copy() (line 196+) which lacks the request_timeout argument. Please add `request_timeout=self.context.request_timeout` there.

instaloader/instaloadercontext.py
```diff
@@ -155,6 +157,10 @@ def get_anonymous_session(self) -> requests.Session:
                                 'ig_vw': '1920', 'csrftoken': '',
                                 's_network': '', 'ds_user_id': ''})
         session.headers.update(self._default_http_header(empty_session_only=True))
+        if self.request_timeout is not None:
+            # Override default timeout behavior.
+            # Need to silence mypy bug for this. See: https://github.com/python/mypy/issues/2427
+            session.request = partial(session.request, timeout=self.request_timeout) # type: ignore
```
aandergr: This also needs to be done in copy_session (line 23), load_session_from_file (line 172) and login (line 195), where a requests.Session object is created.

aandergr:
Resolves #330.

corpone2:
> Looks nice! I left a few comments that need to be addressed.

All issues should now be addressed.

> However, I am curious what we need this for / what problem it does solve.

When using a slow or unreliable connection the socket timeouts can be extremely long, say 60 seconds. This forces the session to timeout faster, and speeds up total crawl time by 60x or more.

fredrik-smedberg:
I suggest setting a default value for timeout.
https://requests.kennethreitz.org/en/master/user/quickstart/#timeouts

To me it seems counterproductive to default to the risk of Instaloader hanging indefinately. Perhaps a default value of 30 s or 60 s would suffice?

corpone2:
> Nice PR! I looked through your suggested code changes and after a short research it looks like this is indeed a good way of adjusting the timeout of the requests. However, I noticed that your typings and the argument for the CLI only allow integer values as timeout but requests also supports floats. It would be nice, if you could adjust this.

Sure. In practice there is minimal benefit to passing granular values here.

aandergr:
Thanks! I merged your PR and it will be included in the next Instaloader release.

PR #444, opened 2019-10-31
Use 'edge_media_to_comment' instead of 'edge_media_to_parent_comment' in Post

fredrik-smedberg:
This change make Instaloader use 'edge_media_to_comment' instead of 'edge_media_to_parent_comment' to speed up crawling the number of comments for posts (Post).

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -326,9 +326,9 @@ class Post:
     def comments(self) -> int:
         """Comment count including answers"""
         try:
-            return self._field('edge_media_to_parent_comment', 'count')
-        except KeyError:
             return self._field('edge_media_to_comment', 'count')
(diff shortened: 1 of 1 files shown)


somera:
Happy New Year!!! Can you relase this fix in the next days? Thx.

Thammus:
Thank you for this useful speedup idea. However, this change would also make Instaloader to fetch the full metadata dict if only `edge_media_to_parent_comment` is available through `self._node`. I committed changes (d8ba30b91b16cc110d3b6d1ff0366780750fee01) that first look for `edge_media_to_comment` without utilizing the `_field` function and thus without the risk of unnecessarily fetching additional data. Hence, the requested speedup will be experiencable in the next release of Instaloader.

Issue #457, opened 2019-11-14
Download Top X comments from a post


amitupreti:
**Provide us a use case of the feature**
It is very useful when we have to get the top comments from a post. In instagram browser/app, the comments are sorted by their popularity i.e top comments appear at first. But with instaloader we pull latest comments first.

**Describe the solution you'd like**
Option 1 : add param in `get_comments` function to indicate if we want top post or post by date

**Describe alternatives you've considered**
I browsed through snippets given in the documentation. We can also use the same logic as we used for top posts. But this would mean that we iterate through all the comments and use a lot of requests.

**If the feature request is accepted, would you be willing to submit a PR?**

Yes I already have a PR ready.

One solution on code snippet
```py
from itertools import islice
from math import ceil

from instaloader import Instaloader, Post

SHORTCODE = 'B41c_1dlf3X'        # post to download from
X_percentage = 10    # percentage of comment that should be downloaded

L = Instaloader()

post = Post.from_shortcode(L.context, SHORTCODE)
comments_sorted_by_likes = sorted(post.get_comments(),key=lambda p: p.likes_count, reverse=True)

for comment in islice(comments_sorted_by_likes, ceil(post.comments * X_percentage / 100)):
    print(comment.text)
```

But again, in case of large number of comment in a post, We would be sending a ton of requests.

nOOBIE-nOOBIE:
Alternatively if we just want to load the top 12 comments. we could do
`https://www.instagram.com/p/B07dk0OnnzE/?__a=1`
This will load the first top 12 comments that are displayed in browser.

What do you think?

Thammus:
These 12 top comments you describe are only intended to be used as a preview. We do not see any possibility to implement this feature of providing the top x comments without first fetching all comments followed by sorting.

Miracleikechi:
Nice one

Miracleikechi:
Nice one

Miracleikechi:
Nice one

PR #458, opened 2019-11-14
Adding code snippet for top x comments from a post

amitupreti:
Example code for top x comments from a post

Issue :https://github.com/instaloader/instaloader/issues/457


Thammus:
Like you described at issue #457, the provided code snippet uses the same logic as ["Top X Posts of User"](https://instaloader.github.io/codesnippets.html#top-x-posts-of-user). From this already documented snippet one can easily derive a working example that matches this use case with "top x comments". Hence we consider it redundant to include this snippet.

Issue #304, opened 2019-05-18
 Wrong directory output for tagged and highlights


UnbelievableCat:
**Describe the bug**
Instaloader creates tagged and higlights folders in same directory as profile folder. Example:
\profile_name
\profile_name ﹨꞉tagged
\profile_name﹨highlight

**To Reproduce**
instaloader -l user -p pass profile_name --tagged --higlights 

**Expected behavior**
Create folder in profile directory

**Instaloader version**
Windows standalone 4.2.7.

psylenced:
Similar problem for me (Windows).

The backslash is some how being replaced with a unicode backslash.

pxssy:
An issue related to this, for older versions, was that some highlights had slashes in their names, which caused certain errors when trying to save. instaloader would give up instantly, leaving the rest of the highlights without issues unable to be downloaded.

psylenced:
Any update on this?

Tagged/highlights are useless at the moment.

psylenced:
I think i've tracked it down to this commit:

https://github.com/instaloader/instaloader/commit/0c9a667da74d5dd27664d307130223edf530f4d9

The purpose of that commit (I assume) is to replace back-slashes in story titles with a unicode variation.

The problem is when you have a multi-path save ```profile\tagged``` or ```profile\highlights``` it seems to replace the windows path separator too.

What's happening:

```ReplaceWithUnicode(profilename\tagged) + \ + ReplaceWithUnicode(Filename)```

What it should do:

```ReplaceWithUnicode(profilename) + \ + tagged + \ + ReplaceWithUnicode(Filename)```

**Comment by psylenced** [comment]
Following on:

https://github.com/instaloader/instaloader/blob/b542936c0927015316a2b3800b6de8c13f9b7426/instaloader/instaloader.py#L456

```
        dirname = _PostPathFormatter(post).format(self.dirname_pattern, target=target)
        filename = dirname + '/' + self.format_filename(post, target=target)
```

The inputs for that are:

```
pattern = {target}
target = profilename\:tagged
```

The output is
```dirname = profilename﹨꞉tagged``` (note the unicode backslash)

So the finale filename is:

```
profilename﹨꞉tagged/anotherprofile_2016-01-11_12-12-12_12345-123456789123456789
```

with ```"profilename﹨꞉tagged"``` being a single first-level folder.

Meaning in the download root folder you have:

```
profile1
profile1﹨꞉tagged
profile1﹨:highlights
profile2
[...]

PR #553, opened 2020-03-28
Fixed issue #304 by undoing the changes in #0c9a667

DankMemeGuy:
Fixed issue #304 by undoing the changes in 0c9a667

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -108,7 +108,7 @@ class _PostPathFormatter(_ArbitraryItemFormatter):
         ret = super().vformat(format_string, args, kwargs)
         if platform.system() == 'Windows':
             ret = ret.replace(':', '\ua789').replace('<', '\ufe64').replace('>', '\ufe65').replace('\"', '\uff02')
-            ret = ret.replace('\\', '\ufe68').replace('|', '\uff5c').replace('?', '\ufe16').replace('*', '\uff0a')
+            ret = ret.replace('\\', '\u005c').replace('|', '\uff5c').replace('?', '\ufe16').replace('*', '\uff0a')
         return ret
 
     def get_value(self, key, args, kwargs):
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -265,7 +265,7 @@ class Post:
 
         .. versionadded:: 4.2.6"""
         def _elliptify(caption):
-            pcaption = ' '.join([s.replace('/', '\u2215') for s in caption.splitlines() if s]).strip()
+            pcaption = ' '.join([s.replace('/', '\u002f') for s in caption.splitlines() if s]).strip()
             return (pcaption[:30] + u"\u2026") if len(pcaption) > 31 else pcaption
         return _elliptify(self.caption) if self.caption else ''
 


Thammus:
Your changes do not actually revert the changes in 0c9a667. 0c9a667 does not break anything and is the wrong commit to revert since it does not introduce the replacement of the backslash.

Your changes intend to remove the replacement of the backslash at the `_PostPathFormatter` and remove the replacement of the slash at the Post property `pcaption`. The former could indeed address problems described in #304 but we would still need to make sure that the single fields of the path do not contain any illegal characters such as the backslash. A promising approach could be to move the replacement to `_PostPathFormatter.get_value()`.
As far as I can see, replacing the slash in the `pcaption` with a similar looking unicode character does not and never did lead to any problems and is necessary to use the pcaption as part of the filename (which was requested by some users).

DankMemeGuy:
> Your changes do not actually revert the changes in [0c9a667](https://github.com/instaloader/instaloader/commit/0c9a667da74d5dd27664d307130223edf530f4d9). [0c9a667](https://github.com/instaloader/instaloader/commit/0c9a667da74d5dd27664d307130223edf530f4d9) does not break anything and is the wrong commit to revert since it does not introduce the replacement of the backslash.
> 
> Your changes intend to remove the replacement of the backslash at the `_PostPathFormatter` and remove the replacement of the slash at the Post property `pcaption`. The former could indeed address problems described in #304 but we would still need to make sure that the single fields of the path do not contain any illegal characters such as the backslash. A promising approach could be to move the replacement to `_PostPathFormatter.get_value()`.
> As far as I can see, replacing the slash in the `pcaption` with a similar looking unicode character does not and never did lead to any problems and is necessary to use the pcaption as part of the filename (which was requested by some users).

Not sure how it can be the wrong commit, or how that ocmmit to revert since after making the changes I made the issue no longer exists. The commit reverts made all subfolders: /tagged, /highlights, erc. into the main directory:

/root/username
/root/username\:tagged (not a typo)
/root/username\:highlights (not a typo)

The changes made now actually nestle the subfolders into the main username's directory. It does seem to keep the ":" in the "tagged" and "highlights" folder names. But that's something I can live with. Not sure which version of Windows you're running, or your environment, but this was definitely an issue since the initial commits were made.

Thammus:
> Not sure how it can be the wrong commit, or how that's the wrong commit to revert
> [...]
> 0c9a667 made all subfolders: /tagged, /highlights, erc. into the main directory

Commit 0c9a667 only replaces the unicode backslash substitute with another unicode backslash substitute. The wrong subfolder behavior was already introduced a year before this commit with 2e517e9. 

I understand what you are trying to achieve and your change does address the incorrect subfolder behavior, but simply undoing the substitution puts things too simple and disobeys the reasons why these substitutions got implemented in the first place. As I wrote in the previous comment, we still need to make sure that the single fields of the path do not contain any illegal characters (for windows). This is why I suggested the approach to move the replacement to `_PostPathFormatter.get_value()`.

54656452:
I think I get what @Thammus is saying. The fix proposed by @DankMemeGuy solves the specific issue, but not the larger issue caused by Windows' accepted characters in filenames. Rather than making sure the final path doesn't contain an issue it is better to ensure that no segment of the final path has an issue? Would a final fix also involve replacing colons from the path?

psylenced:
> 
> 
> > Not sure how it can be the wrong commit, or how that's the wrong commit to revert
> > [...]
> > [0c9a667](https://github.com/instaloader/instaloader/commit/0c9a667da74d5dd27664d307130223edf530f4d9) made all subfolders: /tagged, /highlights, erc. into the main directory
> 
> Commit [0c9a667](https://github.com/instaloader/instaloader/commit/0c9a667da74d5dd27664d307130223edf530f4d9) only replaces the unicode backslash substitute with another unicode backslash substitute. The wrong subfolder behavior was already introduced a year before this commit with [2e517e9](https://github.com/instaloader/instaloader/commit/2e517e972fde3aef007d31ecf5c679f4f4214c13).

Actually I believe you're incorrect.

0c9a667da74d5dd27664d307130223edf530f4d9 was the change that broke #304 for windows. Prior to that everything was working. All the other character replacements did not affect windows downloads until the slash unicode was introduced.

I do agree that ```_elliptify``` does not need to be included as part of this PR.

I believe you should let this PR through and fix the unicode in a future fix.

Currently **all** tagged and highlight downloads are broken in windows, which is far more serious of a bug than the chance of a unicode slash existing in a username.

Ruthalas:
I concur with @psylenced - merge and open an another issue for the sanitizing usernames.
All uses of subfolders are non-functional on Windows.

Thammus:
As we already explained above, commit 0c9a667 only replaces the unicode backslash substitute with another unicode backslash substitute, hence it was not the change that broke #304. Nevertheless, talking about 0c9a667 is splitting hair about the title of this pull request which is misleading because of this falsely linked commit. 

> I do agree that `_elliptify` does not need to be included as part of this PR.

The change to `_elliptify` does not only not need to be included, it has nothing to do with the issue in question and does break a feature. The presence of this change is sufficient to reject this PR.

> 0c9a667 was the change that broke #304 for windows. Prior to that everything was working. All the other character replacements did not affect windows downloads until the slash unicode was introduced.

The other requested change of this PR does indeed logically revert the backslash substitution that was introduced with 2e517e9 along with other substitutions for characters that are illegal in Windows paths and therefore would fix the strange subfolder behavior described in #304. But as stated in all my comments regarding this PR, simply removing the slash substitution would also lead to strange subfolders for example if a slash is used inside a highlight title. 

> I believe you should let this PR through and fix the unicode in a future fix.

Given that there might be a better solution we cannot let this PR through. However, we did suggest a possible solution above: Move all Windows related substitutions to `_PostPathFormatter.get_value()` and only make them apply on single fields of the path, thus preventing the creation of strange folders as described in #304.

psylenced:
Given that 0c9a667 was on 7 May 2019, and #304 shortly after on 18 May 2019, that means the fix involving ```_PostPathFormatter.get_value()``` has been over 12 months without a resolution.

I do agree that there should be a better fix, but having functionality in windows builds broken for 12 months is not ideal.

0c9a667 introduced the regression, so should be rolled back until a bug-free fix for the original change is created.

Unfortunately I don't know python, so can't make the change myself - otherwise I would have.

In testing just now, simply replacing ```ret.replace('\\', '\ufe68')``` back to ```ret.replace('\\', '\uff3c')``` fixes the issue on windows.

I do totally understand what you're saying - by rolling back the fix you break special characters in the title.

However using that logic the reverse is true - in the original fix by removing strange characters in the title you introduce strange characters in **every** highlight, tag, story in windows.

Surely the best solution is to roll back the original buggy fix (0c9) and ensure both issues are fixed - which is what this PR tries to achieve.

aandergr:
Obsolete by #631.

Issue #571, opened 2020-04-15
Do not attempt to fetch invalid targets


e5150:
**Provide us a use case of the feature**
If an argument is mistyped, but valid according to `ArgumentParser`, it will be regarded as a download target. For instance `post-filter=whatever` will be treated as a profile name, resulting in unnecessary suspect queries and 404s.

**Describe the solution you'd like**
`_main` or the corresponding `download_*` functions could filter out seemingly invalid targets. `@followees` and `username` can be matched against `mention_regex`, `#hashtag` and `-shortcode` against `[A-Za-z0-9]+`, but I don't know what's valid for `%location`.

**If the feature request is accepted, would you be willing to submit a PR?**
Yes

aandergr:
We think this is a cool idea and we would like to see it implemented. The target matchings are done in [`__main__.py`, lines 139 ff](https://github.com/instaloader/instaloader/blob/master/instaloader/__main__.py#L139). Many cases of invalid targets could be already catched by using a regex matching instead of the way it is implemented now. I think your proposal would enhance the user friendliness without adding too much complexity. 

> **Describe the solution you'd like**
> `_main` or the corresponding `download_*` functions could filter out seemingly invalid targets. 

I'd prefer `_main()` to do this check, since we are slowly moving away from passing string target names to `download_*()` functions, e.g. `download_profiles()` already gets a `Set[Profile]` as target parameter.

>`@followees` and `username` can be matched against `mention_regex`, `#hashtag` and `-shortcode` against `[A-Za-z0-9]+`, but I don't know what's valid for `%location`.

`-shortcode` is a `-` and a URL-Base64 string, so it matches against `^-[A-Za-z0-9-_]+$`. I think `%location` are just numbers.

PR #577, opened 2020-04-18
Regex target matching

e5150:
Shortcodes seems to be 11 or 39 chars, and hashtags probably has some length restrictions. But without definitive specifications I'd err on the side of being permissive.
Related to #571


aandergr:
Thanks! We rebased it to `v4.4-dev` and merged your changes.

Issue #615, opened 2020-05-14
login fail


Rasko1:
**Describe the bug**
Bad request using --login

**To Reproduce**
instaloader --login=usernamehere

**Error messages and tracebacks**
Fatal error: Login error: JSON decode fail, 400 - Bad Request.

**Instaloader version**
4.4.1

**Additional context**
I have checked the code and it seems to be using the web request and the password there is somehow encrypted, the post data is not the password itself.

I have the mobile api in case you need it!

daprofessa19:
Yes, I'm also having this issue.

caioburin:
Same here. First time using, though.

TomoyaFujita2016:
Same here.

hawks-vision:
For me it was working fine till yesterday but now its not even tried uninstalling simplejson mpdule in python but it didnt help

Whovian9369:
Yep, can confirm - Had this *sometimes* starting with Version 4.4.but am getting this seemingly 100% of the time here.

seisdr:
Yes same here

Fukken-Saved:
Ohhh! Not again!

TomoyaFujita2016:
I felt that the response from instagram is a little bit slower than usual in console.

netmannet:
Login error: JSON decode fail, 400 - Bad Request.

twitter-79:
I analyzed what happens within a normal login procedure on instagram.com using chrome console.
Instagram is loading encryptionUtils.js, encrypts the password, and sends it as enc_password in FormData. 
It is possible to use Instaloader when using the enc_password instead of the normal password as used before.
I hope this information can help the developer to find a solution to the issue.

colemannoah:
Just started using this today, also getting this bug - is it out of the usual?

Fukken-Saved:
Yes, very. I've been running it a bit more than a year, and a few times I've missed stories, but that was because I forgot to run it. Guess I get to do an overlapping highlights scan of everything once it works again. You can't get everything, even if you're me.

u174458:
I started to having this issue since yesterday:

user@user-mycomputer:~/ig$ instaloader --version
4.4.1

user@user-mycomputer:~/ig$ instaloader profile xxxxxxxx --fast-update --login myigusername
Session file does not exist yet - Logging in.
Enter Instagram password for myigusername: 
Traceback (most recent call last):
  File "/home/marcelo/.local/bin/instaloader", line 8, in <module>
[...]

PR #623, opened 2020-05-19
Workaround for Instagram's encrypted login.

Infinitay:
Upon checking the POST request made by Instagram, they started encrypting the password starting with `#PWD_INSTAGRAM_BROWSER:10`, in my case. When looking at the source, I came across the following:

```js
const t = '#PWD_INSTAGRAM_BROWSER',
    n = Object.freeze({
    PLAINTEXT: '0',
    ROTATED_ENCRYPT: '6',
    FALLBACK_ENCRYPT: '9'
});
```

Apparently the new request sends it in the form of something like `<#endpoint>:<encryption_version>:<timestamp>:<aes(?)_encrypted_password>`. However, as @pgrimaud found, we can send the login request using the plaintext password as opposed to having to encrypt it.

Of course this is only a temporary solution. I think it'd be better to implement the encryption; however, this will do in the meantime.

Solves: https://github.com/instaloader/instaloader/issues/615
**Workaround Credit**: @pgrimaud
**See**: https://github.com/pgrimaud/instagram-user-feed/commit/96ad4cf54d1ad331b337f325c73e664999a6d066

---

Further resources:
https://github.com/dilame/instagram-private-api/issues/1010
https://github.com/FeezyHendrix/Insta-mass-account-creator/issues/140

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -7,7 +7,6 @@ import shutil
 import sys
 import textwrap
 import time
-import calendar
 import urllib.parse
 from contextlib import contextmanager
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


instaloader/instaloadercontext.py
```diff
@@ -7,6 +7,7 @@
 import sys
 import textwrap
 import time
+import calendar
```
aandergr, line 10:
```suggestion
```

We can use `datetime` (already imported) instead of `calendar`.

instaloader/instaloadercontext.py
```diff
@@ -215,8 +216,9 @@ def login(self, user, passwd):
         session.headers.update({'X-CSRFToken': csrf_token})
         # Not using self.get_json() here, because we need to access csrftoken cookie
         self.do_sleep()
+        # Workaround credits to pgrimaud. See: https://github.com/pgrimaud/instagram-user-feed/commit/96ad4cf54d1ad331b337f325c73e664999a6d066
         login = session.post('https://www.instagram.com/accounts/login/ajax/',
-                             data={'password': passwd, 'username': user}, allow_redirects=True)
+                             data={'enc_password': f'#PWD_INSTAGRAM_BROWSER:0:{calendar.timegm(time.gmtime())}:{passwd}', 'username': user}, allow_redirects=True)
```
aandergr, line 221:
```suggestion
                             data={'enc_password': enc_password, 'username': user}, allow_redirects=True)
```
(use abovely-mentioned `enc_password` and fix too long line)

Issue #529, opened 2020-02-22
_PostPathFormatter breaks Windows-style path


fireattack:
**Describe the bug**

_PostPathFormatter is used too aggressively in multiple places, which results in wrong folder name and structure.

Example no.1:

```
instaloader --dirname-pattern "D:\test" instagram
```
in command line on Windows, it will download files to `.D꞉﹨test`, instead of proper `D:\test` (because both `:` and `\` are substituted).

Example no.2:

When downloading highlights, it by default download to a subfolder  `Path(name) / Path(user_highlight.title)`:

https://github.com/instaloader/instaloader/blob/819426efdb722596088fe80705c947e6b20815cb/instaloader/instaloader.py#L671

However, this Path-object later is butchered by `_PostPathFormatter` at

https://github.com/instaloader/instaloader/blob/819426efdb722596088fe80705c947e6b20815cb/instaloader/instaloader.py#L597

which broke all the `\`s on Windows.

![image](https://user-images.githubusercontent.com/4130991/75097179-54292400-556d-11ea-8d4c-3817aebba230.png)

My suggestion is to re-think where we want to substitute the invalid characters. Maybe only do it against filename/foldername (and should also replace "/" there for any OS, if we want to avoid unexpected subdirectory creation), not fullpath. Probably only against automatically generated fields, too.

Version: 4.3a1

fireattack:
I feel like we should only replace "\" to "＼" if it's from any automatically generated fields (`{profile}` etc.). If the user intentionally feeds in a "\" themselves it should be treated just like "/".

Actually, now I think about it, we don't do anything special to "/" (regardless of the platform). So while seems practically impossible, having "/" in automatically generated fields will cause it to create subfolders in any OS.

fireattack:
Also, I feel like full-width colon (U+FF1A) would be a better substitution for ":" than what we were using now (MODIFIER LETTER COLON, "U+A789"). It would be more consistent will others (we use full-width backslash, pipe, etc. already).

fireattack:
Updated the description and title to reflect some new information.

This problem is more prominent than I think; it happens all the time when you download highlights (which by default saved to a subdir in the main dir).

blablable123456:
Theres a trick that I use for quite a while and it works wonders. Simple put a 1 at the end of line 671. Of course this is just a fix for this isolated issue. By the way, this issue of folder formatting doesnt bug on linux, only on windows.

` else Path(name) / Path(user_highlight.title + '1')) `

[...]

PR #631, opened 2020-05-22
Only substitute characters from auto generated fields

fireattack:
…in _PostPathFormatter.

Suggested by @Thammus 

This fixes issues of path problem on Windows such as #304, #529, etc.

This also overrides #553.

Also made the following changes:

1. Changes substitution of colon to full-width colon (U+FF1A) instead of MODIFIER LETTER COLON (U+A789) to be consistent with others (we use full-width backslash, pipe, etc. already).
1. Also replaces `\r` on Windows with space, just like `\n`.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -104,10 +104,14 @@ class _ArbitraryItemFormatter(string.Formatter):
 
 class _PostPathFormatter(_ArbitraryItemFormatter):
     def get_value(self, key, args, kwargs):
-        """Replaces '/' with similar looking Division Slash. Also replaces illegal filename characters on windows."""
         ret = super().get_value(key, args, kwargs)
         if not isinstance(ret, str):
             return ret
+        return self.sanitize_path(ret)
+
+    @staticmethod
+    def sanitize_path(ret: str) -> str:
+        """Replaces '/' with similar looking Division Slash and some other illegal filename characters on Windows."""
         ret = ret.replace('/', '\u2215')
         if platform.system() == 'Windows':
             ret = ret.replace(':', '\uff1a').replace('<', '\ufe64').replace('>', '\ufe65').replace('\"', '\uff02')
@@ -675,7 +679,8 @@ class Instaloader:
             name = user_highlight.owner_username
             highlight_target = (filename_target
                                 if filename_target
-                                else Path(name) / Path(user_highlight.title))  # type: Union[str, Path]
+                                else Path(_PostPathFormatter.sanitize_path(name))
+                                / _PostPathFormatter.sanitize_path(user_highlight.title))  # type: Union[str, Path]
             self.context.log("Retrieving highlights \"{}\" from profile {}".format(user_highlight.title, name))
             self.download_highlight_cover(user_highlight, highlight_target)
             totalcount = user_highlight.itemcount
@@ -936,8 +941,10 @@ class Instaloader:
         .. versionadded:: 4.1"""
         self.context.log("Retrieving tagged posts for profile {}.".format(profile.username))
         self.posts_download_loop(profile.get_tagged_posts(),
-                                 target if target else Path(profile.username) / Path(':tagged'), fast_update,
-                                 post_filter)
+                                 target if target
+                                 else Path(_PostPathFormatter.sanitize_path(profile.username))
+                                 / _PostPathFormatter.sanitize_path(':tagged'),
+                                 fast_update, post_filter)
 
     def download_igtv(self, profile: Profile, fast_update: bool = False,
                       post_filter: Optional[Callable[[Post], bool]] = None) -> None:
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -679,8 +679,8 @@ class Instaloader:
             name = user_highlight.owner_username
             highlight_target = (filename_target
                                 if filename_target
-                                else Path(_PostPathFormatter.sanitize_path(name))
-                                / _PostPathFormatter.sanitize_path(user_highlight.title))  # type: Union[str, Path]
+                                else (Path(_PostPathFormatter.sanitize_path(name)) /
+                                      _PostPathFormatter.sanitize_path(user_highlight.title)))  # type: Union[str, Path]
             self.context.log("Retrieving highlights \"{}\" from profile {}".format(user_highlight.title, name))
             self.download_highlight_cover(user_highlight, highlight_target)
             totalcount = user_highlight.itemcount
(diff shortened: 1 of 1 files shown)


fireattack:
Tested downloading `highlights` and using `D:\test` as --dirname-pattern`, all seem to work fine.

Please let me know if I missed something.

instaloader/instaloader.py
```diff
@@ -103,22 +103,16 @@ def format_field(self, value, format_spec):
...
-        if platform.system() == 'Windows':
-            ret = ret.replace(':', '\ua789').replace('<', '\ufe64').replace('>', '\ufe65').replace('\"', '\uff02')
-            ret = ret.replace('\\', '\ufe68').replace('|', '\uff5c').replace('?', '\ufe16').replace('*', '\uff0a')
-        return ret
-
     def get_value(self, key, args, kwargs):
-        """Replaces '/' with similar looking Division Slash and on windows newline with space"""
+        """Replaces '/' with similar looking Division Slash. Also replaces illegal filename characters on windows."""
         ret = super().get_value(key, args, kwargs)
         if not isinstance(ret, str):
```
aandergr, line 109: There are two cases where the _PostPathFormatter is invoked with a `Path` as target:
1. `target=Path(profile.username) / Path(':tagged')` https://github.com/instaloader/instaloader/blob/bf2f4f16ed7648fb5970a3d6c0d3e23152c0a6fd/instaloader/instaloader.py#L944-L946
2. `target=Path(name) / Path(user_highlight.title)` https://github.com/instaloader/instaloader/blob/bf2f4f16ed7648fb5970a3d6c0d3e23152c0a6fd/instaloader/instaloader.py#L682-L684

With get_value() only doing the replacement logic when it gets a `str`, these cases are not covered, so we'd miss the colon replacement in the first case, and the character replacements in the highlight title in the second case.

This could be fixed by invoking the replacement logic when `ret` is a `str`, *and* for each Path part when `ret` is a `Path` – Only for each part of the Path, so that we do not replace the path separator (slash / backslash) between the username and `:tagged` or highlight title while still replacing title characters that are disallowed in Windows paths.

instaloader/instaloader.py
```diff
@@ -103,22 +103,16 @@ def format_field(self, value, format_spec):
...
             return ret
         ret = ret.replace('/', '\u2215')
         if platform.system() == 'Windows':
-            ret = ret.replace('\n', ' ')
+            ret = ret.replace(':', '\uff1a').replace('<', '\ufe64').replace('>', '\ufe65').replace('\"', '\uff02')
```
aandergr, line 113: (it is technically not backwards-compatible to now replace the colon by a different character. However, I think given how broken the Windows path handling is right now, we can still let this through)

fireattack:
Thanks! I can handle `Path` object in `get_value`, but I do have a deeper question about the implementation.

Particularly, I don't think `else Path(name) / Path(user_highlight.title))` is a good way to do it to begin with. This will cause problem even on *nix if the highlight title consists of "/" - and this is no way to fix it later in _PostPathFormatter, because you can't distinguish which "/" is part of the filename, and which is a slash in path reliably any more.

What do you think if we create a new utility function `substitute_character` (substitute "/" only for *nix, and other cases for Windows) and call it in both _PostPathFormatter's `get_value()`, and on string `user_highlight.title` before attach it to Path(name)?

aandergr:
> Particularly, I don't think `Path(name) / Path(user_highlight.title))` is a good way to do it to begin with. This will cause problem even on *nix if the highlight title consists of "/" - and there is no way to fix it later in `_PostPathFormatter`, because you can't distinguish which "/" is part of the title, and which is a slash in path reliably any more.

Ah right, `Path("a/b")` is equivalent to `Path("a") / Path("b")`, so a highlight with a title consisting of slashes would end up in undesired subdirectories.

> What do you think if we create a new utility function `substitute_character` (substitute "/" only for *nix, and other cases for Windows) and call it in both _PostPathFormatter's `get_value()`, and on string `user_highlight.title` before attach it to `Path(name)` (of course, `:tagged` in another case)?

I.e., substitute_character() is called in get_value() when the value is a string, and instead of `Path(user_highlight.title)` we would do `Path(substitute_character(user_highlight.title))` (likewise for :tagged). I think that might be a legit solution. Maybe we should call it sanitize_path() or so, since substitue_character() is a name that does not really tell its purpose.

instaloader/instaloader.py
fireattack: Resolved in 852b0eec2134b2a1b0129dc53be189b06a6dcbfb

fireattack:
Thanks! I added a static method in _PostPathFormatter. Let me know if it's not OK.

instaloader/instaloader.py
```diff
@@ -681,7 +679,8 @@ def download_highlights(self,
             name = user_highlight.owner_username
             highlight_target = (filename_target
                                 if filename_target
-                                else Path(name) / Path(user_highlight.title))  # type: Union[str, Path]
+                                else Path(_PostPathFormatter.sanitize_path(name))
+                                / _PostPathFormatter.sanitize_path(user_highlight.title))  # type: Union[str, Path]
```
aandergr, line 683:
```suggestion
                                else (Path(_PostPathFormatter.sanitize_path(name)) /
                                      _PostPathFormatter.sanitize_path(user_highlight.title)))  # type: Union[str, Path]
```

instaloader/instaloader.py
```diff
@@ -942,8 +941,10 @@ def download_tagged(self, profile: Profile, fast_update: bool = False,
         .. versionadded:: 4.1"""
         self.context.log("Retrieving tagged posts for profile {}.".format(profile.username))
         self.posts_download_loop(profile.get_tagged_posts(),
-                                 target if target else Path(profile.username) / Path(':tagged'), fast_update,
-                                 post_filter)
+                                 target if target
+                                 else Path(_PostPathFormatter.sanitize_path(profile.username))
+                                 / _PostPathFormatter.sanitize_path(':tagged'),
```
aandergr, line 946:
```suggestion
                                 else (Path(_PostPathFormatter.sanitize_path(profile.username)) /
                                       _PostPathFormatter.sanitize_path(':tagged')),
```

instaloader/instaloader.py
fireattack: Thanks! Yeah I'm actually not sure which one we prefer (can't find any other link break at binary operator), so I followed https://www.flake8rules.com/rules/W504.html

Do you think we should use after-operator instead? I can change it.

aandergr: Indeed, having the operator at the beginning of the line might be better. Whatever you prefer :)

fireattack: > Indeed, having the operator at the beginning of the line might be better. Whatever you prefer :)

I will just leave it as what you suggested I think. Thanks!

aandergr:
Merged and released with [Version 4.4.3](https://github.com/instaloader/instaloader/releases/tag/v4.4.3). Thanks a lot for this bugfix!

psylenced:
Many thanks @fireattack for this PR. Really appreciate it!

Issue #257, opened 2019-02-17
fast_update not work with highlights/stories when in commit-mode 


pjw91:
**Describe the bug**
When downloading highlights/stories with the commit-mode and fast-update are on,
the fast-update has no effect.

**To Reproduce**
instaloader -F --highlights SOME_PROFILE --commit-mode --no-posts

**Expected behavior**
Skip those downloaded, as if commit-mode off.

**Instaloader version**
4.2.3

**Additional context**
The reason is that `self._committed` is never set in `download_highlights` and `download_stories`, and never reset to None after a post is processed (either downloaded or skipped).

That means:
1. `_committed` is always `None`, which effectively disables commit-mode
1. After processing a post, the flag is set to True/False, no matter what actually status of the next highlights/stories is.

After a quick review, it seems that add a line of code `self._committed = self.check_if_committed(filename)` at the top of `download_pic` fix this issue.
However, I think it should be add to `download_stories/highlights` or somewhere else.

aandergr:
Thanks for reporting this and sharing your insights!

> After a quick review, it seems that add a line of code `self._committed = self.check_if_commited(filename)` at the top of `download_pic` fix this issue.

That would check_if_committed for each item in a sidecar, though one check would be sufficient.

> However, I think it should be add to `download_stories/highlights` or somewhere else.

Or `download_storyitem`, which is used both for downloading stories and highlights.

Also, I think passing this _committed state through a class variable is unclean, error-prone and blocking thread-parallelization - we may have merged the PR (#139) too early and probably should reimplement the commit mode. @sushilicious any ideas/thoughts about this?

davispuh:
This looks fixed/resolved to me, no ?

aandergr:
> This looks fixed/resolved to me, no ?

No, there hasn't been any work on this.

sushilicious:
Yeah sorry I've never used the stories or highlights feature, so I didn't think to implement the commit mode for that. As is mentioned in the first post, adding the check for committed in `download_stories/highlights` is probably the right thing to do.

[...]

PR #661, opened 2020-06-04
Deprecate commit-mode

e5150:
Rather than checking the json file to make sure posts have been
successfully downloaded, data is stored to in a temporary file which
is renamed when downloading has finished, as suggested in #257.

Basing on master, as there isn't any dev branch atm. Also did not touch the docs, as I don't know the value for .. deprecated::. And maybe check_if_committed ought to remain, but with a deprecation warning.

Should be based on up-to-date master now…

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -348,9 +348,7 @@ def main():
                             'connection fails, it can be manually skipped by hitting CTRL+C. Set this to 0 to retry '
                             'infinitely.')
     g_how.add_argument('--commit-mode', action='store_true',
-                       help='Tries to ensure downloaded images avoid corruption in case of unexpected interruption. '
-                       'If the last picture is corrupted, Instaloader will fix the picture the next time it is run. '
-                       'Requires the JSON metadata to be saved.')
+                       help='This option is deprecated.')
     g_how.add_argument('--request-timeout', metavar='N', type=float,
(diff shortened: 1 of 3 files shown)


aandergr:
Hi! I like the idea of replacing the commit mode with a simple "create \<file\>.temp; mv \<file\>{.temp,}".

It is no problem that the `.. deprecated::` directive is missing, since before releasing, we usually scroll through the code and add/update these directives. Instead of a help text explaining that the option should not be used, we usually pass `help=SUPPRESS` to argparse for the particular argument, making it disappear from the help while still being considered a valid option when parsing.

However – I am confused that you withdraw the pull request. Is there any problem with the approach or the code that I am overseeing?

e5150:
Yeah, using `SUPPRESS` makes more sense.

I found no problems while doing basic testing. Was just having trouble rebasing to get around merge conflicts and ended up deleting my, apparently outdated, github-fork, and didn't get around to fix/make a new PR.
Is this PR usable, or should I make a new one?

aandergr:
I cannot reopen this pull request (maybe because the originating branch was deleted), neither do I have the option to rebase it or the information box for how to access its commits. Even though they are obviously saved somewhere, it would probably be a lot easier if you could make a new PR. It's no problem if it is based on an outdated branch, I could try rebasing it myself, since the maintainers always have write access to branches linked to a PR.

Issue #636, opened 2020-05-26
Continuing using a stored cursor token


salehio:
To avoid throttling, I wanted the ability to fetch data only after a certain date. Trying the method listed [here](https://instaloader.github.io/codesnippets.html#download-posts-in-a-specific-period) it appears that the filter is client side, and would page through all the GraphQL requests grabbing the cursor token to build the next one.

Is it possible anywhere in the client to pass in a previously stored continuation token? Do you know how long these are normally valid for?

If there's no supported way to do that currently, I'd be interested in contributing it as a feature the the module itself, but not necessarily wiring it up to the CLI.

aandergr:
> Is it possible anywhere in the client to pass in a previously stored continuation token?

That is not yet possible. However, it might be a smart way to resume an interrupted download loop. Our client for [GraphQL Cursor Connections](https://relay.dev/graphql/connections.htm), which are typically used for paginated data, is implemented in InstaloaderContext.graphql_node_list():

https://github.com/instaloader/instaloader/blob/72d86533f25180bc7e1df467718c568ac2b77d53/instaloader/instaloadercontext.py#L496-L525

In theory, you could pass as first_data an object that has data['page_info'] setup up such that graphql_node_list() starts with a previously stored end_cursor. However, there is no way to inject such a token into the functions that call graphql_node_list() yet.

> Do you know how long these are normally valid for?

I don't know how long the cursors last, but I am optimistic that they last forever. Some backend implementations use an encoding of the node type and an id unique within types of the same node. In the case of Instagram post connections, I do not know what is is exactly.

salehio:
I'll try passing in the first_post soon to see if that works smoothly.

Implementing this seems dirty. Either we can create an optional parameter for the token and add it to all functions that call `graphql_node_list` --  or have it be something you set on the `Profile` class, and it gets implicitly used, but the user can reset it if they don't want to use it anymore. That's not very clean, since (I guess) a token would only be valid for a single method. You couldn't use a token for getting all the posts for also paging through `similar accounts`.

Any preferences?

salehio:
Just realized, we would also have to expose a way for the module user to retrieve the token for storing.

aandergr:
Cleanly implementing this would require some refactoring. An approach could be making the graphql_node_list iterator to an iterator class which exposes some way to handle setting the token, and make all functions using graphql_node_list to iterators inherited by the new graphql node list class, thereby also allowing the user to set an initial cursor token.

[...]

PR #683, opened 2020-06-11
Allow resuming a fetch operation using end_cursor. Implements #636.

tonycpsu:
* Add new `end_cursor` property to `Post` that grabs the most recent
  `end_cursor` value obtained from the API.
* Add new `end_cursor` parameter to `Post.get_posts` that allows passing
  of a stored end_cursor value (obtained from the above property) to
  resume a fetch operation.
* In order to support this, add new `page_size` parameters to
  `InstaLoader` and `InstaLoaderContext` for overriding the default
  graphql page size.  Because end_cursor points at the last post on a
  page, the client needs to know and be able to set the page size for
  accurate results (i.e. to avoid missing posts that were part of a
  page but not consumed from the generator.

### DIFF
--- a/instaloader/__init__.py
+++ b/instaloader/__init__.py
@@ -1,7 +1,7 @@
 """Download pictures (or videos) along with their captions and other metadata from Instagram."""
 
 
-__version__ = '4.4.5rc1'
+__version__ = '4.4.5'
 
 
 try:


Thammus:
Thank you for this nice proof of concept. It is useful to see that resuming an iteration using the end_cursor actually works.

As far as we understand your code saves the end_cursor of the current page. This implies that some posts go missing on resuming if a page_size greater than one is used. Storing the end_cursor of the previous page instead of the current page would lead to duplicates instead of missing posts on resuming. Also, since duplicate posts do not cause problems and no posts are missed, this way it would not be necessary to alter the page size. 

Exposing the end_cursor as a property and changing the iterating method to accept an end_cursor as parameter would be a necessary change for all structures that allow resuming. It would be better to find a solution that does not require such structural changes.

tonycpsu:
No posts go will go missing as long as the caller consumes the entirety of the result set that comes back from the query.  That's why I chose to add the `page_size` changes, so that a caller who wants to get 10 or 20 items at a time doesn't end up with an `end_cursor` for the 50th item. 

There's really no other way I can see to do it unless you want to try to use each post's individual `end_cursor` value from `Post._full_metadata["edge_media_to_parent_comment"]["page_info"]["end_cursor"]`, but my experimentation with those values is that they don't reliably resume with that specific item, but at an item one or two posts chronologically later (earlier when we're iterating in reverse chronological order).  That would lead to duplicates that the caller would have to manage somehow.

aandergr:
What we in fact need is a graphql node list implementation that, besides yielding the nodes, also exposes information that can later be used to restore an iteration instance to resume an iteration. These resume information are: all the graphql_node_list() parameters and the method's internal state.

I think there is no need to expose/store the method's parameters, since in practice they can be always restored by the caller if we want to resume an interrupted operation. The method's internal state is the current data object (which could be restored by the previous end_cursor or first_data) and the current index within the current data page. Id est, what our graphql node list needs to expose is its current data object and the current index within the current page, or alternatively the previous end_cursor and the current index within the current page. With these two objects, we could resume an iteration without misses or duplicates.

(This can be further optimized, e.g. just using the remaining/unconsumed nodes in the data object plus the page_info)

It feels like violating the relation semantics to store the resume information inside the objects that are yielded by the iteration itself. The same objects can be created by completely different iterations. 

I propose to let the node iterator expose this information in a separate way. E.g. such that the caller does something like

```python
node_iterator = NodeIterator(...)
try:
    for node in node_iterator:
        yield Post(self.context, node)
except SomeInterruption:
    save_somehow(node_iterator.restore_information)
```

This could be achieved by refactoring the current graphql_node_list() into a full class implementing the [iterator protocol](https://docs.python.org/3/library/stdtypes.html#iterator-types).

PR #727, opened 2020-07-12
Bypass IP-based redirects to /accounts/login

cloudrac3r:
There are two main changes made:

For users, we request /{username}/feed/ instead of /{username}/. For some
reason, this completely bypasses the login redirect. This page doesn't
work in browser while blocked, but fortunately all the data we need is
just present in the HTML page.

For posts, we change from using the /p/ page to using the graphql
endpoint for the same data, which is still subject to graphql rate
limits, but is not subject to login redirects. The data is identical
between the two pages, apart from the object keys being sorted
differently and rhx_gis being missing on graphql.

Yes, this now unblocks access from VPNs, Tor, cloud servers, etc.

\-

This is my first time contributing to Instaloader, though I'm familiar with how Instagram works, since I'm the main developer of Bibliogram.

You use both single quotes and double quotes in the code. I didn't know which to use, so I picked single quotes, since there seemed to be more of those.

I don't expect this fix will work for very long, since Instagram will definitely catch on at some point. Still, it's something.

I'm not sure how to run the tests, if you have tests, but it works for me. (tm)

Assuming my code style is fine and the tests pass, this is ready to be merged.

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -142,9 +142,12 @@ class Post:
 
     def _obtain_metadata(self):
         if not self._full_metadata_dict:
-            pic_json = self._context.get_json("p/{0}/".format(self.shortcode), params={})
-            self._full_metadata_dict = pic_json['entry_data']['PostPage'][0]['graphql']['shortcode_media']
-            self._rhx_gis_str = pic_json.get('rhx_gis')
+            pic_json = self._context.graphql_query(
+                '2b0673e0dc4580674a88d426fe00ea90',
+                {'shortcode': self.shortcode}
+            )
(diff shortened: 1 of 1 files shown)


cloudrac3r:
I've applied your patch and unlocked the branch so that you can push more commits to it if you want to.

Surprisingly, I don't actually have an account on Instagram, and unfortunately I don't wish to create one.

I've never needed to use rhx_gis in Bibliogram, and posts I've read about it online seem to suggest that it's no longer required, but I don't know anything else about it or its purpose. You'll have to decide what you want to do with it.

cloudrac3r:
Awesome, glad I could help!!

PR #732, opened 2020-07-18
Resume a previously-aborted post download loop

### LINKED ISSUE #636: Continuing using a stored cursor token

aandergr:
With this change, Instaloader is capable of resuming a previously-aborted download loop.  To do so, it creates a JSON file within the target directory when interrupted, that contains all the necessary information to later resume that operation.

Resuming an interrupted download is supported for most, but not all targets. It is supported for:
- Regular profile posts,
- IGTV posts
- Saved posts,
- Tagged posts,
- Explore posts.

This supersedes #683 and closes #636.

I consider it ready to be merged, however open for discussion, critics, testing, etc.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,7 +49,9 @@
   directory accordingly,
 
 - allows **fine-grained customization** of filters and where to store
-  downloaded media.
+  downloaded media,
+
+- automatically **resumes previously-interrupted** download iterations.
 
 ::
 
--- a/docs/as-module.rst
+++ b/docs/as-module.rst
@@ -226,8 +226,8 @@ Exceptions
 
 .. autoexception:: TooManyRequestsException
 
-``NodeIterator`` (Interruptable Iterator)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Resumable Iterations
+^^^^^^^^^^^^^^^^^^^^
 
 .. versionadded:: 4.5
 
@@ -236,6 +236,8 @@ Exceptions
 
 .. autoclass:: FrozenNodeIterator
    :no-show-inheritance:
+
+.. autofunction:: resumable_iteration
 
 ``InstaloaderContext`` (Low-level functions)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/as-module.rst
+++ b/docs/as-module.rst
@@ -226,6 +226,17 @@ Exceptions
 
 .. autoexception:: TooManyRequestsException
 
+``NodeIterator`` (Interruptable Iterator)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 4.5
+
+.. autoclass:: NodeIterator
+   :no-show-inheritance:
+
+.. autoclass:: FrozenNodeIterator
+   :no-show-inheritance:
+
 ``InstaloaderContext`` (Low-level functions)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -239,19 +250,6 @@ Exceptions
 """"""""""""""""""
 
 .. autoclass:: RateController
-   :no-show-inheritance:
-
-   .. versionadded:: 4.5
-
-``NodeIterator``
-""""""""""""""""
-
-.. autoclass:: NodeIterator
-   :no-show-inheritance:
-
-   .. versionadded:: 4.5
-
-.. autoclass:: FrozenNodeIterator
    :no-show-inheritance:
 
    .. versionadded:: 4.5
--- a/docs/as-module.rst
+++ b/docs/as-module.rst
@@ -242,3 +242,16 @@ Exceptions
    :no-show-inheritance:
 
    .. versionadded:: 4.5
+
+``NodeIterator``
+""""""""""""""""
+
+.. autoclass:: NodeIterator
+   :no-show-inheritance:
+
+   .. versionadded:: 4.5
+
+.. autoclass:: FrozenNodeIterator
+   :no-show-inheritance:
+
+   .. versionadded:: 4.5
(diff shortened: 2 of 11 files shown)


tonycpsu:
@aandergr, unless I'm missing something, it seems like your solution is optimized for the CLI case where someone is Ctrl-C-ing a command-line session, rather than what seems to me to be a more common case where the caller just wants to get N posts at a time and store the cursor to be resmed later (which is the use case I had in mind.)  Is there a way for the caller to get the state from this iterator programmatically?  Also, I'm not sure I like the idea of creating JSON files as a side effect of a library call vs. having the caller manage that state, which allows for the flexibility of persisting the state to a database.

aandergr:
> Is there a way for the caller to get the state from this iterator programmatically?

The NodeIterator provides the method `freeze`, which returns its state as a FrozenNodeIterator. So the caller could easily stop after N posts, *freeze* the iterator, and later *thaw* it on an equally-constructed fresh NodeIterator to resume. 

The iterator itself does not save any JSONs or similar, that's the job of the caller.

tonycpsu:
OK, that sounds great!  I'll try to take a closer look soon.

tonycpsu:
I'm trying to integrate this with my app, and I see that the size of the FrozenNodeIterator object that comes back is something like 18kb when saved to JSON.  Pardon my ignorance of the way the Instagram API does things, Is all of the data in there necessary to support simply resuming from a stored cursor?  My implementation in #683, twhile not as clean from an API perspective, only required saving the `end_cursor` attribute, and seemed to work fine.

aandergr:
Strictly speaking, not all of the data is necessary: there are the fields 'query_hash', 'query_variables', 'query_referer' and 'context_username' which are only used to verify that the resume data comes from a matching iterator instance. However, these fields only take up a few hundred bytes.

For resuming, the FrozenNodeIterator contains not only the `end_cursor`, but all the edges that have not yet been consumed. These might take up several kilobytes, especially when interrupted early within a page.

tonycpsu:
More efficient given a set of assumptions about the importance of network traffic and API calls vs. storage of the frozen iterators, sure.  It just seems strange to be storing node data in the iterator instead of storing a pointer so that the data can be retrieved in the future.  I'd rather throw away and retrieve the node data the next time I resume, knowing that I'm getting the most up-to-date information from those nodes when I fetch them later.

Is there a way you can adapt this to just require passing those mandatory attributes in, and have it re-fetch the nodes that were read but not consumed prior to freezing the iterator?

aandergr:
With the size of these files typically being below 20 kB compressed (and there being at most one such file per target), I think for the vast majority of use cases, we did the right trade-off here in terms of efficiency and simplicity of the code versus storage. Of course, as soon as this already-big code change proves to be reliable, we can implement alternative options for freezing iterators.

tonycpsu:
OK.  Guess I'll stick with my personal fork until such options are present.  Thanks.

aandergr:
OK, I'm sorry that this implementation does not fulfill your requirements. May I ask why you are so constrained regarding the FrozenNodeIterator file size?

Anyway, thanks a lot for sharing your thoughts here and for your PoC in #683, which demonstrated the longevity of pagination cursors!

tonycpsu:
It's not the size constraint so much -- though 20k per feed isn't what I'd call trivial -- but the staleness of the data that concerns me the most here.

What bothers me is the idea of a user loading the first N items in a feed, then having a bunch of retrieved but not consumed items in that feed cached in an iterator object whose job description shouldn't really include caching data under normal circumstances.  Then when the user goes to retrieve more posts from that feed, perhaps days or weeks later, they get the old node data, with expired links to content.

In my experience, Instagram expires URLs fairly often, particularly for videos, or this wouldn't bother me so much.

aandergr:
> Then when the user goes to retrieve more posts from that feed, perhaps days or weeks later, they get the old node data, with expired links to content.

Well, that's false and as such just not a valid concern. The only links saved within the nodes are the URL of pictures and the shortcode (being used as a link to fetch more metadata), which I've not seen change within the lifetime of a Post. So the user won't get any *expired links to content*. The expiry is merely limited to outdated captions or metadata such as "like count" and mistakenly returned nodes that may have been deleted from the feed or that have lost their validity, both cases being very tolerable for the majority of usages.

aandergr:
The display URLs seem to be signed for 30 days, which I still consider sufficient for most cases of resuming a previously-aborted loop, especially since we don't know for sure yet how long pagination tokens are valid, and their expiry might further impose limitations on the resume feature. Either way, as a matter of user-friendliness, we should add a "best before" date to frozen iterators.

We're of course open for any kind of improvements on this in the future. As of now, I think the trade-off is currently between implementation simplicity, and as such risk of errors and unreliability, versus the longevity of frozen iterators (file size is typically no concern while within the magnitude of 10 kB). The current freeze/thaw mechanism is very simple. Maybe I'm overestimating the complexity, but at least one problem I can think of with storing pagination tokens only is that it requires some non-trivial handling of already-consumed nodes (maybe by storing their hashes and omitting equally-hashed nodes at the beginning resuming?). Storing an index number of the next item is obviously not sufficient, since that could lead to missing nodes when previous nodes have been deleted in the meantime.

tonycpsu:
My solution to the "already-consumed nodes" issue was giving the caller control over the batch/page size, and making it their job to consume the iterator.  This may not work for all use cases, but it worked for mine.

I've never had a pagination token expire on me, for what it's worth, and I had this same resume functionality implemented using [a different IG API](https://github.com/ping/instagram_private_api/) for many months before switching to this one.

In any event, this is your project, and you clearly know more about the tradeoffs than I do, so I'm not really interested in prolonging the conversation.  I implemented #683 as the most expedient means to an end, storing only the `end_cursor` token, and have had no issues with it.  I will continue using it until there's a compelling reason to switch.  Thanks for all the work you've put in on this excellent project.

aandergr:
Merged. Thanks for your comments and this discussion, @tonycpsu!

PR #743, opened 2020-07-26
Facebook Group Posts Downloader Request

Sherlock21B0:
NOT sure if its the right place to ask this. But do you have anything similar for FB? Where one can download all posts of an FB group along with its captions?


aandergr:
No, Instaloader downloads from Instagram only.

Issue #739, opened 2020-07-22
Add the option to only download the first, last or x image from an account


Akashic101:
**Provide us a use case of the feature**
Many comic-artists post their comics slide-by-slide and provide a combined picture of all together as the last picture ([example](https://www.instagram.com/p/CC3x1p9Jx--/)). Other artists post their creative work on the first slide since it's too wide and cut the image in half and post those after that ([example](https://www.instagram.com/p/CCYHAIkhvcg/)). If the user could specify which slide to download it could save a lot of time instead of manually filtering them out after download

**Describe the solution you'd like**
An argument like `--slide` could be added which takes the parameter `first`, `last` or a number, for example 2 would only download the second slide 

**Describe alternatives you've considered**
The only alternative would be to do filter out the slides manually after the download is finish which is very tedious since many artists have hundreds of comic

**If the feature request is accepted, would you be willing to submit a PR?**
Yes

**Additional context**
There is no additional context

aandergr:
We think that's a cool feature and we'll be happy to see a PR for this.

Just for clarification of terms:

> slide-by-slide

We assume, the *slide-by-slide* pictures are what we call *sidecars* internally in Instaloader (and in the Instagram's API response), and this would affect the downloading of sidecars. Sidecars are posts that contain multiple media (pictures or videos) bundled into one post object.

> An argument like `--slide` could be added which takes the parameter `first`, `last` or a number

What is `first`? Is it the first media, i.e. the one with number=1? Or is it the slide that is pre-selected, i.e. the slide that is shown before the user makes any swipe on the picture (which is not necessarily the first picture)?

What happens, if a sidecar does not have as many pictures as selected by the `--slide` parameter?

Akashic101:
With `first` I'd expect to download the picture with number=1, so the slide on the furthest left. `last` would in that case be the picture the furthest to the right

If a sidecar has for example only 3 images but the user used --slide 4 I'd say the post will be skipped and the download will continue with the next sidecar. 

I haven't programmed any of this features in since Python is not a language I'd call myself good in. That was on me, I mistook the meaning of PR

aandergr:
> With `first` I'd expect to download the picture with number=1, so the slide on the furthest left. `last` would in that case be the picture the furthest to the right
> 
> If a sidecar has for example only 3 images but the user used --slide 4 I'd say the post will be skipped and the download will continue with the next sidecar.

I think that makes sense.

[...]

PR #749, opened 2020-08-01
Implementation of feature #739

AndreaRe9:
As said in the title, this pull request is about a **feature implementation** of issue #739. The new change allows to **select which image** or **range of images** to download in an Instagram sidecar.

## Changes 
There's a new parameter, `--slide`, that by default is an empty string. If setted, it allows to:
* download just *one image* with the `index` specified
* download an interval of images, using this format: `start_index`:`end_index`, for example `1:3` (endpoints included)

The core of the new feature is almost entirely implemented inside the `download_post()` method.

### Expected behavior
The index starts from 1, where 1 is the leftmost image in the sidecar. The index numbers can't be negative. If the index/interval is invalid, for example in case you are out of range, the download of the post is *skipped*

## The completeness of this change
* **Is it just a proof of concept?** Yes, it is. Future improvements to this PR will certainly come, but I need to better understand if the purpose of the feature requested on #739 is reached.
* **Is the documentation updated (if appropriate)?** No, it's not.
* **Do you consider it ready to be merged or is it a draft?** As said above, it is not ready to be merged.
* **Can we help you at some point?** Yes, I think so. I would like to have opinions and reviews on the PR, to better understand if this PoC is sufficiently good in order to reach the final purpose of #739.

Please note that I have only tested the feature using Instaloader as a *Python module*, not as a standalone executable.

### DIFF
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -87,8 +87,8 @@ What to Download of each Post
    Download only selected images of a sidecar. You can select single images using their
    index in the sidecar starting with the leftmost or you can specify a range of images
    with the following syntax: ``start_index-end_index``. Example:
-   ``--slide 1`` will select only the first image and ``--slide 1-3`` will select only
-   the first three images.
+   ``--slide 1`` will select only the first image, ``--slide -1`` only the last one and ``--slide 1-3`` will select only
+   the first three images. 
 
 .. option:: --no-metadata-json
 
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -84,7 +84,11 @@ What to Download of each Post
 
 .. option:: --slide
 
-   Download only selected images of a sidecar.
+   Download only selected images of a sidecar. You can select single images using their
+   index in the sidecar starting with the leftmost or you can specify a range of images
+   with the following syntax: start_index-end_index. Example:
+   ``--slide 1`` will select only the first image and ``--slide 1:3`` will select only
+   the first three images.
 
 .. option:: --no-metadata-json
 
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -87,7 +87,7 @@ What to Download of each Post
    Download only selected images of a sidecar. You can select single images using their
    index in the sidecar starting with the leftmost or you can specify a range of images
    with the following syntax: ``start_index-end_index``. Example:
-   ``--slide 1`` will select only the first image and ``--slide 1:3`` will select only
+   ``--slide 1`` will select only the first image and ``--slide 1-3`` will select only
    the first three images.
 
 .. option:: --no-metadata-json
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -87,8 +87,8 @@ What to Download of each Post
    Download only selected images of a sidecar. You can select single images using their
    index in the sidecar starting with the leftmost or you can specify a range of images
    with the following syntax: ``start_index-end_index``. Example:
-   ``--slide 1`` will select only the first image, ``--slide -1`` only the last one and ``--slide 1-3`` will select only
-   the first three images. 
+   ``--slide 1`` will select only the first image, ``--slide last`` only the last one and ``--slide 1-3`` will select only
+   the first three images.
 
 .. option:: --no-metadata-json
 
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -86,7 +86,7 @@ What to Download of each Post
 
    Download only selected images of a sidecar. You can select single images using their
    index in the sidecar starting with the leftmost or you can specify a range of images
-   with the following syntax: start_index-end_index. Example:
+   with the following syntax: ``start_index-end_index``. Example:
    ``--slide 1`` will select only the first image and ``--slide 1:3`` will select only
    the first three images.
 
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -90,6 +90,8 @@ What to Download of each Post
    ``--slide 1`` will select only the first image, ``--slide last`` only the last one and ``--slide 1-3`` will select only
    the first three images.
 
+   .. versionadded:: 4.6
+
 .. option:: --no-metadata-json
 
    Do not create a JSON file containing the metadata of each post.
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -82,6 +82,10 @@ What to Download of each Post
    Template to write in txt file for each StoryItem. See
    :ref:`metadata-text-files`.
 
+.. option:: --slide
+
+   Download only selected images of a sidecar.
+
 .. option:: --no-metadata-json
 
    Do not create a JSON file containing the metadata of each post.
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -268,6 +268,8 @@ def main():
                         help="Do not download regular posts.")
     g_prof.add_argument('--no-profile-pic', action='store_true',
                         help='Do not download profile picture.')
+    g_prof.add_argument('--slide', action='store',
+                        help='Set what image/interval of a sidecar you want to download')
     g_post.add_argument('--no-pictures', action='store_true',
                         help='Do not download post pictures. Cannot be used together with --fast-update. '
                              'Implies --no-video-thumbnails, does not imply --no-videos.')
@@ -425,7 +427,8 @@ def main():
                              max_connection_attempts=args.max_connection_attempts,
                              request_timeout=args.request_timeout,
                              resume_prefix=resume_prefix,
-                             check_resume_bbd=not args.use_aged_resume_files)
+                             check_resume_bbd=not args.use_aged_resume_files,
+                             slide=args.slide)
         _main(loader,
               args.profile,
               username=args.login.lower() if args.login is not None else None,
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -269,7 +269,7 @@ def main():
     g_prof.add_argument('--no-profile-pic', action='store_true',
                         help='Do not download profile picture.')
     g_prof.add_argument('--slide', action='store',
-                        help='Set what image/interval of a sidecar you want to download')
+                        help='Set what image/interval of a sidecar you want to download.')
     g_post.add_argument('--no-pictures', action='store_true',
                         help='Do not download post pictures. Cannot be used together with --fast-update. '
                              'Implies --no-video-thumbnails, does not imply --no-videos.')
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -268,7 +268,7 @@ def main():
                         help="Do not download regular posts.")
     g_prof.add_argument('--no-profile-pic', action='store_true',
                         help='Do not download profile picture.')
-    g_prof.add_argument('--slide', action='store',
+    g_post.add_argument('--slide', action='store',
                         help='Set what image/interval of a sidecar you want to download.')
     g_post.add_argument('--no-pictures', action='store_true',
                         help='Do not download post pictures. Cannot be used together with --fast-update. '
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -212,21 +212,27 @@ class Instaloader:
         if self.slide != "":
             splitted = self.slide.split('-')
             if len(splitted) == 1:
-                self.slide_start = self.slide_end = int(splitted[0])
-                if self.slide_start <= 0:
-                    raise InvalidArgumentException("--slide parameter cannot be negative.")
-            elif len(splitted) == 2:
-                if splitted[0] == '':
-                    #download last image of a sidecar
+                if splitted[0] == 'last':
+                    #download only last image of a sidecar
                     self.slide_end = -1
-                elif int(splitted[0]) < int(splitted[1]) and int(splitted[0]) > 0:
+                else:
+                    if int(splitted[0]) > 0:
+                        self.slide_start = self.slide_end = int(splitted[0])
+                    else:
+                        raise InvalidArgumentException("--slide parameter must be greather than 0.")
+            elif len(splitted) == 2:
+                if splitted[1] == 'last':
+                    self.slide_start = int(splitted[0])
+                    self.slide_end = -1
+                elif 0 < int(splitted[0]) < int(splitted[1]):
                     self.slide_start = int(splitted[0])
                     self.slide_end = int(splitted[1])
                 else:
                     raise InvalidArgumentException("Invalid data for --slide parameter.")
             else:
                 raise InvalidArgumentException("Invalid data for --slide parameter.")
-
+        else:
+            self.slide_start = 1
     @contextmanager
     def anonymous_copy(self):
         """Yield an anonymous, otherwise equally-configured copy of an Instaloader instance; Then copy its error log."""
@@ -539,14 +545,7 @@ class Instaloader:
         if self.download_pictures:
             if post.typename == 'GraphSidecar':
                 edge_number = 1
-                for sidecar_node in post.get_sidecar_nodes():
-                    if self.slide_start + self.slide_end > 0:
-                        if not self.slide_start <= edge_number <= self.slide_end:
-                            continue
-                    if self.slide_start + self.slide_end < 0:
-                        # Download only last sidecar node
-                        if not sidecar_node.is_last:
-                            continue
+                for sidecar_node in post.get_sidecar_nodes(self.slide_start, self.slide_end):
                     # Download picture or video thumbnail
                     if not sidecar_node.is_video or self.download_video_thumbnails is True:
                         downloaded &= self.download_pic(filename=filename, url=sidecar_node.display_url,
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -556,7 +556,7 @@ class Instaloader:
         if post.typename == 'GraphSidecar':
             if self.download_pictures or self.download_videos:
                 for edge_number, sidecar_node in enumerate(post.get_sidecar_nodes(self.slide_start,
-                                                                              self.slide_end), start=1):
+                                                                                  self.slide_end), start=1):
                     if self.download_pictures and (not sidecar_node.is_video or self.download_video_thumbnails):
                         suffix = str(edge_number)
                         if '{filename}' in self.filename_pattern:
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -217,15 +217,15 @@ class Instaloader:
                     self.slide_start = -1
                 else:
                     if int(splitted[0]) > 0:
-                        self.slide_start = self.slide_end = int(splitted[0])
+                        self.slide_start = self.slide_end = int(splitted[0])-1
                     else:
                         raise InvalidArgumentException("--slide parameter must be greater than 0.")
             elif len(splitted) == 2:
                 if splitted[1] == 'last':
-                    self.slide_start = int(splitted[0])
+                    self.slide_start = int(splitted[0])-1
                 elif 0 < int(splitted[0]) < int(splitted[1]):
-                    self.slide_start = int(splitted[0])
-                    self.slide_end = int(splitted[1])
+                    self.slide_start = int(splitted[0])-1
+                    self.slide_end = int(splitted[1])-1
                 else:
                     raise InvalidArgumentException("Invalid data for --slide parameter.")
             else:
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -556,7 +556,7 @@ class Instaloader:
         if post.typename == 'GraphSidecar':
             if self.download_pictures or self.download_videos:
                 for edge_number, sidecar_node in enumerate(post.get_sidecar_nodes(self.slide_start,
-                                                                                  self.slide_end), start=1):
+                                                                                  self.slide_end), start=self.slide_start+1):
                     if self.download_pictures and (not sidecar_node.is_video or self.download_video_thumbnails):
                         suffix = str(edge_number)
                         if '{filename}' in self.filename_pattern:
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -220,7 +220,7 @@ class Instaloader:
                 self.slide_start = int(splitted[0])
                 self.slide_end = int(splitted[1])
             else:
-                raise InvalidArgumentException("Invalid data for --slide parameter: ".format(slide))
+                raise InvalidArgumentException("Invalid data for --slide parameter: ")
 
     @contextmanager
     def anonymous_copy(self):
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -211,16 +211,21 @@ class Instaloader:
         self.slide_end = 0
         if self.slide != "":
             splitted = self.slide.split('-')
-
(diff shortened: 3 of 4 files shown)


AndreaRe9:
Thank you @aandergr for your review! 
I think I have solved all the issues about PEP8 and documentation too. Even if the last commit have passed all the checks, I would like to know if the last modifications are in line with what you suggested.

instaloader/instaloader.py
```diff
@@ -204,6 +206,26 @@ def __init__(self,
...
 
+        self.slide = slide or ""
+        self.slide_start = 0
+        self.slide_end = 0
+        if self.slide != "":
+            splitted = self.slide.split('-')
+            try:
+                if len(splitted) == 1:
+                    self.slide_start = self.slide_end = int(splitted[0])
+                    if self.slide_start < 0:
```
aandergr, line 217:
```suggestion
                    if self.slide_start <= 0:
```

slides start counting at 1.

instaloader/instaloader.py
```diff
@@ -204,6 +206,26 @@ def __init__(self,
...
+                        raise InvalidArgumentException("Invalid data for --slide parameter")
```
aandergr, line 218: We can be more specific here, i.e. something like "slide parameter is negative"

instaloader/instaloader.py
```diff
@@ -204,6 +206,26 @@ def __init__(self,
...
+                else:
+                    if int(splitted[0]) < int(splitted[1]) and int(splitted[0]) > 0:
+                        self.slide_start = int(splitted[0])
+                        self.slide_end = int(splitted[1])
+                    else:
+                        raise InvalidArgumentException("Invalid data for --slide parameter")
```
aandergr, line 224: We could easily catch the case of `len(splitted) > 2` here, by nesting the if/else as follows:

```suggestion
                elif len(splitted) == 2 and int(splitted[0]) < int(splitted[1]) and int(splitted[0]) > 0:
                    self.slide_start = int(splitted[0])
                    self.slide_end = int(splitted[1])
                else:
                    raise InvalidArgumentException("Invalid data for --slide parameter: ".format(slide))
```

instaloader/instaloader.py
```diff
@@ -517,6 +539,9 @@ def download_post(self, post: Post, target: Union[str, Path]) -> bool:
             if post.typename == 'GraphSidecar':
                 edge_number = 1
                 for sidecar_node in post.get_sidecar_nodes():
+                    if self.slide_start + self.slide_end != 0:
+                        if edge_number < self.slide_start or edge_number > self.slide_end:
```
aandergr, line 543:
```suggestion
                        if not self.slide_start <= edge_number <= self.slide_end:
```

Is syntactically easier / more intuitive.

instaloader/instaloader.py
```diff
@@ -525,6 +550,7 @@ def download_post(self, post: Post, target: Union[str, Path]) -> bool:
                     if sidecar_node.is_video and self.download_videos is True:
                         downloaded &= self.download_pic(filename=filename, url=sidecar_node.video_url,
                                                         mtime=post.date_local, filename_suffix=str(edge_number))
+                    self.context.log("Downloading instagram.com/p/{} with index {}".format(post.shortcode, edge_number))
```
aandergr, line 553: We do not need these messages and it breaks the formatting of the output.

instaloader/instaloader.py
```diff
@@ -182,7 +183,8 @@ def __init__(self,
                  request_timeout: Optional[float] = None,
                  rate_controller: Optional[Callable[[InstaloaderContext], RateController]] = None,
                  resume_prefix: Optional[str] = "iterator",
-                 check_resume_bbd: bool = True):
+                 check_resume_bbd: bool = True,
+                 slide: Optional[str] = ""):
```
aandergr, line 187:
```suggestion
                 slide: Optional[str] = None):
```

docs/cli-options.rst
```diff
@@ -82,6 +82,14 @@ What to Download of each Post
    Template to write in txt file for each StoryItem. See
    :ref:`metadata-text-files`.
 
+.. option:: --slide
+
+   Download only selected images of a sidecar. You can select single images using their
+   index in the sidecar starting with the leftmost or you can specify a range of images
+   with the following syntax: start_index-end_index. Example:
+   ``--slide 1`` will select only the first image and ``--slide 1:3`` will select only
```
aandergr, line 90:
```suggestion
   ``--slide 1`` will select only the first image and ``--slide 1-3`` will select only
```

docs/cli-options.rst
```diff
@@ -82,6 +82,14 @@ What to Download of each Post
    Template to write in txt file for each StoryItem. See
    :ref:`metadata-text-files`.
 
+.. option:: --slide
+
+   Download only selected images of a sidecar. You can select single images using their
+   index in the sidecar starting with the leftmost or you can specify a range of images
+   with the following syntax: start_index-end_index. Example:
```
aandergr, line 89:
```suggestion
   with the following syntax: ``start_index-end_index``. Example:
```

instaloader/instaloader.py
```diff
@@ -204,6 +206,22 @@ def __init__(self,
...
+
+            if len(splitted) == 1:
+                self.slide_start = self.slide_end = int(splitted[0])
+                if self.slide_start <= 0:
+                    raise InvalidArgumentException("--slide parameter cannot be negative")
+            elif len(splitted) == 2 and int(splitted[0]) < int(splitted[1]) and int(splitted[0]) > 0:
+                self.slide_start = int(splitted[0])
+                self.slide_end = int(splitted[1])
+            else:
+                raise InvalidArgumentException("Invalid data for --slide parameter: ")
```
aandergr, line 223: This prints the message `Invalid data for --slide parameter: ` without that anything follows the colon. Either add the `self.slide` input to the message, or replace the colon by a period.

AndreaRe9: Yeah, my bad.

instaloader/structures.py
```diff
@@ -11,11 +11,12 @@
 from .instaloadercontext import InstaloaderContext
 from .nodeiterator import FrozenNodeIterator, NodeIterator
 
-PostSidecarNode = namedtuple('PostSidecarNode', ['is_video', 'display_url', 'video_url'])
+PostSidecarNode = namedtuple('PostSidecarNode', ['is_video', 'display_url', 'video_url','is_last'])
```
AndreaRe9: I tought that it was cleaner to add a property to the PostSidecar structure, instead of figuring out inside the `download_post()` function when the last one is reached.

AndreaRe9:
Hi @aandergr!
I have fixed the things you suggested and I've also implemented the feature that was missing. Now, using `--slide -1`, the user is able to download only the last node of a sidecar.

instaloader/instaloader.py
```diff
@@ -204,6 +206,27 @@ def __init__(self,
...
+            splitted = self.slide.split('-')
+            if len(splitted) == 1:
+                self.slide_start = self.slide_end = int(splitted[0])
+                if self.slide_start <= 0:
+                    raise InvalidArgumentException("--slide parameter cannot be negative.")
+            elif len(splitted) == 2:
+                if splitted[0] == '':
+                    #download last image of a sidecar
+                    self.slide_end = -1
+                elif int(splitted[0]) < int(splitted[1]) and int(splitted[0]) > 0:
```
aandergr, line 222:
```suggestion
                elif 0 < int(splitted[0]) < int(splitted[1]):
```

instaloader/structures.py
```diff
@@ -11,11 +11,12 @@
 from .instaloadercontext import InstaloaderContext
 from .nodeiterator import FrozenNodeIterator, NodeIterator
 
-PostSidecarNode = namedtuple('PostSidecarNode', ['is_video', 'display_url', 'video_url'])
+PostSidecarNode = namedtuple('PostSidecarNode', ['is_video', 'display_url', 'video_url', 'is_last'])
```
aandergr, line 14: The field `is_last` does not make any sense here. It would be better to let the caller of `Post.get_sidecar_nodes()` determine which node is the last one.

instaloader/structures.py
```diff
@@ -251,30 +250,31 @@ def typename(self) -> str:
         """Type of post, GraphImage, GraphVideo or GraphSidecar"""
         return self._field('__typename')
 
-    def get_sidecar_nodes(self) -> Iterator[PostSidecarNode]:
+    def get_sidecar_nodes(self, start=0, end=0) -> Iterator[PostSidecarNode]:
```
AndreaRe9, line 253: Instead of getting all the nodes, I've added here `start` and `end` parameters, by default setted to 0. 
I've decided to do the job of selecting the posts here, because in this function I have access to the `edges`. Here, I know how many images are inside the post and therefore I can easily work with indexes.

instaloader/structures.py
```diff
@@ -251,30 +250,31 @@ def typename(self) -> str:
...
         """Sidecar nodes of a Post with typename==GraphSidecar."""
         if self.typename == 'GraphSidecar':
             edges = self._field('edge_sidecar_to_children', 'edges')
             if any(edge['node']['is_video'] for edge in edges):
                 # video_url is only present in full metadata, issue #558.
                 edges = self._full_metadata['edge_sidecar_to_children']['edges']
+            if end <= 0:
+                end = len(edges)
+            if start == 0:
```
AndreaRe9, line 262: Basically, if start is equal to 0, it means that the user has selected only the last image, because in all the other cases start is not zero. So, in order to download the last image, `start = end = len(edges)`.

instaloader/structures.py
AndreaRe9: I've tried a different approach: do you thing it's better?

instaloader/__main__.py
```diff
@@ -268,6 +268,8 @@ def main():
                         help="Do not download regular posts.")
     g_prof.add_argument('--no-profile-pic', action='store_true',
                         help='Do not download profile picture.')
+    g_prof.add_argument('--slide', action='store',
```
aandergr, line 271:
```suggestion
    g_post.add_argument('--slide', action='store',
```

This belongs to group *What to download of each Post* rather than *What to Download of each Profile*.

instaloader/instaloader.py
```diff
@@ -204,6 +206,33 @@ def __init__(self,
...
+            splitted = self.slide.split('-')
+            if len(splitted) == 1:
+                if splitted[0] == 'last':
+                    #download only last image of a sidecar
+                    self.slide_end = -1
+                else:
+                    if int(splitted[0]) > 0:
+                        self.slide_start = self.slide_end = int(splitted[0])
+                    else:
+                        raise InvalidArgumentException("--slide parameter must be greather than 0.")
```
aandergr, line 222:
```suggestion
                        raise InvalidArgumentException("--slide parameter must be greater than 0.")
```

(spelling)

instaloader/instaloader.py
```diff
@@ -204,6 +206,33 @@ def __init__(self,
...
         self.check_resume_bbd = check_resume_bbd
 
+        self.slide = slide or ""
+        self.slide_start = 0
+        self.slide_end = 0
+        if self.slide != "":
+            splitted = self.slide.split('-')
+            if len(splitted) == 1:
+                if splitted[0] == 'last':
+                    #download only last image of a sidecar
```
aandergr, line 216:
```suggestion
                    # download only last image of a sidecar
```

(whitespace)

instaloader/instaloader.py
```diff
@@ -204,6 +206,33 @@ def __init__(self,
         self.resume_prefix = resume_prefix
         self.check_resume_bbd = check_resume_bbd
 
+        self.slide = slide or ""
+        self.slide_start = 0
+        self.slide_end = 0
```
aandergr, line 211: If I understand this if-nesting right, we have the following cases:

1. no `--slide` parameter: *start=1, end=0*, however *start=0, end=0* is the default for get_sidecar_posts(),
1. `--slide NUMBER`: *start=NUMBER, end=NUMBER*,
1. `--slide last`: *start=0, end=-1*,
1. `--slide A-B`, *A* and *B* both numbers: *start=A, end=B*,
1. `--slide NUMBER-last`: *start=NUMBER, end=-1*.

I find the meaning of 0 extremely confusing, i.e. the consistency especially regarding (1) and (3). I also suspect here, that 0 has a different meaning for *start* than for *end*. Isn't there a simpler and less confusing way?

instaloader/structures.py
```diff
@@ -250,26 +250,32 @@ def typename(self) -> str:
...
+                start = len(edges)
```
aandergr, line 263: So, with the default arguments, `get_sidecar_nodes()` would only return the last slide instead of all slides? `get_sidecar_nodes()` should return all sidecar nodes if `start` and `end` parameters are not given.

instaloader/instaloader.py
```diff
@@ -204,6 +206,33 @@ def __init__(self,
...
+                    self.slide_end = -1
+                elif 0 < int(splitted[0]) < int(splitted[1]):
+                    self.slide_start = int(splitted[0])
+                    self.slide_end = int(splitted[1])
+                else:
+                    raise InvalidArgumentException("Invalid data for --slide parameter.")
+            else:
+                raise InvalidArgumentException("Invalid data for --slide parameter.")
+        else:
+            self.slide_start = 1
```
aandergr, line 235:
```suggestion
            self.slide_start = 1

```

(add missing empty line between two methods)

instaloader/structures.py
aandergr: I like that the slicing happens here and it seems cleanly implemented. However the default should return all sidecar nodes, we cannot change the default behavior of a function.

I think it would be much more intuitive if the *start* and *end* parameters were indices with the same meaning as in Python array slicing, i.e. *0* is the first item, *1* is the second item, *-1* the last, and so on. This would not require any further explanation as this is what Python developers are used to. Also the implementation is simple: The if comparison below would then be a simple `start <= idx <= end`, and a simple `start %= len(edges)`, `end %= len(edges)` would suffice to handle negative indices. *start* would default to 0, *end* to -1. I think this will also simplify the big if in the Instaloader constructor.

instaloader/instaloader.py
aandergr: (with my suggestion below at the get_sidecar_posts() comment, default (case (1)) becomes *start=0, end=-1*, and we would translate the string input like: *number -> number-1*, *last -> -1*. Case (3) would become *start=-1, end=-1* and as such just like (2) with *-1* instead of *NUMBER-1*)

instaloader/structures.py
```diff
@@ -250,16 +250,16 @@ def typename(self) -> str:
         """Type of post, GraphImage, GraphVideo or GraphSidecar"""
         return self._field('__typename')
 
-    def get_sidecar_nodes(self, start=0, end=0) -> Iterator[PostSidecarNode]:
+    def get_sidecar_nodes(self, start=0, end=-1) -> Iterator[PostSidecarNode]:
```
AndreaRe9, line 253: Here, we have `end=-1` by default, that means `end=last`

instaloader/instaloader.py
AndreaRe9: You are right @aandergr, we can definitely think on a less confusing way to use 0 and -1.

AndreaRe9 (commented):
In order to make things a little bit clearer, thanks to the suggestions of @aandergr, I have applied a couple of changes. Given the parameters `start` and `end`, that is what the user can do:

1. no `--slide` parameter ==> `start=0` and `end=-1`, download everything
2. `--slide last` ==> `start=-1` and `end=-1`, download just last
3. `--slide NUMBER` ==> `start=end=NUMBER`, download just one slide
4. `--slide NUMBER-last` ==> `start=NUMBER` and `end=-1`, download from `NUMBER` to last slide
5. `--slide NUMBER1-NUMBER2` ==> download slides starting from `NUMBER1` till `NUMBER2`

instaloader/structures.py
```diff
@@ -250,26 +250,32 @@ def typename(self) -> str:
...
         """Sidecar nodes of a Post with typename==GraphSidecar."""
         if self.typename == 'GraphSidecar':
             edges = self._field('edge_sidecar_to_children', 'edges')
             if any(edge['node']['is_video'] for edge in edges):
                 # video_url is only present in full metadata, issue #558.
                 edges = self._full_metadata['edge_sidecar_to_children']['edges']
+            if end < 0:
+                end = len(edges)
+            if start < 0:
+                start = len(edges)
```
aandergr, line 263:
```suggestion
            if end < 0:
                end = len(edges) - 1
            if start < 0:
                start = len(edges) - 1
```

(alternatively `%= len(edges)`)

instaloader/structures.py
```diff
@@ -250,26 +250,32 @@ def typename(self) -> str:
...
-                if not is_video and self._context.is_logged_in:
-                    try:
-                        carousel_media = self._iphone_struct['carousel_media']
-                        orig_url = carousel_media[idx]['image_versions2']['candidates'][0]['url']
-                        display_url = re.sub(r'&se=\d+(&?)', r'\1', orig_url)
-                    except (InstaloaderException, KeyError, IndexError) as err:
-                        self._context.error('{} Unable to fetch high quality image version of {}.'.format(err, self))
-                yield PostSidecarNode(is_video=is_video, display_url=display_url,
-                                      video_url=node['video_url'] if is_video else None)
+                if start-1 <= idx <= end-1:
```
aandergr, line 265:
```suggestion
                if start <= idx <= end:
```

aandergr (commented):
I rebased your branch to current `upcoming/v4.6`, removed by own commits from there, and force-pushed to your branch, to make it possible to review this.

> In order to make things a little bit clearer, thanks to the suggestions of @aandergr, I have applied a couple of changes. Given the parameters `start` and `end`, that is what the user can do:
> 
> 1. no `--slide` parameter ==> `start=0` and `end=-1`, download everything
> 2. `--slide last` ==> `start=-1` and `end=-1`, download just last
> 3. `--slide NUMBER` ==> `start=end=NUMBER`, download just one slide
> 4. `--slide NUMBER-last` ==> `start=NUMBER` and `end=-1`, download from `NUMBER` to last slide
> 5. `--slide NUMBER1-NUMBER2` ==> download slides starting from `NUMBER1` till `NUMBER2`

I still do not understand 3-5. I think, if *0* is the first slide and *-1* is the last slide, then `--slide NUMBER` should set `start` and `end` to *NUMBER-1*, likewise `--slide NUMBER-last` should set `start` to *NUMBER-1* and `end` to *-1*, and finally `--slide A-B` should set `start` to *A-1* and `end` to *B-1*.

AndreaRe9 (commented):
Hi @aandergr, is this what you meant?

aandergr:
This branch now contains commits which are completely unrelated to the here-discussed change. We cannot cleanly merge this, with these commits included. Please rebase this to `upcoming/v4.6` with no other commits included than yours.

aandergr:
@AndreaRe9 Have you given up on this?

AndreaRe9:
Sorry but I've been busy.
No, I haven't. I just need to figure out how to rebase this, because I have never done it.

AndreaRe9:
So, I have rebased on upcoming/v4.6 and created a new branch [here](/AndreaRe9/instaloader/tree/upcoming/v4.6/instaloader)
Would you mind to take a look @aandergr?

aandergr:
Thanks! Yes, on the first view, the branch [`AndreaRe9:upcoming/v4.6`](https://github.com/instaloader/instaloader/compare/upcoming/v4.6...AndreaRe9:upcoming/v4.6) looks good! I will give it a deeper look soon.

aandergr:
> I still do not understand 3-5. I think, if _0_ is the first slide and _-1_ is the last slide, then `--slide NUMBER` should set `start` and `end` to _NUMBER-1_, likewise `--slide NUMBER-last` should set `start` to _NUMBER-1_ and `end` to _-1_, and finally `--slide A-B` should set `start` to _A-1_ and `end` to _B-1_.

Has this somehow been taken care of? I don't see this addressed in the branch you referred.

Could you also push the rebased branch to the branch linked with this PR so that it lists the now-rebased version of your changes?

AndreaRe9:
Yes, I'll take care of it soon

AndreaRe9:
> > I still do not understand 3-5. I think, if _0_ is the first slide and _-1_ is the last slide, then `--slide NUMBER` should set `start` and `end` to _NUMBER-1_, likewise `--slide NUMBER-last` should set `start` to _NUMBER-1_ and `end` to _-1_, and finally `--slide A-B` should set `start` to _A-1_ and `end` to _B-1_.
> 
> Has this somehow been taken care of? I don't see this addressed in the branch you referred.
> 
> Could you also push the rebased branch to the branch linked with this PR so that it lists the now-rebased version of your changes?

Hello @aandergr,
I have made 6e5a771d5a095f26c244e21548fde0c1bc351162 to add the changes and I have force pushed to the branch _upcoming/v4.5_. 
I'm not sure if this is what you meant. Is everything correct?

aandergr:
Thanks for continuing! However, there is still a conflict in a file, so we cannot merge it (the big green "merge" button is grey).

AndreaRe9:
No problem @aandergr, solved now.

instaloader/instaloader.py
```diff
@@ -514,16 +542,16 @@ def download_post(self, post: Post, target: Union[str, Path]) -> bool:
...
-                for edge_number, sidecar_node in enumerate(post.get_sidecar_nodes(), start=1):
-                    if self.download_pictures and (not sidecar_node.is_video or self.download_video_thumbnails):
-                        # Download sidecar picture or video thumbnail (--no-pictures implies --no-video-thumbnails)
-                        downloaded &= self.download_pic(filename=filename, url=sidecar_node.display_url,
-                                                        mtime=post.date_local, filename_suffix=str(edge_number))
-                    if sidecar_node.is_video and self.download_videos:
-                        # Download sidecar video if desired
-                        downloaded &= self.download_pic(filename=filename, url=sidecar_node.video_url,
-                                                        mtime=post.date_local, filename_suffix=str(edge_number))
+            for edge_number, sidecar_node in enumerate(post.get_sidecar_nodes(self.slide_start,
```
aandergr, line 545: Please do not remove the `if self.download_pictures or self.download_videos`.

instaloader/instaloader.py
```diff
@@ -527,7 +555,8 @@ def download_post(self, post: Post, target: Union[str, Path]) -> bool:
         downloaded = True
         if post.typename == 'GraphSidecar':
             if self.download_pictures or self.download_videos:
-                for edge_number, sidecar_node in enumerate(post.get_sidecar_nodes(), start=1):
+                for edge_number, sidecar_node in enumerate(post.get_sidecar_nodes(self.slide_start,
+                                                                                  self.slide_end), start=1):
```
aandergr, line 559:
```suggestion
                for edge_number, sidecar_node in enumerate(post.get_sidecar_nodes(self.slide_start,
                                                                                  self.slide_end), start=self.slide_start+1):
```

because otherwise the filenames are wrong. (`--slide last` currently names all files `..._1.jpg`).

aandergr (changes requested):
Just tried it and found a minor and easy to fix bug: The edge_number in the loop that downloads the post erroneously always starts at 1, so e.g. with `--slide last` each sidecar image  is named as if it was the first.

aandergr:
Thanks, I did some final fixes and merged it.

AndreaRe9:
Thank you for your patience @aandergr :smile:

Issue #769, opened 2020-08-13
--fast-update --highlights crash if cover.jpg exist


psonnosp:
**Describe the bug**
If you already have downloaded highlights from a profile and try to do it again the session crash because cover.jpg already exist. If you delete cover.jpg everything works as expected.

**To Reproduce**
.\instaloader profile(s) --fast-update --highlights --no-posts --no-profile-pic --no-metadata-json -l "username"
Then do it again when the highlight has new content.

**Expected behavior**
That it skips cover.jpg and download the new content.

**Error messages and tracebacks**
2020-02-08_21-14-02_UTC_cover.jpg Traceback (most recent call last):
  File "__main__.py", line 462, in <module>
  File "__main__.py", line 459, in __main
  File "__main__.py", line 429, in main
  File "__main__.py", line 203, in _main
  File "instaloader\instaloader.py", line 1149, in download_profiles
  File "instaloader\instaloader.py", line 63, in call
  File "instaloader\instaloader.py", line 707, in download_highlights
  File "instaloader\instaloader.py", line 429, in download_highlight_cover
  File "instaloader\instaloader.py", line 79, in call
  File "instaloader\instaloader.py", line 416, in download_title_pic
  File "instaloader\instaloadercontext.py", line 490, in write_raw
FileExistsError: [WinError 183] Cannot create a file when that file already exists: '2020-02-08_21-
14-02_UTC_cover.jpg.temp' -> '2020-02-08_21-14-02_UTC_cover.jpg'

**Instaloader version**
4.5

**Additional context**

bebunw:
i confirm the same bug, it does not update highlight because script stop after cover.jpg of highlights

psonnosp:
Will that commit also fix the same thing happening to profile_pic? Cause now all of a sudden I get the same thing there.

psonnosp:
Thank you!
And to answer myself it seem to have fixed them both.

aandergr:
Fixed in [Version 4.5.1](https://github.com/instaloader/instaloader/releases/tag/v4.5.1).

PR #775, opened 2020-08-16
Fixed issue #769 by using os.replace instead os.rename

lapourgagner:
It should work but I don't have any Windows machine to test it.


aandergr:
Thanks!

Issue #778, opened 2020-08-18
Username does not exist error on public profiles.


DMScoder:
**Describe the bug**
When running `instaloader.Profile.from_username()` with a logged in context and a username that exists, an error is returned saying the profile doesn't exist, and recommending that same exact profile. I have double checked to make sure the usernames are correct and that there isn't any hidden formatting. 

**To Reproduce**
Using the `instaloader.Profile.from_username(loader.context, username)` on a few dozen accounts, some should error out saying they don't exist, but showing the most similar account as exactly the same name. This occurs specifically when logged in.

When I then try this same username from the command line using `instaloader username` it works fine 

**Expected behavior**
When getting the profile using the from_username() function, it should always return a user if that user's profile exists and is public.

**Error messages and tracebacks**
`[ERROR] with instaloader username uniliga_gg: Profile uniliga_gg does not exist.
The most similar profile is: uniliga_gg.`

**Instaloader version**
4.4.5

**Additional context**
The profiles that don't work seem to work fine from the commandline.

threnbo:
Same here. Same error on public and private accounts (that I am a comfirmed follower of). I created a new 'test' account and tried to scrape the same public accounts - no errors. I changed the username/name/birthday of my 'real' account and checked again, nothing changed, still the same error. I am glad if I could provide more information to solve the problem. Thanks!

DMScoder:
> The most likely explanation for the described behavior is that the profile in question has blocked the account you are using. Is it possible to view the profile with the Instagram App using the same account?
> 
> > When I then try this same username from the command line using `instaloader username` it works fine
> 
> Using the CLI, Instaloader accesses a blocked profile anonymously.
> 
> `instaloader.Profile.from_username(loader.context, 'uniliga_gg')` works fine for me.

So I cycle several accounts for scraping but have not had trouble accessing the non-existing users from the app. The non-existing accounts have no relation to the accounts I am logging in with, so if there is a block in place it is by instagram itself. Is it possible for accounts to be blocked on a region basis?

Looking at @Drod917's edits, my current workaround is to check if a recommended username is the same as the one in question, and scrape it anonymously if it is.

mbande:
exactly the same problem, i'm nearly sure that my accounts is not blocked by target profiles. so there should be something else related to account or target profile

mbande:
[...]

PR #785, opened 2020-08-28
Fix re.search

naohide:
- Fixes # https://github.com/instaloader/instaloader/blob/master/instaloader/instaloadercontext.py#L354-L355.
- What problem does the pull request solve?
    - https://github.com/instaloader/instaloader/issues/778
    - regex

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -351,10 +351,10 @@ class InstaloaderContext:
                     raise ConnectionException("\"window._sharedData\" does not contain required keys.")
                 # If GraphQL data is missing in `window._sharedData`, search for it in `__additionalDataLoaded`.
                 if 'graphql' not in post_or_profile_page[0]:
-                    match = re.search(r'window\.__additionalDataLoaded\([^{]+.*"graphql":({.*}).*}\);<\/script>',
+                    match = re.search(r'window\.__additionalDataLoaded\(.*?({.*"graphql":.*})\);</script>',
                                       resp.text)
                     if match is not None:
-                        post_or_profile_page[0]['graphql'] = json.loads(match.group(1))
+                        post_or_profile_page[0]['graphql'] = json.loads(match.group(1))['graphql']
                 return resp_json
             else:
                 resp_json = resp.json()
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -351,7 +351,7 @@ class InstaloaderContext:
                     raise ConnectionException("\"window._sharedData\" does not contain required keys.")
                 # If GraphQL data is missing in `window._sharedData`, search for it in `__additionalDataLoaded`.
                 if 'graphql' not in post_or_profile_page[0]:
-                    match = re.search(r'window\.__additionalDataLoaded\([^{]+{"graphql":({.*})}\);</script>',
+                    match = re.search(r'window\.__additionalDataLoaded\([^{]+.*"graphql":({.*}).*}\);<\/script>',
                                       resp.text)
                     if match is not None:
                         post_or_profile_page[0]['graphql'] = json.loads(match.group(1))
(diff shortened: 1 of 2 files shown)


instaloader/instaloadercontext.py
```diff
@@ -351,7 +351,7 @@ def get_json(self, path: str, params: Dict[str, Any], host: str = 'www.instagram
                     raise ConnectionException("\"window._sharedData\" does not contain required keys.")
                 # If GraphQL data is missing in `window._sharedData`, search for it in `__additionalDataLoaded`.
                 if 'graphql' not in post_or_profile_page[0]:
-                    match = re.search(r'window\.__additionalDataLoaded\([^{]+{"graphql":({.*})}\);</script>',
+                    match = re.search(r'window\.__additionalDataLoaded\([^{]+.*"graphql":({.*}).*}\);<\/script>',
                                       resp.text)
                     if match is not None:
                         post_or_profile_page[0]['graphql'] = json.loads(match.group(1))
```
Thammus, line 357:
```suggestion
                    match = re.search(r'window\.__additionalDataLoaded\(.*({"graphql":.*})\);</script>',
                                      resp.text)
                    if match is not None:
                        post_or_profile_page[0]['graphql'] = json.loads(match.group(1))['graphql']
```

Thammus (changes requested):
Thanks for the Pull Request! I suggest a simpler regex that should also work. I also hope that it is more reliable but I am unable to test it against the changed JSON data described in #778 since I have not faced this problem yet.

instaloader/instaloadercontext.py
naohide: @Thammus Thanks your comment. But, `({"` seems not correct. `"graphql"` doesn't appear at first. So, should be `({.*"`?

mbande:
@naohide I'm seeing the bug even after your PR

naohide:
@mbande please show me your html including graphql.

mbande:
@naohide my mistake, everything works as expected.

instaloader/instaloadercontext.py
Thammus: Ah, okay. You are right. If `"graphql"` is not the first key in the object, additional adjustments to the regex are necessary in order to match correctly. Simply adjusting `({.*"` as you suggest could cause other problems though. Consider the regular expression 
```regexp
window\.__additionalDataLoaded\(.*({.*"graphql":.*})\);</script>
```
and the test string
```javascript
window.__additionalDataLoaded(path, {"first_key": {"something": 123}, "graphql": {"foo": "bar", "baz": {"foo": "bar"}}, "something_different": {"foo": "baz"}});</script>
```
then just the string
```json
{"something": 123}, "graphql": {"foo": "bar", "baz": {"foo": "bar"}}, "something_different": {"foo": "baz"}}
```
is matched, which is not syntactically correct JSON. However, this can be corrected by making the very first `.*` in the regex "lazy" i.e. "non-greedy" by using `.*?` instead.

Thammus:
```suggestion
                    match = re.search(r'window\.__additionalDataLoaded\(.*?({.*"graphql":.*})\);</script>',
                                      resp.text)
                    if match is not None:
                        post_or_profile_page[0]['graphql'] = json.loads(match.group(1))['graphql']
```

naohide: @Thammus ok. thanks👍  I fixed.

naohide:
@aandergr please merge this.

instaloader/instaloadercontext.py
VasiliPupkin256: If you are using `json.loads` just access the `json.loads(match.group('json'))['graphql']` field from the parsed json object and use a simple and straightforward regex `r'window\.__additionalDataLoaded\(.*?,(?P<json>.*)\);</script>'` to extract the json data. There is no need to parse json data and extract its field in a complex regex.

PR #809, opened 2020-09-17
Avoid additional http request to obtain `owner_id`

ingo-m:
I noticed that a post's `owner_id` is obtained from the respective `owner_profile`. If the `owner_profile` has not been fetched yet, an additional http request is made. However, if the `Post` instance has been created from a `hashtag.get_posts()` iterator, the `owner_id` is already available, and the additional http request is unnecessary. As far as I can see, the PR avoids this unnecessary (and costly) http call.

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -207,7 +207,13 @@ class Post:
     @property
     def owner_id(self) -> int:
         """The ID of the Post's owner."""
-        return self.owner_profile.userid
+        # The ID may already be available, e.g. if the post instance was created
+        # from an `hashtag.get_posts()` iterator, so no need to make another
+        # http request.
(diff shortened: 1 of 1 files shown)


ingo-m:
Oh sorry seems like I made changes to `master` but compared against `upcoming/v4.6`; should I close and create a new PR where I make changes to `upcoming/v4.6`?

aandergr (approved):
Nice, thanks!

> Oh sorry seems like I made changes to `master` but compared against `upcoming/v4.6`; should I close and create a new PR where I make changes to `upcoming/v4.6`?

No problem, I merge it directly into master, which I think is okay for little performance improvements that do not require extensive testing.

ingo-m:
Great, thanks!

Issue #710, opened 2020-06-29
Highlights ending with ... has a missing folder/file error


mlalcevski:
**Describe the bug**
One of the highlights is labeled "Something ...", the "..." does not work as a folder name in windows so it gets an error every time.

**To Reproduce**
Download any highlights ending in ...

**Expected behavior**
Well rename the folder and ignore the ..

**Error messages and tracebacks**
[Errno 2] No such file or directory

**Instaloader version**
4.4.4

**Additional context**

aandergr:
Thanks for reporting. Indeed, creating a folder containing ".." or "..." in the name does not always work on Windows.

> **Expected behavior**
> Well rename the folder and ignore the ..

Maybe the dots in "..." should be replaced by another character that is not invalid in filenames and does not collide with characters that might occur in highlight titles, such as "-". There is also #576 with an idea of how we could approach this.

psonnosp:
It just does not always work, it never works with dots at the end of a folder name in Windows.
I know nothing about whats talked about in that other issue, but is it a complicated thing to make Instaloader ignore dots or use _ instead in the ending of folder names? Cause this seems like kind of a basic thing to handle for a program that runs in Windows.

psonnosp:
If this can't be fixed by replacing the dots would it be possible to make instaloader at least continue and not just quit with this error?
...\2019-05-23_11-39-37_UTC_cover.jpg Traceback (most recent call last):
  File "__main__.py", line 462, in <module>
  File "__main__.py", line 459, in __main
  File "__main__.py", line 429, in main
  File "__main__.py", line 203, in _main
  File "instaloader\instaloader.py", line 1150, in download_profiles
  File "instaloader\instaloader.py", line 63, in call
  File "instaloader\instaloader.py", line 708, in download_highlights
  File "instaloader\instaloader.py", line 429, in download_highlight_cover
  File "instaloader\instaloader.py", line 79, in call
  File "instaloader\instaloader.py", line 416, in download_title_pic
  File "instaloader\instaloadercontext.py", line 485, in write_raw
FileNotFoundError: [Errno 2] No such file or directory: ...\\2019-05-23_11-39-37_UTC_cover.jpg.temp'
[18116] Failed to execute script __main__

Issue #576, opened 2020-04-17
Gracefully handle invalid directories and files


e5150:
**Provide us a use case of the feature**
Avoid fatal errors due to OS/filesystem character limitations.

**Describe the solution you'd like**
A wrapper could be made around `os.makedirs` and in `write_raw`, such that if an exception is raised (is it always `OSError`?), retry with a sanitised path, where everything except something like `string.ascii_letters + string.digits + '_'` (` + '/'` for dirs) is replaced with `'_'`

**Describe alternatives you've considered**
Keep adding exceptions to `_PostPathFormatter`, but having to take underlying filesystem (as per #575) into account in addition to `platform.system()` seems cumbersome.
(I'm not suggesting the existing replacement be removed, just an addition of a failsafe that covers corner cases.)

**If the feature request is accepted, would you be willing to submit a PR?**

Yes

aandergr:
Good idea! It absolutely makes sense to retry with a sanitized path when filesystem related errors occur. Due to not removing the existing replacement, just adding a failsafe that covers the cases where the existing logic fails, backwards compatibility is assured. I also think this might help regarding our many [Windows-related bugs](https://github.com/instaloader/instaloader/issues?q=is%3Aissue+is%3Aopen+label%3AWindows), at least #185, #560, #575.

> (is it always `OSError`?)

Going through the issues, I saw mentions of `OSError` and `NotADirectoryError` that need to be handled.

e5150:
This seems a bit more complicated than I first though, as there are a lot of places where paths are handled. If a substitution is made in a dirname, not only mkdir needs to know about it, but also the id-file handling, metadata creation. And things such as file existance check and timestamping in `download_pic()` needs to know if a file has been renamed down in `write_raw()`.
Since we can't know if a dir/file creation will be allowed before mkdir/open is called, one could do something like the following, long before the actual file creation is really needed.
```
try:
    open(path, 'w').close()
    os.unlink(path)
except OSError:
    path = sanitize(path)
```

But to avoid sprinkling a lot ugly hacks all over the code, a solution might involve a significant and well considered overhaul…

Issue #575, opened 2020-04-17
[Errno 22] Invalid argument


renox92:
**Describe the bug**
Instaloader is unable to create a highlight folder on a raspberry pi. My guess is it's either # or emoji's fault.

**To Reproduce**
`instaloader banghana --no-posts --highlights -l login -p password`

**Expected behavior**
Instaloader downloading said 

**Error messages and tracebacks**
```
Retrieving highlights "#はなちゃれ 🏠💕" from profile banghana
Traceback (most recent call last):
  File "/home/pi/.local/bin/instaloader", line 10, in <module>
    sys.exit(main())
  File "/home/pi/.local/lib/python3.7/site-packages/instaloader/__main__.py", line 426, in main
    storyitem_filter_str=args.storyitem_filter)
  File "/home/pi/.local/lib/python3.7/site-packages/instaloader/__main__.py", line 195, in _main
    fast_update, post_filter, storyitem_filter)
  File "/home/pi/.local/lib/python3.7/site-packages/instaloader/instaloader.py", line 1084, in download_profiles
    self.download_highlights(profile, fast_update=fast_update, storyitem_filter=storyitem_filter)
  File "/home/pi/.local/lib/python3.7/site-packages/instaloader/instaloader.py", line 49, in call
    return func(instaloader, *args, **kwargs)
  File "/home/pi/.local/lib/python3.7/site-packages/instaloader/instaloader.py", line 679, in download_highlights
    self.download_highlight_cover(user_highlight, highlight_target)
  File "/home/pi/.local/lib/python3.7/site-packages/instaloader/instaloader.py", line 409, in download_highlight_cover
    self.download_title_pic(highlight.cover_url, target, 'cover', highlight.owner_profile)
  File "/home/pi/.local/lib/python3.7/site-packages/instaloader/instaloader.py", line 65, in call
    return func(instaloader, *args, **kwargs)
  File "/home/pi/.local/lib/python3.7/site-packages/instaloader/instaloader.py", line 395, in download_title_pic
    os.makedirs(os.path.dirname(filename), exist_ok=True)
  File "/usr/lib/python3.7/os.py", line 221, in makedirs
    mkdir(name, mode)
OSError: [Errno 22] Invalid argument: 'banghana/#はなちゃれ 🏠💕'
```

**Instaloader version**
4.3.2

**Additional context**
Instaloader 4.3 on macos creates this folder just fine.

e5150:
[...]

Issue #185, opened 2018-10-08
Invalid windows directory starting with "aux."


MatthieuBonenfant:
Hi,

it appears that you cannot create a folder in Windows starting with "aux." The script crashed after saving a few hundred images from a random hashtag until I hit this specific error. I was able to reproduce directly by trying to grab the images from the account.

Microsoft MS-DOS had reserved these names for these system device drivers. Also, for this specific example, all folders starting with these names + "." + anything are also not supported (CON.test or PRN.test or COM1.test, or LPT1.test, etc.) .

https://www.thewindowsclub.com/create-restricted-files-folders-con-aux-nul-click-windows

CON : Keyboard and display
PRN : System list device, usually a parallel port
AUX : Auxiliary device, usually a serial port
CLOCK$ : System real-time clock
NUL: Bit-bucket device
A:-Z: : Drive letters
COM1 : First serial communications port
LPT1 : First parallel printer port
LPT2 : Second parallel printer port
LPT3 : Third parallel printer port
COM2 : Second serial communications port
COM3 : Third serial communications port
COM4 : Fourth serial communications port

I would suggest that instead of crashing, the script should just skip the image(s) and report the directory creation error when it ends instead of killing the queue. 

Alternatively, the script could potentially generate an alternate name, but since this probably just a windows specific issue it would not be cool to non-windows users. Skipping the file would probably the best for everybody at this point.

With the skip rule in place, and if someone would really want to save images for a non-supported account, maybe there is a function already to force a different folder name to be created that is different from the user account name would someone want to target specifically an account with a non-supported folder name. But the chances that it happens again are very slim.

Thank you.

c:\python\scripts\instaloader aux.redkop --fast-update --filename-pattern={profile}_{date_utc}_UTC --no-captions --no-metadata-json
Traceback (most recent call last):
  File "c:\python\scripts\instaloader-script.py", line 11, in <module>
    load_entry_point('instaloader==4.1.1', 'console_scripts', 'instaloader')()
  File "c:\python\lib\site-packages\instaloader\__main__.py", line 388, in main
    storyitem_filter_str=args.storyitem_filter)
  File "c:\python\lib\site-packages\instaloader\__main__.py", line 143, in _main
    profile = instaloader.check_profile_id(target)
  File "c:\python\lib\site-packages\instaloader\instaloader.py", line 776, in check_profile_id
    self.save_profile_id(profile)
[...]

PR #854, opened 2020-11-03
WIP RFC: Sanitize paths also on Linux

UweLau:
I download most of my stuff on Linux machines but organize it on Windows machine. I find it most useful if the paths are already in a valid format so I can transfer the directories directly to my Windows partitions.
I would not mind if there is a command line option (like --sanitizepaths) to make the Linux one optional but this is beyond my capabilities.

Also sanitize the paths on Linux so the data can be transferred to windows without getting invalid data

  - Do you consider it ready to be merged or is it a draft?
It is more a draft.
  - Can we help you at some point?
Yes, please. It would be better to have a command line options rather to make it default for all users.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -131,6 +131,10 @@ class _PostPathFormatter(_ArbitraryItemFormatter):
             ret = ret.replace(':', '\uff1a').replace('<', '\ufe64').replace('>', '\ufe65').replace('\"', '\uff02')
             ret = ret.replace('\\', '\ufe68').replace('|', '\uff5c').replace('?', '\ufe16').replace('*', '\uff0a')
             ret = ret.replace('\n', ' ').replace('\r', ' ')
+        if platform.system() == 'Linux':
+            ret = ret.replace(':', '\uff1a').replace('<', '\ufe64').replace('>', '\ufe65').replace('\"', '\uff02')
+            ret = ret.replace('\\', '\ufe68').replace('|', '\uff5c').replace('?', '\ufe16').replace('*', '\uff0a')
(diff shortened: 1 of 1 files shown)


aandergr:
Hello. First, thanks for sharing your ideas about this. We really messed up the path handling by using e.g. highlight titles as part of filenames. The purpose of the `sanitize_path()` function that your pull request touches is merely to alleviate problems caused by characters that are invalid as part of filenames under Windows (POSIX operating systems such as Linux or Mac OS X are much more permissive). Still we have a lot of pathname-related problems, such as #185, #575, #710, #859 and the one you mentioned here.

Your proposed change effectively enables our less-than-ideal solution of Windows-'sanitizing' the path on Linux.

What I really like though is the idea of adding a flag `--sanitizepaths`. Adding a flag allows to change the behavior without having to worry about issues with backwards incompatibility. Maybe it would be nice to add a flag `--sanitizepaths` to do a proper path sanitation, independent of the OS, that properly removes invalid characters, dots at the ends, and generally everything that could cause problems on any OS or any filesystem. An alternative has been proposed in #576, but I'm inclined to consider adding a flag `--sanitizepaths` to do a proper path sanitation simpler to implement in our current code structure.

canh:
@aandergr I have the same use case and some free time to implement the path sanitisation logic you suggested above. Would this be accepted if I submitted an MR?

aandergr:
@canh nice! Yes, we still consider the above suggested path sanitisation logic a good way to go. We look forward to your pull request!

canh:
Thanks @aandergr, I've submitted my pull request.

PR #872, opened 2020-11-14
Update graphql query hash used for stories

### LINKED ISSUE #871: Unable to fetch all fields in 'tappable_objects'
Comparing the json metadata saved by Instaloader for stories with the one from Instagram Web, I noticed that Instaloader somehow manages to only get a partial `tappable_objects` node.

For example, with `GraphTappableStoryPoll`, Instaloader saves the following fields (values redacted/irrelevant):
```
"tappable_objects":[
   {
      "__typename":"GraphTappableStoryPoll",
      "x":0.0,
      "y":0.0,
      "width":0.0,
      "height":0.0,
      "rotation":0.0,
      "custom_title":null,
      "attribution":null
   }
]
```
whereas Instagram Web returns the full object, with additional fields (again, values redacted/irrelevant):
```
"tappable_objects":[
   {
      "__typename":"GraphTappableStoryPoll",
      "x":0.0,
      "y":0.0,
      "width":0.0,
      "height":0.0,
      "rotation":0.0,
      "custom_title":null,
      "attribution":null,
      "id":"0",
      "option_a":"",
      "option_b":"",
      "option_a_font_size":35.0,
      "option_b_font_size":35.0,
      "question":"",
      "viewer_can_interact":true,
      "viewer_vote":null,
      "tallies":[
         {
            "text":"",
            "count":0,
            "font_size":35.0
         },
         {
            "text":"",
            "count":0,
            "font_size":35.0
         }
      ]
   }
]
```

This appears to be the case for all `tappable_objects` item types. `GraphTappableFeedMedia` for example is also missing most of the fields - including useful ones such as `id`.

~Is there a reason why the full json metadata is not being fetched/saved?
Is Instaloaders query somehow not including those in the request or is this a limitation in the way Instaloader does requests?~ See edit

Returning this sort of partial and limited data is quite useless as only the basic fields like position are present, omitting all the useful fields.

Do note that the Web version is able to return the _full json_ with a _**single query**_, along with all the other data Instaloader saves, so I don't see why Instaloader wouldn't be able to do the same.

~Edit: I noticed in https://github.com/instaloader/instaloader/issues/586#issuecomment-619344484 that tappable objects are "not supported", but what is the actual limitation? Does Instaloader not just verbatim save the json returned from the request? I assume it does, so does the returned json not contain those fields? (and if so - why not?)~

**Edit: Edit:** Instagram Web uses an **updated GraphQL query hash**: `303a4ae99711322310f25250d988f3b7`, issuing the request with this one indeed appears to return all fields.

PR https://github.com/instaloader/instaloader/pull/872 fixes this by updating to the new query hash.

Tested Instaloader version: `4.5.3`

Ristovski:
This updated graphql query hash returns all fields for `tappable_objects`. These fields were missing/partial when using the old query hash.
This allows to get the full json metadata of a story. Fixes #871

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -584,7 +584,7 @@ class Instaloader:
                 yield userids[i:i + userids_per_query]
 
         for userid_chunk in _userid_chunks():
-            stories = self.context.graphql_query("bf41e22b1c4ba4c9f31b844ebb7d9056",
+            stories = self.context.graphql_query("303a4ae99711322310f25250d988f3b7",
                                                  {"reel_ids": userid_chunk, "precomposed_overlay": False})["data"]
             yield from (Story(self.context, media) for media in stories['reels_media'])
 


aandergr (approved):
That's a nice query hash, thanks!

Story-related tests still pass, and all attributes of the yielded Story and StoryItems objects are still accessible. It seems the data returned using this query hash contains all the (relevant) fields, and it includes more-complete tappable_objects information.

Deleora:
So now is it possible to download the actuals posts shared in stories?

aandergr:
> So now is it possible to download the actuals posts shared in stories?

Yes, there is a field `tappable_objects` in the JSON that is now returned for each StoryItem. So with the JSON saved, something like the following will give you an `id` and a `shortcode` of a post shared in a story:

```sh
$ xzcat 2021-01-30_05-57-43_UTC.json.xz | jq .node.tappable_objects[0].media
```

or download it:

```sh
$ instaloader -- -`xzcat 2021-01-30_05-57-43_UTC.json.xz | jq -r .node.tappable_objects[0].media.shortcode`
```

However, the tappable objects are not yet exposed in the programmatic interface of StoryItem.

Deleora:
Ohh, that explains why I wasn't seeing any such posts getting downloaded. Anyway, Thanks for one more cool feature!!

So one question, can I download such posts using the uncompressed JSON files? If so how? because I have only uncompressed JSON files.
And I didn't understand the purpose of having those backticks(`) in the command. Is it mandatory? [I have Windows 10]

aandergr:
The backticks do what is called [_Command substitution_](https://docs.microsoft.com/de-de/powershell/scripting/learn/deep-dives/everything-about-string-substitutions?view=powershell-7.1#command-substitution), allowing the output of a command to replace the command name.

I think in Powershell and with the JSON files not compressed, the following command should do the job

```posh
instaloader -- -$(jq -r '.node.tappable_objects[0].media.shortcode' '2021-01-30_05-57-43_UTC.json')
```

(it requires [jq](https://stedolan.github.io/jq/download/) to be installed)

Issue #217, opened 2018-12-25
Is Instaloader broken for good? Still doesn't login after SMS Challenge


Zachary24:
**Describe the bug**
Requests Challange every single time, and doesn't allow me to use it, used multiple accounts (4), which all reproduce the same result, as well as tried using a vpn, also tried 3 different devices (macbook, windows, and a raspberry pi), every time the problem still occurs, It started happening every so often in 4.1.1, and always happens with 4.2 and the earlier version. I've also seen PLENTY of others with the exact same problem.

**Error messages and tracebacks**

Session file does not exist yet - Logging in.
Fatal error: Login: Checkpoint required. Point your browser to https://www.instagram.com/challenge/{ID}/{ID}/, follow the instructions, then retry.
**Instaloader version**
`4.2`  (but also sometimes happened on 4.1.1 and always happened on the version before 4.2)

e5150:
When I was stuck with this issue a while back, I ended up copying the session data from my browser by hand.
```
$ python3
>>> import pickle
>>> data = pickle.load(open(<old-sessionfile>, 'rb'))
>>> data['csrftoken'] = <browser-csrftoken>
>>> data['sessionid'] = <browser-sessionid>
>>> pickle.dump(data, open(<new-sessionfile>, 'wb'))
```
I think those are the only cookies that need to be copied. This could be automated, if one were to figure out how to export cookies from the various browsers. Firefox is easy enough, `select name, value from moz_cookies where baseDomain = "instagram.com"`. Chromium and its derivatives is more of a challenge `select name,encrypted_value from cookies where host_key like "%instagram.com"` (AES-128-CBC, on linux without kde/gnome-keyring-stuff password: "peanuts" salt: "saltysalt")...

aandergr:
I'm closing this in favor of #92, our thread for login problems.

nOOBIE-nOOBIE:
> When I was stuck with this issue a while back, I ended up copying the session data from my browser by hand.
> 
> ```
> $ python3
> >>> import pickle
> >>> data = pickle.load(open(<old-sessionfile>, 'rb'))
> >>> data['csrftoken'] = <browser-csrftoken>
> >>> data['sessionid'] = <browser-sessionid>
> >>> pickle.dump(data, open(<new-sessionfile>, 'wb'))
> ```
[...]

Issue #92, opened 2018-04-08
Login error: 400 Bad Request


jonathanlaniado:
Logging in used to work, but I get this error every time now.

If I go to www.instagram.com, I get a suspicious activity/unusual login attempt page. Even after confirming that, yes, "This was me", I receive the same login connection error and another suspicious activity page.

aandergr:
Unfortunately, I cannot reproduce this error, making it impossible to me for fix it. Could you provide more information about when it happens?

aandergr:
Does this problem still persist? If so, does our [4.0 pre-release](https://github.com/instaloader/instaloader/releases/tag/v4.0rc1) alleviate the problem?
```
pip3 install --upgrade --pre instaloader~=4.0rc1
```

secure12:
Yes, the problem still persists.
If one calls the function instaloader.Instaloader().login/interactive_login with valid username and password, `instaloader.exceptions.ConnectionException: Login error: 400 Bad Request` will prompt.

secure12:
One interesting observation is that the "suspicious" login time as shown on Instagram is not the same as actual login attempt time, e.g. the actual time: 28-May-2018 3:31 AM (UTC+8), login attempt time shown on Instagram: 27-May-2018 12:31 PM (US West Coast?).

nagualcode:
I face this same problem when loggin with instaloader from a vps located in UK.
I have tried loggin with web browser from that same vps, it works, I face the "unusual loggin" challenge, but instaloader is never able to login from that same machine.

cristianCarrerasCastillo:
ConnectionException('Login error: 400 Bad Request',)
I use Instaloader 4.1.1 and I also have that error
(country chile, south america)

MattrAus:
I got this error while I had 2FA enabled. I disabled it and was able to login.

Hope this helps.

PhantomTech:
In cases where 2FA is disabled, this issue seems to be caused when Instagram detects the login as suspicious. When this happens Instagram asks the user to verify the login with a code sent to the user's registered phone number or email address. https://github.com/rarcega/instagram-scraper seems to handle this through its interactive login (-i) feature, maybe something similar could be implemented here?

aandergr:
With version 4.2rc2 we changed how Instaloader's login function copes with failure responses (commit 4ee867c61b41db14cef7bfb7897519046540a242). Could the ones affected here please try that version and comment which error message is shown?

Upgrade with
```
[...]

PR #877, opened 2020-11-17
Support importing cookies.txt to workaround infinite challenge requests

elexis1:
Same issue as described in the following issues:

#92
#217
#615 

When using instaloader, it asks for a challenge performed in the browser:
> Fatal error: Login: Checkpoint required. Point your browser to https://www.instagram.com/challenge/xxx/yyy/, follow the instructions, then retry.

Been there, done that, only result is that you just get another challenge, forever.

There are code snippet shareds in the issues, and committed in 27b7c98be2beaf8af899d53e7ab30c373872a598 (see  #615).

It loads the Firefox "cookies.sqlite" and writes a new instaloader session file.

However, this snippet doesn't work anymore, because Firefox uses an **exclusive lock**.
Hence sqlite3 cannot read from `cookies.sqlite`, and the script can't function.
For example running `sqlite3 ~/.mozilla/firefox/*/cookies.sqlite` and `SELECT name, value FROM moz_cookies;`.
See also:
https://superuser.com/questions/635317/sqlite-database-browser-fails-to-open-firefox-cookie-database
https://www.sqlite.org/pragma.html#pragma_locking_mode

I'm using Firefox 82.0.3, the lock was introduced probably much sooner.

To work around this issue, I have adapted the script to work with `cookies.txt` files.

To obtain the cookies.txt from Firefox, one can install a plugin, such as this https://addons.mozilla.org/de/firefox/addon/export-cookies-txt/

This is a severe issue for the people affected by the infinite challenge loop, because it makes it impossible to use the program for them.

### DIFF
--- a/docs/codesnippets/615_import_firefox_session_sqlite.py
+++ b/docs/codesnippets/615_import_firefox_session_sqlite.py
@@ -1,8 +1,14 @@
+#!/usr/local/bin/python3
+
 from argparse import ArgumentParser
 from glob import glob
 from os.path import expanduser
 from platform import system
 from sqlite3 import OperationalError, connect
+
+# This script attempts to load Firefox cookies.sqlite file,
+# login to instagram and write the instaloader session file.
+# Try this method if you have trouble logging in using instaloader.
 
 try:
     from instaloader import ConnectionException, Instaloader
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -63,15 +63,15 @@ creates when using :option:`--login`. If a session file is present,
 Also, session files usually do not expire.
 
 If you do not have a session file present, you may use the following script
-(:example:`615_import_firefox_session.py`) to workaround login problems by
+(:example:`615_import_firefox_session_sqlite.py`) or (:example:`615_import_firefox_session_txt.py`) to workaround login problems by
 importing the session cookies from Firefox and bypassing Instaloader's login and
 so still use Instaloader's logged-in functionality.
 
-.. literalinclude:: codesnippets/615_import_firefox_session.py
+.. literalinclude:: codesnippets/615_import_firefox_session_sqlite.py
 
 To use this script,
 
-#. Download the script: :example:`615_import_firefox_session.py`,
+#. Download the script: :example:`615_import_firefox_session_sqlite.py`,
 
 #. Login to Instagram in Firefox,
 
@@ -83,3 +83,6 @@ This script also supports specifying a cookie file path, which may be useful if
 you use multiple Firefox profiles or if your operating system has the directory
 structure differently set up. Also, you can specify an alternative session file
 path.
+
+If this script doesn't work for you (for example because Firefox locks the sqlite file),
+you can alternatively try to export the cookies to a `cookies.txt` file and consume that using :example:`615_import_firefox_session_sqlite.py`.
--- a/docs/codesnippets/615_import_firefox_session_txt.py
+++ b/docs/codesnippets/615_import_firefox_session_txt.py
@@ -1,0 +1,36 @@
+#!/usr/local/bin/python3
+
+import http.cookiejar
+from argparse import ArgumentParser
+
+# This script attempts to load a cookies.txt file,
+# login to instagram and write the instaloader session file.
+# Try this method if you have trouble logging in using instaloader.
+
+try:
+    from instaloader import ConnectionException, Instaloader
+except ModuleNotFoundError:
+    raise SystemExit("Instaloader not found.\n  pip install [--user] instaloader")
+
+def import_session(cookiefile, sessionfile):
+    print("Using cookies from {}.".format(cookiefile))
+    cj = http.cookiejar.MozillaCookieJar()
+    cj.load(cookiefile, ignore_discard=True, ignore_expires=True)
+    instaloader = Instaloader(max_connection_attempts=1)
+    instaloader.context._session.cookies.update(cj)
+    username = instaloader.test_login()
+    if not username:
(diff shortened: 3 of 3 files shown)


aandergr:
> However, this snippet doesn't work anymore, because Firefox uses an **exclusive lock**.
> Hence sqlite3 cannot read from `cookies.sqlite`, and the script can't function.
>
> [...]
> 
> This is a severe issue for the people affected by the infinite challenge loop, because it makes it impossible to use the program for them.

That is not true. Firefox can only hold the lock while it is running. Calling the snippet while Firefox  is running results in the message _Cookie import failed: database is locked_. The snippet works fine when no Firefox instance is running.

elexis1:
Yes, the lock should be gone if firefox is closed.

At the very least the instructions are incomplete:
> #. Login to Instagram in Firefox,
> #. Execute the snippet, e.g. with ``python 615_import_firefox_session.py``,

I don't know how many people store their cookies across browser sessions or what the defaults are.
People who delete cookies upon closing the browser would have to get the hassle to:
- disable deletion of cookies
-  shutting down firefox, which kills all active sessions, possibly open forms data, selected text, the location of each page one scrolled in
- has to reload all tabs again (for some the content might have changed)
- change the settings again to delete cookies upon quit.

That hassle is multiplied everytime one gets into this verification loop.

So to me those 3 lines of code load the cookies file from a cookies.txt instead of the cookie.sqlite file not only make it much easier but make it feasible at all for me. I'd have to keep this script around merely to keep firefox settings and be able to use instaloader.

Judging by that, there are two things we observe:
- The lockout loop shouldn't be a thing that happens to begin with. The challenge should be solved once and then it should be fine instead of loop forever. So the real question is why is there a challenge loop and how can it be avoided without a manual cookie extraction with a script.

- If the challenge loop can't be fixed for good, perhaps it would be cleaner to not use one or two code snippets for the needy but to integrate it into instaloader as a commandline argument `--cookies cookiefile`. Perhaps that could support both txt and sqlite format (either by two arguments or content type detection).

aandergr:
Ah that's right. I'm sorry, I hadn't thought about it in enough detail on Saturday when I wrote my comment, what consequences it can actually have for a user that the script cannot access the cookies while Firefox is running. Thanks for your patience and your explanatory comment!

How about if we just get the script to open the original cookies database even if it is locked? We could do that by passing the [immutable flag](https://www.sqlite.org/uri.html#uriimmutable) to SQLite, since we only access the database read-only:

```diff
--- a/docs/codesnippets/615_import_firefox_session.py
+++ b/docs/codesnippets/615_import_firefox_session.py
@@ -23,7 +23,7 @@ def get_cookiefile():
 
 def import_session(cookiefile, sessionfile):
     print("Using cookies from {}.".format(cookiefile))
-    conn = connect(cookiefile)
+    conn = connect(f"file:{cookiefile}?immutable=1", uri=True)
     try:
         cookie_data = conn.execute(
             "SELECT name, value FROM moz_cookies WHERE baseDomain='instagram.com'"
```

This fixes importing the cookies from Firefox for me, and I think it is easier than taking the workaround via the cookiejar.

elexis1:
I can confirm that this avoids running into the lock issue.

However the cookies file does not contain the cookies, because firefox presumably cached them and didn't write it to the cookies.sqlite until having closed firefox, and when closing firefox, the cookies are deleted depending on setting.

Hence the instaloader script tells me I'm not logged in in firefox (but I am).
Equivalently I get an empty table when doing this:
> sqlite3 file:///home/user/.mozilla/firefox/layinfzi.default/cookies.sqlite?immutable=1
> SELECT * FROM moz_cookies;

To ensure that the file isn't broken somehow, I've closed firefox, deleted (renamed) cookies.sqlite, started firefox, logged in, cookies.sqlite was recreated, but still claims an empty table.

I assume that this is because the cookies are cached and the cache is only written to the file on the disk when the buffer is flushed, not everytime a cookie is created. And in the case of sticking with the automatically deleting cookies on exit setting, I suspect there might be no way to flush cookies.sqlite explicitly.
The file modification date does not change when I create more cookies (login elsewhere), even 20 minutes later, until I close firefox.

docs/codesnippets/615_import_firefox_session_sqlite.py
```diff
@@ -1,9 +1,15 @@
+#!/usr/local/bin/python3
```
aandergr, line 1: Could we remove the shebang here? The snippet is included verbatim in the documentation, and this adds two lines without adding any real value, I think.

docs/troubleshooting.rst
```diff
@@ -83,3 +83,6 @@ This script also supports specifying a cookie file path, which may be useful if
 you use multiple Firefox profiles or if your operating system has the directory
 structure differently set up. Also, you can specify an alternative session file
 path.
+
+If this script doesn't work for you (for example because Firefox locks the sqlite file),
```
aandergr, line 87: Two ideas here:

- With connecting to the file with immutable=1 (which still has to be changed), Firefox locking the file is not a reason for 615_import_firefox_session.py to not work, so we should choose another example, or remove the "for example" clause.
- Maybe (not so sure about that, just want to suggest it), "If this script doesn't work for you" should be written in bold (surrounded by two asterisks `**`) to be more visible when being frustrated and not reading all the text.

aandergr (changes requested):
I see, thanks for experimenting with that! So the issue addressed by adding the cookiejar import script is not that Firefox locks its cookie database, since we can still open it for reading (after a small change that is not committed yet), but that it seems that Firefox under certain (unknown?) conditions or settings does not persist the cookies.

I added a few inline comments in this review, with some (very minor) issues I noticed when going over the changes.

Issue #892, opened 2020-11-28
Save content using original filenames


thomas694:
**Provide us a use case of the feature**
The new feature would download posts and save the images/videos with the original filenames.
That would make it possible to change the tool resp. update existing profile downloads that have been saved by filenames.

**Describe the solution you'd like**
An extension of the filename-pattern parameter: --filename-pattern={filename}

**Describe alternatives you've considered**
Considered, but not found.

**If the feature request is accepted, would you be willing to submit a pull request?**
Yes

**Additional context**
There exists a closed issue #395 and some other user commented something similar in another issue.

PR #893, opened 2020-11-29
Save content using original filenames

thomas694:
Please describe:

  - Implement feature issue #892 
The new feature would download posts and save the images/videos with the original filenames.

The implementation follows the other options for the filename-pattern parameter. As filename is related to a single file and not the whole post, for sidecar nodes the filename is determined again.
- Remark
Using this option alone media from one post is only grouped together by the file modification timestamp.

  - Is it just a proof of concept? *No*
  - Is the documentation updated (if appropriate)? *Yes*
  - Do you consider it ready to be merged or is it a draft? *ready*
  - Can we help you at some point? *please review*

### DIFF
--- a/docs/basic-usage.rst
+++ b/docs/basic-usage.rst
@@ -171,6 +171,9 @@ dirname pattern. The following tokens are defined for usage with
 - ``{mediaid}``
    Integer representation of shortcode.
 
+- ``{filename}``
+   Instagram's internal filename.
+
 - ``{date_utc}`` (same as ``{date}``)
    Creation time in UTC timezone.
    `strftime()-style formatting options <https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior>`__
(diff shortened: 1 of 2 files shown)


thomas694:
I'm not familiar with Python, especially PyLint and MyPy. I installed them locally, but they behaved differently to your automatic checks. A few hints how to configure them etc. in the contributing guidelines would be helpful.
But now they're green.

aandergr:
That's a beautiful pull request, thanks! One thing though: It does not work with stories, download_storyitem() also has a call to format_filename().

> I'm not familiar with Python, especially PyLint and MyPy. I installed them locally, but they behaved differently to your automatic checks. A few hints how to configure them etc. in the contributing guidelines would be helpful.

Thanks for that note, I see that this section needs a little more work. Strange that Pylint and MyPy behaved differently. Maybe a different version? Did you install them directly via pip or did you use pipenv to setup our dependencies?

thomas694:
I added code for story items too.

In the contributing guide I didn't see any steps and actually I didn't remember the installation guide as I already "installed" it some time ago.
Here are my versions and obviously it's neither your recommended nor the current version which it used. Maybe because Python was already installed and I only added PyLint and MyPy (not using pip).
pylint 1.9.4
mypy 0.670
Python 2.7.16

aandergr (approved):
Thanks!

Issue #121, opened 2018-05-16
Collect Data in a specific period (Since - Until)


pvita:
Hi there,
is it possible to collect pictures (and metadata) only from a specific period? 
For example all the pictures of the last year with a specific hashtag.

P.s. is it possible to collect just metadata (+comments; geotags etc) without downloading all the pictures?

Best
P

Thammus:
Hi,
you can play around with `dropwhile` and `takewhile` from [itertools](https://docs.python.org/3/library/itertools.html) like in the following snippet:
```python
from datetime import datetime
from itertools import dropwhile, takewhile

import instaloader

L = instaloader.Instaloader()

posts = L.get_hashtag_posts('milfgarden')
# or
# posts = instaloader.Profile.from_username(L.context, PROFILE).get_posts()

SINCE = datetime(2015, 5, 1)
UNTIL = datetime(2015, 3, 1)

for post in takewhile(lambda p: p.date > UNTIL, dropwhile(lambda p: p.date > SINCE, posts)):
    print(post.date)
    L.download_post(post, '#milfgarden')
```

stagenam3:
Hi! I tried out that feature (copy pasted the code, changed the hashtag and dates, and ran the script)  and I get nothing, i.e. I do not get any error message and the script seems to have ran but I also do not get any results.

suecys:
Hi, I have the same error as stagenam3, I get nothing when I change the dates and the hashtag. If I change the hashtag only, still nothing. What can I do to get results?

oktayekici:
Same problem here. It does nothing after you run the code like the code has finished conducting.

rohimari:
I get a problem, 429 Too Many Requests, how I can resolve it?

rohimari:
> Hi,
> you can play around with `dropwhile` and `takewhile` from [itertools](https://docs.python.org/3/library/itertools.html) like in the following snippet:
> 
> ```python
> from datetime import datetime
> from itertools import dropwhile, takewhile
> 
> import instaloader
> 
> L = instaloader.Instaloader()
> 
> posts = L.get_hashtag_posts('milfgarden')
> # or
> # posts = instaloader.Profile.from_username(L.context, PROFILE).get_posts()
> 
> SINCE = datetime(2015, 5, 1)
> UNTIL = datetime(2015, 3, 1)
> 
> for post in takewhile(lambda p: p.date > UNTIL, dropwhile(lambda p: p.date > SINCE, posts)):
[...]

PR #898, opened 2020-11-30
Added ability to retrieve posts in NodeIterator[Post] since and between given times.

synchronizing:
As discussed in #121, Instaloader does not have the capability out-of-the-box to retrieve posts from a `NodeIterator[Post]` type between dates and as far back from the current time. This pull-request adds those two features by adding two functions, `NodeIterator.between()` and `NodeIterator.since()` which only works with `NodeIterator[Post]` that are chronologically organized (specifically `get_posts`, `get_tagged_posts`, and `get_igtv_posts`). 

### `.between()`

```python
from datetime import datetime, timedelta
import instaloader

L = instaloader.Instaloader()
profile = instaloader.Profile.from_username(L.context, "instagram")
posts = profile.get_posts()

for post in posts.between(start=datetime(2020, 11, 27), end=datetime(2020, 11, 30)):
    print(post.date)
```

### `.since()`

```python
from datetime import datetime, timedelta
import instaloader

L = instaloader.Instaloader()
profile = instaloader.Profile.from_username(L.context, "instagram")
posts = profile.get_posts()

for post in posts.since(ago=timedelta(days=3)):
    print(post.date)
```

## Implementation

To make this possible, a new param `chronological` was added to `NodeIterator`:

```python
class NodeIterator(Iterator[T]):
    def __init__(self,
                 context: InstaloaderContext,
                 query_hash: str,
                 edge_extractor: Callable[[Dict[str, Any]], Dict[str, Any]],
                 node_wrapper: Callable[[Dict], T],
                 query_variables: Optional[Dict[str, Any]] = None,
                 query_referer: Optional[str] = None,
                 first_data: Optional[Dict[str, Any]] = None,
                 chronological: Optional[bool] = None):
        ...
```

This variable is used to make the distinction between those `NodeIterators` that are chronological, and those that are not. All functions in the codebase that returns a `NodeIterator[Post]` were modified.

## Note

Since `Hashtag.get_posts()` returns an `Iterator[Post]` type, this modification will not change this section of the codebase. Instead, it only applies to the `Post` class.

## Help

If anything, I just ask documentation is looked through for any spelling mistakes or improvement. Engrish hard. Besides that, I had a hard time trying to figure out two good words for the functions -- left it as `between` and `since(ago=)`, but any recommendation is welcomed.

## Todo

- [x] Functionality.
- [x] Proper type-hinting.
- [x] Added documentation.
- [x] Fixed PyLint and MyPy complaints.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 __pycache__/
 .mypy_cache/
+instaloader.egg-info
--- a/docs/codesnippets.rst
+++ b/docs/codesnippets.rst
@@ -28,21 +28,19 @@ Download Posts in a Specific Period
 -----------------------------------
 
 To only download Instagram pictures (and metadata) that are within a specific
-period, you can simply use :func:`~itertools.dropwhile` and
-:func:`~itertools.takewhile` from :mod:`itertools` on a generator that returns
-Posts in **exact chronological order**, such as :meth:`Profile.get_posts`.
+period, you can simply use :func:`NodeIterator.between` and :func:`NodeIterator.since`.
+Only works with posts in **exact chronological order**, such as :func:`Profile.get_posts`.
 
 .. literalinclude:: codesnippets/121_since_until.py
 
+Or, similarly, to retrieve posts a `timedelta` amount of time since now.
+
+.. literalinclude:: codesnippets/121_since_until_to.py
+
 See also :class:`Post`, :meth:`Instaloader.download_post`.
 
-Discussed in :issue:`121`.
-
-The code example with :func:`~itertools.dropwhile` and
-:func:`~itertools.takewhile` makes the assumption that the post iterator returns
-posts in exact chronological order.  As discussed in :issue:`666`, the following
-approach fits for an **almost chronological order**, where up to *k* older posts
-are inserted into an otherwise chronological order, such as a Hashtag feed.
+As discussed in :issue:`666`, the following approach fits for an **almost chronological order**, 
+where up to *k* older posts are inserted into an otherwise chronological order, such as a Hashtag feed.
 
 .. literalinclude:: codesnippets/666_historical_hashtag_data.py
 
--- a/docs/codesnippets/121_since_until_to.py
+++ b/docs/codesnippets/121_since_until_to.py
@@ -1,9 +1,9 @@
-from datetime import datetime
+from datetime import timedelta
 import instaloader
 
 L = instaloader.Instaloader()
 profile = instaloader.Profile.from_username(L.context, "instagram")
 posts = profile.get_posts()
 
-for post in posts.between(start=datetime(2020, 11, 27), end=datetime(2020, 11, 30)):
+for post in posts.since(ago=timedelta(days=3)):
     L.download_post(post, "instagram")
--- a/docs/module/nodeiterator.rst
+++ b/docs/module/nodeiterator.rst
@@ -22,6 +22,9 @@ Iterator :class:`NodeIterator` and a context manager
 .. autoclass:: NodeIterator
    :no-show-inheritance:
 
+   .. automethod:: between
+   .. automethod:: since
+
 .. autoclass:: FrozenNodeIterator
    :no-show-inheritance:
 
--- a/instaloader/exceptions.py
+++ b/instaloader/exceptions.py
@@ -58,6 +58,10 @@ class PostChangedException(InstaloaderException):
     pass
 
 
+class IncorrectType(InstaloaderException):
+    pass
+
+
 class QueryReturnedNotFoundException(ConnectionException):
     pass
 
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -928,6 +928,7 @@ class Instaloader:
             lambda d: d['data']['user']['edge_web_discover_media'],
             lambda n: Post(self.context, n),
             query_referer='https://www.instagram.com/explore/',
+            chronological=False
         )
 
     def get_hashtag_posts(self, hashtag: str) -> Iterator[Post]:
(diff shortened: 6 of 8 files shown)


aandergr:
Thanks for your PR! However, I feel I do not understand the motivation of introducing these methods. What is the advantage of `.between()` and `.since()` over just using `takewhile()` and `dropwhile()` from Python's itertools?

synchronizing:
`takewhile()` and `dropwhile()` either no longer works, or never worked. Within #121 there has been discussion of such code not working, but just in case, here is a modified version of @Thammus code to use `instagram` account and recent dates:

```python
from datetime import datetime
from itertools import dropwhile, takewhile

import instaloader

L = instaloader.Instaloader()

posts = instaloader.Profile.from_username(L.context, 'instagram').get_posts()

SINCE = datetime(2020, 11, 27)
UNTIL = datetime(2020, 12, 2)

for post in takewhile(lambda p: p.date > UNTIL, dropwhile(lambda p: p.date > SINCE, posts)):
    print(post.date)
```

Running the following:

```python
from datetime import datetime, timedelta
import instaloader

L = instaloader.Instaloader()
profile = instaloader.Profile.from_username(L.context, "instagram")
posts = profile.get_posts()

for post in posts.between(start=datetime(2020, 11, 27), end=datetime(2020, 12, 2)):
    print(post.date)
```

we receive:

```
2020-12-01 17:03:02
2020-11-30 17:23:39
2020-11-29 17:12:21
2020-11-28 16:51:40
2020-11-27 18:34:20
```

Frankly, I would not mind changing the API. If you believe adding functions to `NodeIterator[Post]` is to grotesque, I wouldn't mind resolving the issue with some sort of `utils` addition that would add `utils.between(posts, start, end)` functionality.

Thammus:
In order to correctly use the `takewhile()`/`dropwhile()` example (which is also listed in our [documentation](https://instaloader.github.io/codesnippets.html#download-posts-in-a-specific-period)) you have to make sure that `UNTIL` is less (i.e. represents an earlier date) than `SINCE`. The confusion arises because in this context the terms _since_ and _until_ should not be interpreted in the usual chronological way. They are just meant to mark two points in a list of posts that get yielded by an iterator. Because the posts are received in descending chronological order (i.e. from _newest_ to _oldest_), the start marker (`SINCE`) needs to be a more recent date than the end marker (`UNTIL`). 

Ok, so lets go through the code step by step.
```python
START_MARKER = datetime(2020, 12, 2)
posts_after_start_marker = dropwhile(lambda p: p.date > START_MARKER, posts)
```
` dropwhile(lambda p: p.date > START_MARKER, posts)` _drops_ all posts _while_ their `date` is greater than the start marker. This skips all posts that are newer.
```python
END_MARKER = datetime(2020, 11, 27)
posts_before_end_marker = takewhile(lambda p: p.date > END_MARKER, posts)
```
`takewhile(lambda p: p.date > END_MARKER, posts)` provides posts as long as their `date` is greater than the end marker. This skips all posts that are older.

Combining those two gives us posts that are _older_ than the start marker and _newer_ than the end marker:
```python
posts_between = takewhile(lambda p: p.date > END_MARKER, posts_after_start_marker)
# or
posts_between = takewhile(lambda p: p.date > END_MARKER, dropwhile(lambda p: p.date > START_MARKER, posts))
```

synchronizing:
Very confusing indeed. Just tested it once again with the tips above, and it is working as expected. Any use for the PR, or a `instaloader.utils` file? Will close otherwise.

aandergr:
I don't think there is a need to duplicate parts of Python's itertools in an `instaloader.utils` file.

What I really like about this PR though is that it documents where a user may expect the posts to be returned in chronological order. Adding this information to the docstrings of the respective methods would be a great enhancement to our documentation.

Also we are always open for ideas on how to make the documentation easier to understand and less confusing.

synchronizing:
I think @UweLau raises a good point that I did not think over when implementing this PR -- the _real_ usefulness comes from folks using the tool via it's CLI. The idea behind `instaloader.utils` would be to implement the combination of `takewhile` and `dropwhile` as mentioned by @Thammus so that it's more easily accessible when either using the tool via the CLI, or via scripting. That is, the `.utils` would be useful in the implementation of (say):

```
instaloader profile --start 2020-10-15 --end 2020-12-31 
```

and

```python
from datetime import datetime, timedelta
import instaloader

L = instaloader.Instaloader()
profile = instaloader.Profile.from_username(L.context, "instagram")
posts = profile.get_posts()

for post in instaloader.between(posts, start=datetime(2020, 11, 27), end=datetime(2020, 11, 30)):
    print(post.date)
```

Just some thoughts.

aandergr:
We are not against CLI flags that allow downloading only specific date ranges. 

However, the here-proposed change does not implement such flags for the CLI, rather it adds functions that do what can already easily be done by functions provided by Python's standard library. The here-proposed functions are redundant, whether they are used in the CLI or not.

Also, they are not abstracting complexity. Instead, they specialize the abstract takewhile() and dropwhile() functions, which work with any iterator and any condition, to the specific use case of only filtering NodeIterator Posts by their creation date.

I agree with since() and between() being easier to understand; but I think that this is not a point given that takewhile()/dropwhile() are a) not complicated either, b) well explained in the Python documentation and c) explained in a copy&paste-able example that even links to the Python documentation.

FayzulSaimun:
I am having trouble using since() and between function here. I want to get the post from the last 30 minutes. The error I get is
AttributeError: 'NodeIterator' object has no attribute 'between' and AttributeError: 'NodeIterator' object has no attribute 'since'

` for post in posts.between(start=datetime.now(), end=start - datetime.timedelta(minutes=30)):
        print(post.date)`

Issue #724, opened 2020-07-10
Only downloads the first 150 profiles it sees with :story.


Liz-chan:
**Describe the bug**

When you use :story to download all the stories from all the profiles you follow, it only gets the first 150.

**To Reproduce**

Try to use the :story target and see that it misses accounts that are past the initial 150.

**Expected behavior**

I expected all the profile's stories to be downloaded, not just the first 150 it can see.

**Error messages and tracebacks**

No error message, it just says it's complete.

**Instaloader version**

4.4.4

**Additional context**

I'm following around 1.5k accounts, so maybe that's why it's having trouble downloading all the stories for all those profiles.

e5150:
Might be the same thing as with #204. How many stories are available through the website? If only the first 150 are visible on the site you'd have to resort to `instaloader --stories @myusername`. But if more user stories are loaded as you scroll, then pagination has been implemented on the server side, and a fix in Instaloader shouldn't be too hard.

Liz-chan:
Oh I see, yeah, only the first 150 are available on the website. I did try to use the combo that downloads only stories and only of the people you follow, but it kept hitting the rate limit and it seems like it wasn't doing anything, since after 2 hours it was still stuck in the same place.

Here's the command that I used for reference, there doesn't seem to be anything wrong with it right?
`instaloader --login=xxxxxxxxxxxxxx --dirname-pattern={profile} --filename-pattern={profile}-{date_utc}-{target} --no-video-thumbnails --no-metadata-json --stories --no-posts --no-profile-pic @xxxxxxxxxxxxxx`

I was thinking that a way this could be solved is that while downloading the initial 150, there's an option to mark them as read and restart, so that you're able to access the fresh new unread ones. :stories is so so much faster than --stories --no-posts so I think that would be the best for efficiency.

e5150:
Your command looks fine, but with many followees I get that is might take too long.
I quicky made a potential fix, by marking stories as seen after downloading them, which I've been thinking about for a while anyway. It's available at https://github.com/e5150/instaloader/tree/seen_stories
It unconditionally sends a POST request to the IG servers after each `StoryItem` has been downloaded, similarly to the official website, but I don't think that would count towards the ratelimit. I don't follow sufficiently many accounts to be able to really stress test it though.

Liz-chan:
Alright, thank you! I'll give it a try and report back.

Liz-chan:
[...]

PR #905, opened 2020-12-05
Marks story as seen when downloaded

Liz-chan:
This fixes issue #724, which wasn't able to download every story available to your account because it wasn't marking them as read, and therefore couldn't get the other stories it wasn't showing you. This fixes the issue by marking a story as read as soon as you download it, so the other stories can then be downloaded since Instagram always shows you the unread stories first.

I would consider this to be sufficient in 99% of cases, but an improvement would be if it only marked it as read when accessing the 1st storyitem _and_ it successfully downloads, rather than marking it as read every time the 1st story is accessed, which would help with the mark as read rate limits. I would do it myself, but I'm not good at python, although it doesn't seem like it would be particularly difficult, so any help would be appreciated.

EDIT: [This](https://github.com/e5150/instaloader/commit/2ab43b0705d88baaf0656a3aa3ef09f12acb12f3) commit does it, but I don't know how to do it along with only doing the 1st storyitem of each story.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -618,6 +618,7 @@ class Instaloader:
             totalcount = user_story.itemcount
             count = 1
             for item in user_story.get_items():
+                if count == 1: item.mark_as_seen()
                 if storyitem_filter is not None and not storyitem_filter(item):
                     self.context.log("<{} skipped>".format(item), flush=True)
                     continue
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -618,7 +618,8 @@ class Instaloader:
             totalcount = user_story.itemcount
             count = 1
             for item in user_story.get_items():
-                if count == 1: item.mark_as_seen()
+                if count == 1:
+                    item.mark_as_seen()
                 if storyitem_filter is not None and not storyitem_filter(item):
                     self.context.log("<{} skipped>".format(item), flush=True)
                     continue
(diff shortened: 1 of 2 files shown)


kandayo:
I think it should not be the default behavior. It's kind of a breaking change.

But it's a great feature indeed.

Liz-chan:
> I think it should not be the default behavior. It's kind of a breaking change.
> 
> But it's a great feature indeed.

Hm I'm not sure, usually you want the thing that works to be the default, or else people are going to be confused why it's not downloading all the stories on their account, but I guess it depends on what the main developer wants.

Thammus:
Marking stories as seen is a feature that strictly speaking violates Instaloader's main goal to access the data without changing any of it online. We could possibly imagine an exception to this rule if it is otherwise not possible to access some important data. The combo that downloads only stories of the people you follow like described in https://github.com/instaloader/instaloader/issues/724#issuecomment-656909683 did have one major flaw compared to using `:stories` because it unnecessarily fetched the metadata of all followees which was very time consuming. This inefficient behavior is fixed with ed55ce140df608b972419b3fb351560aba299f3a. So, using the current master branch with the command
```
instaloader -l xxxxxxxxxxxxxx @xxxxxxxxxxxxxx --stories --no-profile-pic --no-posts --no-metadata-json
```
downloads all available stories efficiently. Since this is a well working alternative to using `:stories` we do not need the hack of marking stories as seen.

Liz-chan:
Mmm I see, well if that way works as well then that's good, but maybe it'd be good to but a notice with the :stories command that it won't download the stories of everyone the person is following if you're following too many people.

Thammus:
That is a great idea! If the length of the list of user IDs that gets fetched while using `:stories` is greater than or equal 150, Instaloader should display a warning that not all available stories may be fetched and suggest the other method to download them.

Issue #933, opened 2020-12-19
Possible solution of Too many requests?


roboshim:
Hello,
is someone getting "Too Many Requests" in Firefox too? I can browse many posts in browser and still works. Is there some hidden feature in web UI maybe? I have seen some special requests in Firefox:

`POST https://graph.instagram.com/logging_client_events`

`POST https://www.instagram.com/logging/falco`

But I'm no programmer, I don't unterstand that JS code. Could maybe someone of you check, when the FF sends these requests and if these requests maybe inform IG server about real client? Maybe you could send these requests too to fake real client and IG will then thing the instaloader is normal Firefox?

Thank you for your work!

Regards.

roboshim:
Hmmm, the problem 429 Too Many Requests is in iphone requests. If I replace `orig_url = self._iphone_struct['image_versions2']['candidates'][0]['url']` in structures.py on line 243 with `orig_url = self._full_metadata['display_url']` I can now download images.

Now it's failing 429 on sidecar.

Issue #946, opened 2020-12-28
Redirecting to login


harshjadon9:
*ubuntu
**command**
`instaloader --no-compress-json --no-metadata-json --no-captions --dirname-pattern=/django/instaloader/static/downloads/{profile} --filename-pattern={shortcode} -- -CJPM6GKJI_a`

**error**
```
HTTP redirect from https://www.instagram.com/graphql/query to https://www.instagram.com/accounts/login/
-CJPM6GKJI_a: Redirected to login page. Use --login.

Errors or warnings occurred:
-CJPM6GKJI_a: Redirected to login page. Use --login.
```

restyler:
In case you try to access public account and don't want to log in, try clean residential proxy or https://rapidapi.com/restyler/api/instagram40 which uses proxy network balancer and response health checks.

Issue #922, opened 2020-12-14
429 Too Many Requests and locked account


somera:
Your question here...

I'm using the current version (4.5.5). And I get often "429 Too Many Requests". After one day my Instagram account is locked. It's possible to setup sleep between every image download?

Showmaster3000:
There are different reasons why you reach error 429. You can find multiple discussions in the issue archive, f.e. #864 #889 #834
Instaloader has a internal mechanism to throttle down if it is near the request limit.
Furthermore you do not say anythiny about your setup. How many profiles do you download? How often do you check them? Are you logged in? Do you use a proxy? And some other questions...

somera:
> There are different reasons why you reach error 429. You can find multiple discussions in the issue archive, f.e. #864 #889 #834
> Instaloader has a internal mechanism to throttle down if it is near the request limit.
> Furthermore you do not say anythiny about your setup. How many profiles do you download? How often do you check them? Are you logged in? Do you use a proxy? And some other questions...

sorry.

I'm logged in. I start the script with

instaloader.exe --no-compress-json --login <myprofile> %%x -F

where %%x ist one profile from a list with ~200 profiles. I run it every 2-3 days.

Issue #918, opened 2020-12-12
Encountering the "429 - Too Many Requests" error with no instagram parallels running.


nafanko:
Hello everyone! I seem to have encountered this problem. 
I am not logged in to instagram through any browser or through instaloader. 
This error comes up as soon as I input my code.

>HTTP redirect from https://www.instagram.com/graphql/query to https://www.instagram.com/accounts/login/
JSON Query to graphql/query: 429 Too Many Requests: redirected to login [retrying; skip with ^C]
Requests within last 10/11/20/22/30/60 minutes grouped by type:
other: 1 1 1 1 1 1
472f257a40c653c64c666ce877d59d2b: 5 5 5 5 5 5
2b0673e0dc4580674a88d426fe00ea90: 1 1 1 1 1 1
97b41c52301f77ce508f55e66d17620e: 1 1 1 1 1 1
Instagram responded with HTTP error "429 - Too Many Requests". Please
do not run multiple instances of Instaloader in parallel or within
short sequence. Also, do not use any Instagram App while Instaloader
is running.
The request will be retried in 666 seconds, at 09:53.

Thank you for your help!

AplhaSlayer1964:
Make sure your account hasn't been locked by instagram. If it hasn't, try waiting half a day and try again. If you run instaloader for like an hour you will always run into 429 sadly there is no real work around besides waiting your request ban out. I'm sure the devs are working on a better way to get around the timeouts (there are like 100+ other people asking about the same 429 error).

nafanko:
> Make sure your account hasn't been locked by instagram. If it hasn't, try waiting half a day and try again. If you run instaloader for like an hour you will always run into 429 sadly there is no real work around besides waiting your request ban out. I'm sure the devs are working on a better way to get around the timeouts (there are like 100+ other people asking about the same 429 error).

Thank you for your reply. Would using a proxy help? Is the pull request ip related maybe? The weird thing is when I use the instaloader to just download photos from a page it'll do it, but when you use the python code to download top posts by user it crashes right away.

AplhaSlayer1964:
So if you are using the login it seems to be a combination of ip and account. If you switch your ip a lot you account can get flagged and you are using the same ip as another account it will flag you. I recommend having a seperate account and ip for instaloader but I would be careful about making a new account and right away and using it because they seem to flag new accounts worse. My bad solution for now is running a vm with a vpn in the background that way using instagram on my pc won't affect instaloader in any way.

[...]

PR #947, opened 2020-12-29
fix 429 Too Many requests: proxy balancer support for public profiles

restyler:
Fixes 429 rate limit via optional proxy balancer support.
Uses https://rapidapi.com/restyler/api/instagram40 balancer.

Fixes:
https://github.com/instaloader/instaloader/issues/933
https://github.com/instaloader/instaloader/issues/922
https://github.com/instaloader/instaloader/issues/946
https://github.com/instaloader/instaloader/issues/918
https://github.com/instaloader/instaloader/issues/883
https://github.com/instaloader/instaloader/issues/886
and many other cases.

Should be used (and was tested) for public profiles download.

Demo:
image

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -373,7 +373,7 @@ def main():
                         help='Disable user interaction, i.e. do not print messages (except errors) and fail '
                              'if login credentials are needed but not given. This makes Instaloader suitable as a '
                              'cron job.')
-    g_misc.add_argument('--rapidapi-key', metavar='SECRET_KEY', 
+    g_misc.add_argument('--rapidapi-key', metavar='SECRET_KEY',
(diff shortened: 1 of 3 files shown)


skeith:
Do this make request (per image) is delayed between each other? #944

restyler:
> Do this make request (per image) is delayed between each other? #944

No. But, if the content you need to download is public, this pr can help - just don't specify login details. Instagram won't be able to block it then

Thammus:
Not being able to access as many posts as you want without being logged in is a legitimate restriction on Instagram's part, and Instaloader is not supposed to circumvent this limitation with nasty hacks.
The PR does not actually solve the linked issues but introduces an unnecessary dependency on a morally reprehensible proprietary service.

Issue #969, opened 2021-01-13
getting incomplete caption of post or is there any way to get full caption with title?


17it151tejaschauhan:
I am using instaloader since four months,

It's great library for scrapping of instagram and handled most of possible exceptions of scrapping,

I would like to get your attention in below issue:

It gives incomplete caption of post,

Example:

Check this post in desktop browser :- https://www.instagram.com/p/CJ5wh77h92w/

To get details of this post, i have written following code:

`from instaloader import Instaloader, Post`
`L = Instaloader()`
`instagram_post_shortcode = "CJ5wh77h92w"`
`post = Post.from_shortcode(L.context, instagram_post_shortcode)`
`print(post.caption)`

Output:

`Watch India's Test specialist talk about India's special effort in the 3rd Test at SCG 🤩

🇦🇺🆚🇮🇳, 4th Test LIVE
🗓️ FRI, JAN 15
🕔 5 AM
📺 Sony SIX, Sony TEN 1, Sony TEN 3

#SirfSonyPeDikhega #SonySports #CheteshwarPujara #AUSvIND #INDvAUS #TeamIndia`

which is incomplete caption, here **"Test of Champions"** is missing in caption.

Complete caption is : 

`Test of Champions

Watch India's Test specialist talk about India's special effort in the 3rd Test at SCG 🤩

🇦🇺🆚🇮🇳, 4th Test LIVE
🗓️ FRI, JAN 15
🕔 5 AM
📺 Sony SIX, Sony TEN 1, Sony TEN 3

#SirfSonyPeDikhega #SonySports #CheteshwarPujara #AUSvIND #INDvAUS #TeamIndia`

PR #971, opened 2021-01-14
Added method to get title of post

17it151tejaschauhan:
On behalf of issue [#969](https://github.com/instaloader/instaloader/issues/969)

-- A motivation for this change
  - Fixes # 
  - Some post of instagram has title as same as caption but current version of instaloader does not give title of post.
  - To get title of post, I've added one property method called title as same as caption method in class Post.

-- The completeness of this change
 - In json response of graphql query, it gives title of post with key name as title as same as location in value of shortcode_media 
   but instaloader was not considering it.
 - Added one method named _title_ in class _Post_ which will get value of title from __full_metadata_dict_  dictionary variable.
 - If title is not present in json response, it will return None.
 - Need to update documentation for this method.
 - I consider it for ready to be merged.
 - Please guide me if necessary.

Example of use case:

`from instaloader.instaloader import Instaloader, Post`
`L = Instaloader()`
`post = Post.from_shortcode(L.context, "CKA_09eMMVE")`
`print("Title of Post : ", post.title)`

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -121,7 +121,9 @@ class Post:
     @property
     def title(self) -> Optional[str]:
         """Title of post"""
-        return self._full_metadata_dict.get("title", None)
+        if "title" in self._full_metadata_dict:
+            return self._full_metadata_dict["title"]
+        return None
 
     @property
     def shortcode(self) -> str:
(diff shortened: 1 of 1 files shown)


aandergr:
Same as #976.

Issue #986, opened 2021-01-20
extend Resumable Iterations to handle other interrupts other than KeyboardInterrupt


fsantagostinobietti:
HI,
**resumable_iteration** works great with KeyboardInterrupt, saving current iteration state so that you can resume it later.
But sometimes iteration is interrupted by some other exceptions (e.g. io error or login redirect) and in such cases iteration state is not saved.

Is it possible to modify resumable_iteration so that it can intercept any exception that occurs during iteration?

aandergr:
Please use our [Feature Suggestion Template](https://github.com/instaloader/instaloader/issues/new?labels=feature+suggestion&template=feature_request.md) for suggesting new features, as this makes it easier to see how the new feature could fit into Instaloader.

PR #990, opened 2021-01-21
Resumable Iteration info saved on any exception

fsantagostinobietti:
Feature enhancement:
When resumable iteration is interrupted by some other exceptions other than KeyboardInterrupt  (e.g. login redirect)  iteration state is not saved.  (issue  #986)

Change in this pull request:
Resumable_iteration now intercepts any exception occurred during iteration (not only KeyboardInterrupt ).

I have tested it locally and works as expected.
If accepted probably a note in documentation would be needed.

### DIFF
--- a/instaloader/nodeiterator.py
+++ b/instaloader/nodeiterator.py
@@ -262,7 +262,7 @@ def resumable_iteration(context: InstaloaderContext,
             context.error("Warning: Not resuming from {}: {}".format(resume_file_path, exc))
     try:
         yield is_resuming, start_index
-    except KeyboardInterrupt:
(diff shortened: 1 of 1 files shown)


aandergr:
This will not work. The exception triggered by e.g. login redirection is catched deep inside the download loop and doesn't reach the point where you changed the `except KeyboardInterrupt` to a simple `except`.

fsantagostinobietti:
uhmm ... this is strange, since I'm using this patch successfully in the last 2 weeks.
**LoginRequiredException** is caught and resume information file is store for later use.

PR #997, opened 2021-01-25
Issue of Skipped Items Being Counted Towards Actual Total Downloaded Items

vsengar26:
--------Motivation-------
Instaloader counts the items that are skipped too in the total number of downloaded items.

Like if a user asks to download 5 images of a hashtag using some post filters, Instaloader counts the skipped posts too. For example, if 3 posts are skipped and 2 are accepted and downloaded, then the user is getting only 2 posts but he demanded 5.

------ Changes Proposed ------
I have added a skippy variable that counts towards the total skipped posts and breaks if the actual posts that are downloaded are equal to those demanded.

------- Completeness----------
It's a minor change but creates a huge impact. Thus, no change in documentation is needed and you may consider the code as final.

You can help me by just saying thanks and since I am new to opensource, please forgive me if I messed up somewhere.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -14,7 +14,6 @@ from hashlib import md5
 from io import BytesIO
 from pathlib import Path
 from typing import Any, Callable, IO, Iterator, List, Optional, Set, Union, cast
-from urllib.parse import urlparse
 
 import requests
 import urllib3  # type: ignore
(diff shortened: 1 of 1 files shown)


Thammus:
Changing the behavior of the `--count` flag would break backward compatibility. Maybe the documentation should state clearer that skipped posts are also counted.

Issue #999, opened 2021-01-25
On rescans load existing GraphSidecar json instead of downloading again


thomas694:
**Provide us a use case of the feature**
For every GraphSidecar node that has videos in it, a graph query is executed. Especially on rescans that is a waste of api calls.

**Describe the solution you'd like**
Instead of downloading that json again, just load it from disk if it has been downloaded and saved before.

**Describe alternatives you've considered**
None found.

**If the feature request is accepted, would you be willing to submit a pull request?**
Yes.

**Additional context**

aandergr:
> Especially on rescans that is a waste of api calls.

We fixed this rescan inefficiency with 672b0b184019bf840e3003e90a1c662fcb4df79a in `upcoming/v4.7`. Now, that graph query is only executed if one of the files that should be downloaded is missing.

thomas694:
The current `upcoming/v4.7` ([640066d](https://github.com/instaloader/instaloader/commit/640066dd607f191eba6a2bf7cfb5446acda88490)) is runnable (and contains latest master too), right?

Sorry, I don't know how to use the upcoming version. Please help.
On a re-run it does a query for every sidecar node.
<br/>

As for every post the json is saved anyway (default parameters), so why not reload that one, if needed in case of a sidecar node?
For a re-run of 62 posts (12+50) I need two query/api calls, regardless of how many sidecar nodes therein. And it works independent from the chosen filename template. Imho that's effiecency.

aandergr:
This feature suggestion seems a great idea to me.

Instaloader goes through the posts online, determines the filenames that would be created for each post, and skips the post if it already exists in the target's folder locally. For sidecars in a profile that are not at the first page, e.g. posts with index > 12, the sidecar details are not included in the page overview returned by Instagram, and so Instaloader currently needs to do an extra graphql query to obtain the information that is needed to determine which filenames would be created, and as such, to determine whether something has to be downloaded. (my previous comment was wrong)

The information in the 'old' JSON files from a previous run should be sufficient to determine the filenames yield by the Post in question. So your suggestion is a nice idea to save that query and so speed up rescans of profiles (and most probably other targets as well) that have some sidecars.

I think that the number of media in a sidecar post cannot change, i.e. that it is not possible to add or remove photos/videos after posting the sidecar. This is a prerequisite for this feature to work properly.

To not introduce side-effects, the information from the 'old' JSON file probably should only be used to determine whether the Post needs to be re-downloaded or not.

[...]

PR #1015, opened 2021-02-07
Reload graphsidecar json

thomas694:
Thanks for your willingness to contribute to Instaloader! Please briefly
describe

  - Implement feature issue #999
    Load existing GraphSidecar json instead of downloading again. Previously for every GraphSidecar node that had videos in it, a graph query was executed even the json has been already downloaded before.

    Instead of downloading that json again, just load it from disk if it has been downloaded and saved before.

  - Is it just a proof of concept? _No_
  - Is the documentation updated (if appropriate)? _not needed_
  - Do you consider it ready to be merged or is it a draft? _ready_
  - Can we help you at some point? _please review_

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -577,7 +577,7 @@ class Instaloader:
                 if post.has_sidecar_video:
                     postloaded = self.load_metadata_json(filename)
                     if postloaded is Post:
-                        post = postloaded
+                        post = cast(Post, postloaded)
                 for edge_number, sidecar_node in enumerate(
                         post.get_sidecar_nodes(self.slide_start, self.slide_end),
                         start=post.mediacount if self.slide_start < 0 else self.slide_start + 1
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -289,6 +289,21 @@ class Instaloader:
         os.utime(filename, (datetime.now().timestamp(), mtime.timestamp()))
         return True
 
+    def load_metadata_json(self, filename: str) -> JsonExportable:
+        """Load metadata JSON file of a structure."""
+        fn = filename + '.json.xz'
+        if Path(fn).is_file():
+            structure = load_structure_from_file(self.context, fn)
+            if isinstance(structure, (Post, StoryItem)):
+                return structure
+        else:
+            fn = filename + '.json'
+            if Path(fn).is_file():
+                structure = load_structure_from_file(self.context, fn)
+                if isinstance(structure, (Post, StoryItem)):
+                    return structure
+        return None
+
     def save_metadata_json(self, filename: str, structure: JsonExportable) -> None:
         """Saves metadata JSON file of a structure."""
         if self.compress_json:
@@ -559,6 +574,10 @@ class Instaloader:
         downloaded = True
         if post.typename == 'GraphSidecar':
             if self.download_pictures or self.download_videos:
+                if post.has_sidecar_video:
+                    postloaded = self.load_metadata_json(filename)
+                    if postloaded is not None:
+                        post = postloaded
                 for edge_number, sidecar_node in enumerate(
                         post.get_sidecar_nodes(self.slide_start, self.slide_end),
                         start=post.mediacount if self.slide_start < 0 else self.slide_start + 1
@@ -568,18 +587,18 @@ class Instaloader:
                         if '{filename}' in self.filename_pattern:
                             suffix = ''
                         # pylint:disable=cell-var-from-loop
-                        filename = self.__prepare_filename(filename_template, lambda: sidecar_node.display_url)
+                        filename2 = self.__prepare_filename(filename_template, lambda: sidecar_node.display_url)
                         # Download sidecar picture or video thumbnail (--no-pictures implies --no-video-thumbnails)
-                        downloaded &= self.download_pic(filename=filename, url=sidecar_node.display_url,
+                        downloaded &= self.download_pic(filename=filename2, url=sidecar_node.display_url,
                                                         mtime=post.date_local, filename_suffix=suffix)
                     if sidecar_node.is_video and self.download_videos:
                         suffix = str(edge_number)
                         if '{filename}' in self.filename_pattern:
                             suffix = ''
                         # pylint:disable=cell-var-from-loop
-                        filename = self.__prepare_filename(filename_template, lambda: sidecar_node.video_url)
+                        filename2 = self.__prepare_filename(filename_template, lambda: sidecar_node.video_url)
                         # Download sidecar video if desired
-                        downloaded &= self.download_pic(filename=filename, url=sidecar_node.video_url,
+                        downloaded &= self.download_pic(filename=filename2, url=sidecar_node.video_url,
                                                         mtime=post.date_local, filename_suffix=suffix)
         elif post.typename == 'GraphImage':
             # Download picture
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -289,7 +289,7 @@ class Instaloader:
         os.utime(filename, (datetime.now().timestamp(), mtime.timestamp()))
         return True
 
-    def load_metadata_json(self, filename: str) -> JsonExportable:
+    def load_metadata_json(self, filename: str) -> Optional[JsonExportable]:
         """Load metadata JSON file of a structure."""
         fn = filename + '.json.xz'
         if Path(fn).is_file():
@@ -576,7 +576,7 @@ class Instaloader:
             if self.download_pictures or self.download_videos:
                 if post.has_sidecar_video:
                     postloaded = self.load_metadata_json(filename)
-                    if postloaded is not None:
+                    if postloaded is Post:
                         post = postloaded
                 for edge_number, sidecar_node in enumerate(
                         post.get_sidecar_nodes(self.slide_start, self.slide_end),
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -566,6 +566,27 @@ class Instaloader:
         :return: True if something was downloaded, False otherwise, i.e. file was already there
         """
 
+        def _already_downloaded(path: str) -> bool:
+            if not os.path.isfile(path):
+                return False
+            else:
+                self.context.log(path + ' exists', end=' ', flush=True)
+                return True
+
+        def _all_already_downloaded(path_base, is_videos_enumerated) -> bool:
+            if '{filename}' in self.filename_pattern:
+                # full URL needed to evaluate actual filename, cannot determine at
+                # this point if all sidecar nodes were already downloaded.
+                return False
+            for idx, is_video in is_videos_enumerated:
+                if self.download_pictures and (not is_video or self.download_video_thumbnails):
+                    if not _already_downloaded("{0}_{1}.jpg".format(path_base, idx)):
+                        return False
+                if is_video and self.download_videos:
+                    if not _already_downloaded("{0}_{1}.mp4".format(path_base, idx)):
+                        return False
+            return True
+
         dirname = _PostPathFormatter(post).format(self.dirname_pattern, target=target)
         filename_template = os.path.join(dirname, self.format_filename(post, target=target))
         filename = self.__prepare_filename(filename_template, lambda: post.url)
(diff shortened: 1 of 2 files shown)


thomas694:
I should get my pylint/mypy running here ;)

instaloader/instaloader.py
```diff
@@ -580,45 +574,41 @@ def _all_already_downloaded(path_base, is_videos_enumerated) -> bool:
...
+                        filename2 = self.__prepare_filename(filename_template, lambda: sidecar_node.video_url)
+                        # Download sidecar video if desired
+                        downloaded &= self.download_pic(filename=filename2, url=sidecar_node.video_url,
+                                                        mtime=post.date_local, filename_suffix=suffix)
         elif post.typename == 'GraphImage':
             # Download picture
             if self.download_pictures:
-                downloaded = (not _already_downloaded(filename + ".jpg") and
-                              self.download_pic(filename=filename, url=post.url, mtime=post.date_local))
+                downloaded = self.download_pic(filename=filename, url=post.url, mtime=post.date_local)
```
aandergr, line 606: This not _already_downloaded check was to make sure that the URL is not accessed (which can trigger an extra request) if the file is already downloaded.

instaloader/structures.py
```diff
@@ -150,10 +152,11 @@ def __hash__(self) -> int:
     def _obtain_metadata(self):
         if not self._full_metadata_dict:
             pic_json = self._context.graphql_query(
-                '2b0673e0dc4580674a88d426fe00ea90',
+                'a9441f24ac73000fa17fe6e6da11d59d',
                 {'shortcode': self.shortcode}
             )
             self._full_metadata_dict = pic_json['data']['shortcode_media']
+            self._node['is_loaded_graphsidecar'] = '1'
```
aandergr, line 159: What is this needed for?

aandergr (commented):
This changes a lot more than the changes described in your PR's description or in #999. Some of the changes just revert the recent changes of `upcoming/v4.7` as if this was accidentally not properly rebased before pushing. It also includes new graphql query hashes, which is probably not needed.

thomas694:
> This changes a lot more than the changes described in your PR's description or in #999. Some of the changes just revert the recent changes of `upcoming/v4.7`

This... At least for logged out users, that commit is not doing what it promises. Actually it's doing much more queries than before on a re-run.

> It also includes new graphql query hashes, which is probably not needed.

I copied the block from my version and forgot that I already changed the query hash to be up to date. Instaloader is using outdated query hashes as you can see by using a network sniffer or chrome's developer mode when browsing manually. It seems the IG api doesn't care much if you provide an old/wrong query hash and just runs the real query for the parameters, but maybe that gives additional penalties to our session limit.

instaloader/instaloader.py
thomas694: Ahh, now I see it. The problem was for logged in users, because of a query needed for `_iphone_struct` (high-res images). As I told before, I run it with default parameters, so not logged in. That's why that query didn't catch my eye.

Inside the `Post` constructor there has already been code to reload the `iphone_struct`. But isn't the real problem that no one ever loads the structure from disk!? I'm doing it now, so `iphone_struct` should be loaded too, isn't it? Can you please grab the branch and verify it?

instaloader/structures.py
thomas694: That's necessary because the iterator's `Post` object is not the full `Post` object loaded by the post url. Only in the latter case the Post constructor shall overwrite the `_full_metadata_dict` as would have been done by a download.

aandergr: Would anything bad happen if we did not overwrite the `_full_metadata_dict`?

instaloader/instaloader.py
aandergr: > I'm doing it now, so `iphone_struct` should be loaded too, isn't it?

`iphone_struct` is saved and loaded too with `save_structure_to_file()`/`load_structure_from_file()`.

aandergr:
Thanks for leaving the new query hashes out for now.

There might be valid reasons to change the query hashes and so on, and I am happy that you have noticed these things, but it is important for us to keep different changes separate, as this allows a much more efficient development and review process.

instaloader/instaloader.py
```diff
@@ -596,15 +615,15 @@ def _all_already_downloaded(path_base, is_videos_enumerated) -> bool:
                             suffix = None
                         if self.download_pictures and (not sidecar_node.is_video or self.download_video_thumbnails):
                             # pylint:disable=cell-var-from-loop
-                            filename = self.__prepare_filename(filename_template, lambda: sidecar_node.display_url)
+                            filename2 = self.__prepare_filename(filename_template, lambda: sidecar_node.display_url)
```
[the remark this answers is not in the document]
thomas694: I renamed this variable, because the filename is generated before the loop in [L592](https://github.com/instaloader/instaloader/blob/ec6d1f1453b8949acd989426b285fdbae579a683/instaloader/instaloader.py#L592), then for sidecar nodes the filename would be changed inside ([L618](https://github.com/instaloader/instaloader/blob/ec6d1f1453b8949acd989426b285fdbae579a683/instaloader/instaloader.py#L618), [L624](https://github.com/instaloader/instaloader/blob/ec6d1f1453b8949acd989426b285fdbae579a683/instaloader/instaloader.py#L624)) depending on which file hasn't been already downloaded and/or which slides to download, and after the loop the (post) filename is used for saving caption, metadata and so on.
So the sidecar code should use another variable and not change the post's filename, or not?

instaloader/instaloader.py
```diff
@@ -580,6 +595,10 @@ def _all_already_downloaded(path_base, is_videos_enumerated) -> bool:
         downloaded = True
         if post.typename == 'GraphSidecar':
             if self.download_pictures or self.download_videos:
+                if post.has_sidecar_video:
+                    postloaded = self.load_metadata_json(filename)
+                    if postloaded is Post:
+                        post = cast(Post, postloaded)
```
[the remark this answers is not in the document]
thomas694: You are right. Sadly old IG posts can be edited, at least caption, tags, location and alt text.

I still have to look at it again.

instaloader/structures.py
thomas694: What do you mean? When the post is deserialized from disk, then I recognize it in the constructor and overwrites the `_full_metadata_dict` just as if a download just had happened. Of course, for a download `_full_metadata_dict` is overwritten in `_obtain_metadata` as before. Without doing that, `_field`/`_full_metadata` could trigger a new (unwanted) download.

thomas694:
Why we don't just use a new query_hash?

Urls are from [https://www.instagram.com/cats_of_instagram/](https://www.instagram.com/cats_of_instagram/).

This one is currently used in Instaloader:
[https://www.instagram.com/graphql/query/?query_hash=472f257a40c653c64c666ce877d59d2b&variables=%7B%22id%22%3A%227013409%22%2C%22first%22%3A50%2C%22after%22%3A%22QVFBNmlPTlE4YU15U0R1X1otUVNuSFVxY3M2QWVqbWlfcFBVVGNyY0NQeFR4Nkp6RVR3Vl9TV016T3UyY3kyTm1Ib2h2NGw5aDBXc3dMUmFiVWY5M0J3Vg%3D%3D%22%7D](https://www.instagram.com/graphql/query/?query_hash=472f257a40c653c64c666ce877d59d2b&variables=%7B%22id%22%3A%227013409%22%2C%22first%22%3A50%2C%22after%22%3A%22QVFBNmlPTlE4YU15U0R1X1otUVNuSFVxY3M2QWVqbWlfcFBVVGNyY0NQeFR4Nkp6RVR3Vl9TV016T3UyY3kyTm1Ib2h2NGw5aDBXc3dMUmFiVWY5M0J3Vg%3D%3D%22%7D)

Now look here at the current one from the website:
[https://www.instagram.com/graphql/query/?query_hash=003056d32c2554def87228bc3fd9668a&variables=%7B%22id%22%3A%227013409%22%2C%22first%22%3A50%2C%22after%22%3A%22QVFBNmlPTlE4YU15U0R1X1otUVNuSFVxY3M2QWVqbWlfcFBVVGNyY0NQeFR4Nkp6RVR3Vl9TV016T3UyY3kyTm1Ib2h2NGw5aDBXc3dMUmFiVWY5M0J3Vg%3D%3D%22%7D](https://www.instagram.com/graphql/query/?query_hash=003056d32c2554def87228bc3fd9668a&variables=%7B%22id%22%3A%227013409%22%2C%22first%22%3A50%2C%22after%22%3A%22QVFBNmlPTlE4YU15U0R1X1otUVNuSFVxY3M2QWVqbWlfcFBVVGNyY0NQeFR4Nkp6RVR3Vl9TV016T3UyY3kyTm1Ib2h2NGw5aDBXc3dMUmFiVWY5M0J3Vg%3D%3D%22%7D)

Especially look on node indexes 3 (GraphSidecar, CLAIIh5lwWa) and 4 (GraphVideo, CK_znJ_nxYa).
Do you see any difference?

aandergr:
Cool! That sounds like a better alternative. Would you make a PR for that?

instaloader/instaloader.py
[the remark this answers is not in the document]
aandergr: Oh indeed. This means that when using `{filename}` in a filename pattern, the files that are created below the loop (caption txt, metadata json, etc.) are created with a wrong filename.

Could you make a PR for this change separately? Also, could we find a more descriptive name than `filename2`?

thomas694:
> Cool! That sounds like a better alternative. Would you make a PR for that?

Sure. I'll do a new PR to update this query_hash. The new PR can be based on master, or not?

Of course, if the query_hash will be changed by IG in a few weeks/months again, Instaloader is at the same point as today, it needs working fallbacks (like already implemented for videos and sidecars) to still work in an acceptable way until the next release with updated query_hash comes out.
<br/>

Before more resarch on the query_hash I came to the following, based on the situation from before (outdated query_hash):

Getting eventually new data (caption, tags, location or alt text) and reloading the old post as a means of caching contradicts somehow. Especially as only the first 12 posts contain "full" data. From the second page on there is no more complete data, also the sidecar nodes are missing. So from the 13th post on, it would be necessary anyway to download the full post to be able to update the (maybe) changed four data items.

For the same reasons your mentioned fix triggers the queries for the sidecar nodes from page two onwards. So for checking if a file needs to be downloaded, it already triggers a download of the full post. Somehow pointless then.

Without doing a post download to get the complete data only an update of the caption could have been done. It would have been necessary to first update the disk loaded post with the page's post details, before using it. But it would have been necessary to load and keep the "cached" post, as the full details shall not be lost on another run and the post is normally saved at the end of the `download_post`.

As alternative my suggestion would have been to implement it as an additional option which, if chosen, speeds up the process resp. uses less api calls (knowingly that it doesn't update captions/texts), but keeps the default behavior as before. The new option's name still could have been chosen.

instaloader/instaloader.py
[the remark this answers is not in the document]
thomas694: created PR #1025

instaloader/structures.py
aandergr: Ah I see. Thanks for sharing your observation!

f85e49d521ca6a838c23213b42acc8c83e9a7e6e (currently only in master) fixes that unnecessary fetch triggered in `get_sidecar_nodes()`. This should also make your change in `Post._obtain_metadata()` and `Post.__init__()` unnecessary.

`_full_metadata` triggers the metadata fetch on a Post created with `load_structure_from_file()`; however, `_full_metadata` is not supposed to be called unless we can be sure that the required data is not available in `_node`. `_field()` does not do a full metadata fetch if the required data is available in `_node`.

aandergr:
The information in the 'old' JSON files from a previous run is sufficient to determine the filenames yield by the Post in question. So your suggestion is a nice idea to spare the extra full metedata query that sometimes is needed to download the Post and so speed up rescans of some targets (such as e.g. `:saved`) that have some sidecars.

To not introduce side-effects, the information from the 'old' JSON file should only be used to determine whether the Post needs to be re-downloaded or not. It should not be used for *anything* else, to not use outdated information.

There should not be any changes in how the caption text file etc. are generated. Leaving this as is will in most scenarios not trigger an extra query.

So the suggestion in #999 will (in the majority of use cases) significantly speed up going through a mostly-already-downloaded target that contains many sidecar posts.

aandergr:
I accidentally closed this pull request by removing the branch it based on (originally `upcoming/v4.7`). It is now reopened and based on the new `upcoming/v4.8` branch. Sorry for the inconvenience.

aandergr:
@thomas694 Are you still interested to work on this PR?

thomas694:
Sorry, after finding out that the most easy way to get it working is to update the query_hash once in a while and that a workaround that doesn't interfere with future data updates needs a little bit more work, it has been forgotten.

Aside from that my linux/python system is broken since several weeks and I haven't had the time and the luck to fix it yet.
So I'll close it for the time being.

aandergr:
Alright, thanks for the update! Good luck with fixing your system!

PR #1025, opened 2021-02-14
Fix issue 892 Save content using original filenames (PR 893)

thomas694:
- When a post contained sidecars, the filename for caption, json etc. was changed. It was no longer the original filename like before.
- If the post contains sidecars, a local variable is used to build the filenames for the sidecar media.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -563,7 +563,7 @@ class Instaloader:
                         post.get_sidecar_nodes(self.slide_start, self.slide_end),
                         start=post.mediacount if self.slide_start < 0 else self.slide_start + 1
                 ):
-                    suffix = str(edge_number)
+                    suffix: Optional[str] = str(edge_number)
                     if '{filename}' in self.filename_pattern:
                         suffix = None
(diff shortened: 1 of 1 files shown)


instaloader/instaloader.py
```diff
@@ -563,23 +563,20 @@ def download_post(self, post: Post, target: Union[str, Path]) -> bool:
                         post.get_sidecar_nodes(self.slide_start, self.slide_end),
                         start=post.mediacount if self.slide_start < 0 else self.slide_start + 1
                 ):
+                    suffix = str(edge_number)
+                    if '{filename}' in self.filename_pattern:
+                        suffix = ''
```
aandergr, line 568: I think this should be:

```suggestion
                    suffix = str(edge_number)  # type: Optional[str]
                    if '{filename}' in self.filename_pattern:
                        suffix = None
```

Because if suffix is not None, download_pic will add `_` to the filename:

https://github.com/instaloader/instaloader/blob/828fa71edee9cbe82b00fe2591a6691b9000b3cc/instaloader/instaloader.py#L282-L283

thomas694: correct

aandergr:
Thanks!

PR #1052, opened 2021-03-11
Max count for profile post downloads

miclejackson12:
* incorporate max_count

* incorporate max_count


aandergr:
This would unfortunately change Instaloader's behavior in a backward-incompatible manner, since an invocation with --count and a mix of hashtags/feed/location and a profile would now behave differently in that the max count now also affects the profile and not only hashtags/feed/location.

Also, it is quite trivial for the user to only download a certain count of posts of a profile, e.g. by calling download_post() for each post of an islice()'d profile.get_posts(). There is no need to include that in the CLI.

PR #1072, opened 2021-03-30
Fixed Profile Command

iamjawa:
The word profile is no longer required, this simply searches for an Instagram profile with the username of "profile" instead of being an argument

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -72,7 +72,7 @@ To **download all pictures and videos of a profile**, as well as the
 
 ::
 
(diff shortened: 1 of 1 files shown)


Thammus:
The word profile was never required. The syntax `instaloader profile [profile ...]` follows exactly the example from Python's ArgumentParser documentation:
https://docs.python.org/3/library/argparse.html#usage
Especially the reuse of the same argument name is perfectly in accordance.
We do not want to deviate from the current description, because it follows a well established convention.

PR #1076, opened 2021-04-04
Make iPhone endpoint query optional

### LINKED ISSUE #1035: [Regression] Story Downloading is artificially rate limited
**Describe the bug**
Using 4.5.5, I would never hit a rate limit, but with 4.6.2, I get the "Too many queries in the last time" message even though I have always downloaded stories fine without the rate controller limiting me.

**To Reproduce**
1. Use 4.6.2 (:stories) to try to download the stories of around 100 accounts, or just a lot of stories in general.

**Expected behavior**
I don't get rate limited and it lets me download the stories.

"Too many queries in the last time, wait 20 minutes" or something similar.

**Instaloader version**
4.6.2

**Additional context**
I use a custom setup to download all the stories from my account (:stories with marking stories as read), and it has always worked fine to download all the stories from my account. The rate controller started to limit me on 4.6.2, and downgrading to 4.5.5 seems to have fixed the issue, so it seems the rate controller is incorrectly rate limiting story downloading when it fact it's almost impossible to hit the limit under normal usage. I follow over 1.5k accounts and I've never been rate limited with my setup before now.

aboredpervert:
This makes the support for calling the iPhone media endpoint optional, using the `--no-iphone` argument.
Should fix #1035. The motivation for this change is that no matter what I did before, always got 429s.

### DIFF
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -277,6 +277,8 @@ How to Download
 
    Do not attempt to download iPhone version of images and videos.
 
+   .. versionadded:: 4.8
+
 Miscellaneous Options
 ^^^^^^^^^^^^^^^^^^^^^
 
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -100,7 +100,6 @@ What to Download of each Post
 
    Do not xz compress JSON files, rather create pretty formatted JSONs.
 
-
 What to Download of each Profile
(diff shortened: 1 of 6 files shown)


Thammus (commented):
Thanks for this great Pull Request. The changes look really good and thoroughly thought through. Just one little improvement: The documentation of the option `--no-iphone` does better fit into the "How to download" category instead of "What to Download of each Post".

aboredpervert:
Thanks, I have moved it in rebase.

PR #1107, opened 2021-04-20
Adding feature to retrieve source files URLs instead of downloading them

alistvt:
  previously there weren't any methods inside the Instaloader class to let us retrieve the files' source URLs, but here I added some methods to add this functionality. It is helpful because sometimes it is not efficient to download the files and we can just place the posts' source URLs inside our code.

  added fetch_post_src_urls
  added fetch_post_caption
  added fetch_stories_src_urls
  added fetch_story_item_from_shortcode

  - Do you consider it ready to be merged or is it a draft? I recommend merging it 
  - Can we help you at some point? Yes, please review.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -574,6 +574,15 @@ class Instaloader:
         .. versionadded:: 4.1"""
         return _PostPathFormatter(item).format(self.filename_pattern, target=target)
 
+    def fetch_post_caption(self, post: Post) -> str:
+        """
+        Get post's caption
+
+        :param post: Post to fetch its caption.
+        :return: Caption of the post in string format.
+        """
+        metadata_string = _ArbitraryItemFormatter(post).format(self.post_metadata_txt_pattern).strip()
+        return metadata_string
 
     def fetch_post_src_urls(self, post: Post) -> List((str, str)):
         """
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -870,7 +870,7 @@ class Instaloader:
 
         return sources
 
-    def fetch_story_item_src_url(self, item: StoryItem) -> List[Tuple[str, Optional[str]]]:
+    def fetch_story_item_src_url(self, item: StoryItem) -> List[Tuple[str, str]]:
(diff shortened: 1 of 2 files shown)


aandergr:
> * A motivation for this change, e.g.
>   
>   * More general: What problem does the pull request solve?
>     previously there weren't any methods inside the Instaloader class to let us retrieve the files' source URLs, but here I added some methods to add this functionality. It is helpful because sometimes it is not efficient to download the files and we can just place the posts' source URLs inside our code.

Instaloader does provide methods to retrieve the source URLs. Yes, it is not the Instaloader class itself that provides these methods, rather they are properties in the classes Post and StoryItem. They can be easily invoked from outside of Instaloader. What would be the advantage of having methods inside the Instaloader class to let us retrieve the files' source URLs?

alistvt:
Thanks for your immediate review. @aandergr 
Well, first and foremost I needed to implement these methods for my own projects and then I thought it would be helpful if we add these methods inside the instaloader class, because it prevents the developers from reading the instaloader's Post and StoryItem codes and writing these methods again for theirs own. 
Technically speaking, according to the [encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)) and [abstraction](https://www.guru99.com/java-data-abstraction.html#:~:text=Abstraction%20is%20the%20concept%20of,unnecessary%20details%20from%20the%20users.&text=It%20is%20one%20of%20the%20most%20important%20concepts%20of%20OOPs.), which are two of the most important paradigms of OOP, it is better to prevent the users of the classes to access the objects' attributes and instead, it is better to write different methods which let the user do the task.
This is helpful because for example if we change the instaloader's implementation, the users of the library do not need to change their own codes, because they just use some methods which are not changed.

aandergr:
I see that this might be needed by many users, however I think this kind of functionality is better to keep on the caller side rather than the library itself, especially since all the needed information can be likewise accessed and assembled from the outside. We always maintain backwards compatibility between Instaloader upgrades and thanks to using property interfaces in the representation classes of the Instagram structures we can continue to do so without any worries should something need to be changed.

Issue #253, opened 2019-02-10
Split filename-pattern into post-pattern and storyitem-pattern


aandergr:
**Provide us a use case of the feature**
When downloading a profile's posts and stories with `--stories`, posts and stories get mixed up in the same directory. The user should have the possibility to instruct Instaloader to save them in different places.

**Describe the solution you'd like**
Options `--post-pattern` and `--storyitem-pattern` that define the filename pattern for Post and StoryItem respectively, rather than the current `--filename-pattern`, which is used for both types of media. For backwards compatibility, `--filename-pattern` should then set both the post pattern and the storyitem pattern.

**Describe alternatives you've considered**
- Download posts and stories in separate invocations of Instaloader.
- The one proposed in #245 (option `--story-folder` which prefixes the filename pattern with `story/` for all storyitems).

**Additional context**
> I like the idea of giving the user the choice to have stories saved in another directory than regular posts for a profile without having to use separate Instaloader invocations. However, your implementation also creates a 'stories' folder when downloading with `:stories` and when downloading highlights, and adds a special case to the file naming logic which is handled with dirname-pattern and filename-pattern. We could give the user the option to store posts and stories in different directories by splitting the filename-pattern into a post-pattern and storyitem-pattern, which would be a more elegant and flexible solution.

_Originally posted by @aandergr in https://github.com/instaloader/instaloader/pull/245#issuecomment-462146645_

github-userx:
@aandergr any update on this? Hundreds of stories get mixed into the regular posts directory right now when I download both

PR #1112, opened 2021-04-23
[WIP] [Help needed] Stories are now saved in a separate folder

### LINKED ISSUE #1108: [Feature] Save stories in a separate folder
It is currently difficult to distinguish stories from normal posts. To avoid this it would be helpful to save all stories in an extra folder, like it happens with tagged-posts.

**Additional context**
This would also fix the bug that if a story has the same name as a post the post is not saved and the download aborts if "--fast-update" was used.

rm1410:
## Motivation:
* It is currently difficult to distinguish stories from normal posts because they are mixed together.
* The download aborts if a story and a post have the same name and the --fast-update attribute is passed, because the file technically already exists.

## Changes:
* Stories are now saved in a separate folder with the name :stories

## Other
* Not ready to merge
* Documentation may need to be updated.

close #1108 
maybe #253

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -745,7 +745,8 @@ class Instaloader:
                 self.context.log("[%3i/%3i] " % (count, totalcount), end="", flush=True)
                 count += 1
                 with self.context.error_catcher('Download story from user {}'.format(name)):
-                    downloaded = self.download_storyitem(item, filename_target if filename_target else name)
+                    target: Path = Path(filename_target.format(name))
+                    downloaded = self.download_storyitem(item, target if target else name)
                     if fast_update and not downloaded:
                         break
 
@@ -765,8 +766,7 @@ class Instaloader:
                 return True
 
         date_local = item.date_local
-        dirname = (Path(_PostPathFormatter(item).format(self.dirname_pattern, target=target)) /
-                  _PostPathFormatter.sanitize_path(':stories'))
+        dirname = _PostPathFormatter(item).format(self.dirname_pattern, target=target)
         filename_template = os.path.join(dirname, self.format_filename(item, target=target))
         filename = self.__prepare_filename(filename_template, lambda: item.url)
         downloaded = False
@@ -1293,7 +1293,8 @@ class Instaloader:
         if stories and profiles:
             with self.context.error_catcher("Download stories"):
                 self.context.log("Downloading stories")
-                self.download_stories(userids=list(profiles), fast_update=fast_update, filename_target=None,
+                self.download_stories(userids=list(profiles), fast_update=fast_update,
+                                      filename_target="{}/:stories",
                                       storyitem_filter=storyitem_filter)
 
     def download_profile(self, profile_name: Union[str, Profile],
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -765,7 +765,8 @@ class Instaloader:
                 return True
 
         date_local = item.date_local
-        dirname = Path(_PostPathFormatter(item).format(self.dirname_pattern, target=target)) / _PostPathFormatter.sanitize_path(':stories')
+        dirname = (Path(_PostPathFormatter(item).format(self.dirname_pattern, target=target)) /
+                  _PostPathFormatter.sanitize_path(':stories'))
         filename_template = os.path.join(dirname, self.format_filename(item, target=target))
         filename = self.__prepare_filename(filename_template, lambda: item.url)
         downloaded = False
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -765,7 +765,7 @@ class Instaloader:
                 return True
 
         date_local = item.date_local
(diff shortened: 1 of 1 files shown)


rm1410:
I don't know why the build is failing. There are no problems for me locally.

instaloader/instaloader.py
```diff
@@ -745,7 +745,8 @@ def download_stories(self,
                 self.context.log("[%3i/%3i] " % (count, totalcount), end="", flush=True)
                 count += 1
                 with self.context.error_catcher('Download story from user {}'.format(name)):
-                    downloaded = self.download_storyitem(item, filename_target if filename_target else name)
+                    path_target: Path = Path(filename_target.format(name))
```
rm1410, line 748: [Help needed]
Error: `instaloader/instaloader.py:748: error: Item "None" of "Optional[str]" has no attribute "format"`

Isn't it replaced with ":stories" when None is passed?

Thammus:
Thanks for your effort! Unfortunately though, this changes the destination folder for stories in a way that is not backwards compatible, which sadly makes it impossible for us to merge it.

rm1410:
@Thammus Thanks for the feedback! However, I do not think that this PR would break backwards compatibility. In the case of an update, there would be no errors and old stories would still be unsorted along with posts. However, all new stories would be sorted into the new folder. I would see this more as an advantage than a disadvantage, since the stories are sorted after a certain point.

Thammus:
> However, all new stories would be sorted into the new folder.

This changes the default behavior for downloading new stories which will confuse users and dependent software expecting the stories to be downloaded to their usual location. A possible way to implement such a feature in a backward compatible manner could be to split `--filename-pattern` into `--post-pattern` and `--storyitem-pattern` like suggested in #253.

rm1410:
@Thammus I have noticed that lately many Pull Requests which would bring good additional features and improvements have been closed because of backwards compatibility. And while I understand that you want to keep a version which has full backwards compatibility, I thought that it is sad to not be able to use such improvements.
So I have gotten an idea. What about if you create a 5.x branch where you could accept all such pull requests and which doesn't have the requirement of full backwards compatibility. The current 4.x version branch is still the default and can get new features, but there would be the alternate option of a new improved 5.x version which doesn't provide backwards compatibility to the 4.x versions anymore, but therefore can have many good new features and improvements. New features and changes that don't break backwards compatibility could be added to both version branches (e.g. via rebase).
After a longer period, e.g. one year, the 5.x version could get the new default version, with the 4.x legacy version still being available for people who want to use it. Programs or dependencies can still use that version, and with a longer period of support for both versions, they get much time to adapt to the changes if they want to switch to the 5.x one.
I think that this would be a good method of combining both the need for backwards compatibility while not having to reject new features that would break those.

Thammus:
> I have noticed that lately many Pull Requests which would bring good additional features and improvements have been closed because of backwards compatibility.

I just reviewed all closed Pull Request of the past year and only found two trivial ones where backwards compatibility was of concern. Also the changes of those PRs can hardly be seen as a feature or improvement. They change the behavior according to subjective preferences but do not generally improve it. These PRs were poorly thought through and would also have been rejected if backwards compatibility was not a point. This is why generally it is recommended to discuss new features in a feature suggestion issue beforehand.

Version 4 of Instaloader has proven to be extremely extensible. For the issue of separating stories from other posts there is already a clean solution outlined in #253.

rm1410:
@Thammus I'm sorry, I made a mistake there. I think the problem is that we both see the purpose of Instaloader differently. I don't use it through another program, but directly through the CLI. You want it to be as flexible as possible so other programs can use it well. I can understand that and will close the PR after you read the message, as I don't have the time to program this feature at the moment.

Thammus:
The problem is that this PR would change the default behavior for downloading new stories in a way that is not backwards compatible which will confuse users and require to change dependent software. This point is independent of the purpose of Instaloader or whether it gets used through another program or through CLI.

rm1410:
@Thammus 
> require to change dependent software


But isn't that exactly my point? If you saw the purpose of instaloader in just using it directly through CLI, the change would make sense as then the default would be something more expected.
But you prefer having a more complex but powerful configuration (which makes more sense to use via other programs then the CLI itself) and having the default not to change because it would **require to change the dependent software**. So if you see it from that way and see the purpose of instaoader in being used by other software, this approach with having more detailed configuration options (that are easy to use with a program but not via every cli command) completely makes sense, so I think our different opinions are based in that; seeing the purpose differently.

aandergr:
> If you saw the purpose of instaloader in just using it directly through CLI, the change would make sense as then the default would be something more expected.

Not at all. It is most expected that Instaloader behaves the way it always did. Changing the folder where stories are saved is in no way *something more expected*. Neither it is more convenient; it may be for you, but it is not in general.

> this approach with having more detailed configuration options (that are easy to use with a program but not via every cli command) completely makes sense

That is not the point. The approach suggested in #253 is especially easier to use via CLI command.

PR #1139, opened 2021-05-16
update documentation for --count option

SelfAdjointOperator:
## Motivation

The documentation for `--count` was out of date in the changed files. It has been corrected to note that `--count` works for the four options `#hashtag`, `%location_id`, `:feed`, and `:saved` as evidenced in `_main` within `instaloader/__main__.py`

I have tested my changes and this pull request seems ready.


aandergr:
Thanks!

PR #1180, opened 2021-06-06
Profile renames are reported even if --quiet is enabled (fixes #1159)

### LINKED ISSUE #1159: Profile renamed message should be displayed even with --quiet
I believe the message 'Profile XXX has changed its name to YYY.' displayed when a profile gets renamed should be displayed even if `--quiet` is enabled.

While it's not really an error, it is something that requires user attention: for the next run, the new name must be used; the old name XXX will not be recognized anymore after the directory has been renamed to YYY.

ekalin:
While this isn't really an error, it requires user action, so the message
is printed to stderr even with --quiet, and repeated at the end when that
option is not enabled.


aandergr:
Thanks!

PR #1183, opened 2021-06-07
Fix and Update has_public_story Functionality

spykard:
If you perform too many queries from a certain IP `has_public_story()` no longer works.

Using this simple example:

``` python
import instaloader

L = instaloader.Instaloader()
L.load_session_from_file('...')

profile = instaloader.Post.from_shortcode(L.context, 'bmw')

print(profile.has_public_story)
print(profile.has_viewable_story)
```

``` bash
HTTP redirect from https://www.instagram.com/graphql/query to https://www.instagram.com/accounts/login/
  File "e:/instaloader/TODO.py", line 9, in <module>
    print(profile.has_public_story)
  File "e:\instaloader\instaloader\structures.py", line 803, in has_public_story
  File "e:\instaloader\instaloader\instaloadercontext.py", line 430, in graphql_query
    session=tmpsession)
  File "e:\instaloader\instaloader\instaloadercontext.py", line 335, in get_json
    raise LoginRequiredException("Redirected to login page. Use --login.")
instaloader.exceptions.LoginRequiredException: Redirected to login page. Use --login.
```

This occurs because the query is invoked anonymously.

``` python
# query not rate limited if invoked anonymously:
with self._context.anonymous_copy() as anonymous_context:
```

Fix: Give users the option to call the query both anonymously and not anonymously with the following functions

* ```has_public_story()```
* ```has_public_story_anon()```

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -591,7 +591,7 @@ class Profile:
         assert 'username' in node
         self._context = context
         self._has_public_story = None  # type: Optional[bool]
-        self._has_public_story_anon = None  # type: Optional[bool]        
+        self._has_public_story_anon = None  # type: Optional[bool]
         self._node = node
         self._has_full_metadata = False
         self._iphone_struct_ = None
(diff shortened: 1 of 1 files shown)


Thammus:
> If you perform too many queries from a certain IP `has_public_story()` no longer works.
[…]
> This occurs because the query is invoked anonymously.

If the query is now rate limited when performed anonymously, I see no advantage in forcing it into an anonymous context anymore.

> Fix: Give users the option to call the query both anonymously and not anonymously with the following functions

The new version of `has_public_story()` which omits the anonymous context does also give users the option to call the query anonymously by simply not logging in. Therefore there is no need to have a separate `has_public_story_anon()`.

So, my suggestion would be to just remove the anonymous context instead of splitting the function into an anonymous and a non-anonymous version.

spykard:
> The new version of `has_public_story()` which omits the anonymous context does also give users the option to call the query anonymously by simply not logging in. Therefore there is no need to have a separate `has_public_story_anon()`.
> 
> So, my suggestion would be to just remove the anonymous context instead of splitting the function into an anonymous and a non-anonymous version.

Correct, now that I think about it the user has the option to call the new version of `has_public_story()` both anonymoysly and while logged in, so it's good to go. I removed `has_public_story_anon()`.

Thammus:
Thanks for reporting and fixing!

Issue #1150, opened 2021-05-25
Login error: JSON decode fail, 403 - Forbidden.


tndgpp:
**Describe the bug**
Today I tried to login from two different accounts, from different IPs, and received a login error, 429 forbidden.

**To Reproduce**
`instaloader --login <username> profile nasa`

**Expected behavior**
I expected to login and start downloading posts.

**Error messages and tracebacks**
Using above command line, I receive this:
`Fatal error: Login error: JSON decode fail, 403 - Forbidden.`

Using module, I get:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/instaloader/instaloadercontext.py", line 224, in login
    resp_json = login.json()
  File "/usr/local/lib/python3.7/dist-packages/requests/models.py", line 900, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/bin/ig-login.py", line 57, in <module>
    L.login(USER, PASSWORD)
  File "/usr/local/lib/python3.7/dist-packages/instaloader/instaloader.py", line 546, in login
    self.context.login(user, passwd)
  File "/usr/local/lib/python3.7/dist-packages/instaloader/instaloadercontext.py", line 228, in login
    ) from err
instaloader.exceptions.ConnectionException: Login error: JSON decode fail, 403 - Forbidden.
```

**Instaloader version**
4.7.1

**Additional context**
If I try to login via web or via app, I don't have any problems. My accounts seem perfectly fine. Also, without logging in, e.g.
`instaloader nasa`
I can download posts, so I guess my IP range is not banned.

santinmike:
I have the exact same problem. I noticed the issue just now.

cschwem2er:
Can confirm that this issue also happens on our end. Any idea what's going on here?

fatpandaclub:
Same issue here.

Javieralonb:
[...]

PR #1184, opened 2021-06-08
Create user session using cookies exported using the EditThisCookie addon (Chrome and others)

Michounet:
This should help while a fix to #1150 is implemented. It's a modified version of 615_import_firefox_session.py that will create a user session file from cookies dumped in JSON format. I personally use [EditThisCookie ](http://www.editthiscookie.com/)on Chrome.

Since I don't use Firefox and I also use instaloader on a Linux headless server  the 615_import_firefox_session.py script didn't fix my login issue reported in #1150 . What I do now is log in to my Instagram account on my usual Chrome browser, export the cookies in JSON format (I use [EditThisCookie](http://www.editthiscookie.com/) but there's many addons for this) and I just have to paste that into a "cookies.txt" file on a remote ssh connection to the Linux server.

### DIFF
--- a/docs/codesnippets/616_import_cookies_from_EditThisCookie_addon.py
+++ b/docs/codesnippets/616_import_cookies_from_EditThisCookie_addon.py
@@ -1,0 +1,59 @@
+# This will automatically import cookies and create a session file
+# Works on all browsers that support EditThisCookie extension (http://www.editthiscookie.com/)
+#
+# Instructions:
+#     1. Install EditThisCookie extension on your supported browser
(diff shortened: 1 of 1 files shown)


aandergr:
I think the added value of this is negligible, since one can just use Firefox to benefit from the published workaround script, rather than install a third-party addon to another browser for exporting cookies.

nzec:
I had made a similar script which doesn't require any extensions: https://github.com/nzec/instaloader_chrome_cookie

(it does require pycookiecheat though)

PR #1193, opened 2021-06-15
Make downloads using --latest-stamps resumable

ekalin:
Using itertools.takewhile() on a NodeIterator returns a plain Iterator,
and so it's not resumable.

The strategy has been altered to pass an extra argument to
posts_download_loop, a lambda that is evaluated for each post, and
causes the loop to stop when it returns false.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -1199,12 +1199,12 @@ class Instaloader:
         .. versionchanged:: 4.8
            Add `latest_stamps` parameter."""
         self.context.log("Retrieving tagged posts for profile {}.".format(profile.username))
-        posts_to_download: Iterator[Post] = profile.get_tagged_posts()
+        posts_takewhile: Optional[Callable[[Post], bool]] = None
         if latest_stamps is not None:
             last_scraped = latest_stamps.get_last_tagged_timestamp(profile.username)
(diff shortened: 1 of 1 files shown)


instaloader/instaloader.py
```diff
@@ -1197,14 +1202,13 @@ def download_tagged(self, profile: Profile, fast_update: bool = False,
...
-            posts_to_download = takewhile(lambda p: p.date_utc.replace(tzinfo=timezone.utc) > last_scraped,
-                                          posts_to_download)
+            posts_takewhile = lambda p: p.date_utc.replace(tzinfo=timezone.utc) > last_scraped
             scraped_timestamp = datetime.now().astimezone()
         self.posts_download_loop(posts_to_download,
                                  target if target
                                  else (Path(_PostPathFormatter.sanitize_path(profile.username)) /
                                        _PostPathFormatter.sanitize_path(':tagged')),
-                                 fast_update, post_filter)
+                                 fast_update, post_filter, takewhile=posts_takewhile)
```
aandergr, line 1211: Here is `posts_takewhile` referenced even though it is undefined when `latest_stamps` is None.

(same in the other altered calls to `posts_download_loop`)

ekalin:
Curious that pylint didn't catch that. Well, I've fixed that and removed one unnecessary variable.

aandergr:
Looks good, thanks! :)

PR #1219, opened 2021-07-03
Fix interruped downloads with --latest-stamps (#1206)

### LINKED ISSUE #1206: --latest-stamps and resuming
There's still one small problem with `--latest-stamps`, when used together with interrupting and resuming a download. And I'm not sure what would be the best way to fix it.

In the first run, if the download is interrupted, then the last timestamp is not stored, because the exception handling skips the code that writes the timestamp. But this is actually a good thing.

In the next run, continuing the interrupted download, the last timestamp is still the same of the first run. As I said, this is good because if the time were saved and used in this second run, nothing would be downloaded.

However, when this second run finishes, it stores as last timestamp the time of this second run, so any media included between both runs would be ignored in the next update.

---

One possibility would be to store in the json file the timestamp, but I'm not sure how could this be done. It would require some huge and intrusive changes.

---

Another idea, that seems more feasible, is to store not the current timestamp in the stamps file, but the timestamp of the newest media downloaded. But this would require having `NodeIterator` cache the first item it returns (or at least its timestamp) so that it can be queried afterwards, and including this timestamp in `FrozenNodeIterator` so that it is stored and loaded in subsequent runs.

---

Please let me know what you think and if there's another way to solve this problem.

ekalin:
Fixes #1206 

The timestamp of the most recent post is cached in NodeIterator (and saved to the disk), and this timestamp is stored, instead of the timestamp instaloader was run.

This way, even in later resuming runs the timestamp stored is the same that would have been stored if the first run.

This is not exactly what has been propsed in the bug discussion, but please take a look and tell me what you think of the solution.

### DIFF
--- a/instaloader/nodeiterator.py
+++ b/instaloader/nodeiterator.py
@@ -129,7 +129,8 @@ class NodeIterator(Iterator[T]):
                 raise
             item = self._node_wrapper(node)
             if self._first_item_timestamp is None and hasattr(item, 'date_utc'):
-                self._first_item_timestamp = item.date_utc.replace(tzinfo=timezone.utc)
+                # Suppress MyPy error, it is checked that T has date_utc in the previous line
+                self._first_item_timestamp = item.date_utc.replace(tzinfo=timezone.utc) # type: ignore
             return item
         if self._data['page_info']['has_next_page']:
             query_response = self._query(self._data['page_info']['end_cursor'])
--- a/instaloader/nodeiterator.py
+++ b/instaloader/nodeiterator.py
@@ -3,7 +3,7 @@ import hashlib
 import json
 import os
 from contextlib import contextmanager
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from lzma import LZMAError
 from typing import Any, Callable, Dict, Iterable, Iterator, NamedTuple, Optional, Tuple, TypeVar
 
@@ -17,7 +17,8 @@ FrozenNodeIterator = NamedTuple('FrozenNodeIterator',
                                  ('context_username', Optional[str]),
                                  ('total_index', int),
                                  ('best_before', Optional[float]),
-                                 ('remaining_data', Optional[Dict])])
+                                 ('remaining_data', Optional[Dict]),
+                                 ('first_item_timestamp', Optional[str])])
 FrozenNodeIterator.query_hash.__doc__ = """The GraphQL ``query_hash`` parameter."""
 FrozenNodeIterator.query_variables.__doc__ = """The GraphQL ``query_variables`` parameter."""
 FrozenNodeIterator.query_referer.__doc__ = """The HTTP referer used for the GraphQL query."""
@@ -26,7 +27,7 @@ FrozenNodeIterator.total_index.__doc__ = """Number of items that have already be
 FrozenNodeIterator.best_before.__doc__ = """Date when parts of the stored nodes might have expired."""
 FrozenNodeIterator.remaining_data.__doc__ = \
     """The already-retrieved, yet-unprocessed ``edges`` and the ``page_info`` at time of freezing."""
-
+FrozenNodeIterator.first_item_timestamp.__doc__ = """Timestamp (UTC) of first item, if an item has been produced."""
 
 T = TypeVar('T')
 
@@ -89,6 +90,7 @@ class NodeIterator(Iterator[T]):
             self._best_before = datetime.now() + NodeIterator._shelf_life
         else:
             self._data = self._query()
+        self._first_item_timestamp: Optional[datetime] = None
 
     def _query(self, after: Optional[str] = None) -> Dict:
         pagination_variables = {'first': NodeIterator._graphql_page_length}  # type: Dict[str, Any]
@@ -125,7 +127,10 @@ class NodeIterator(Iterator[T]):
             except KeyboardInterrupt:
                 self._page_index, self._total_index = page_index, total_index
                 raise
-            return self._node_wrapper(node)
+            item = self._node_wrapper(node)
+            if self._first_item_timestamp is None:
+                self._first_item_timestamp = item.date_utc.replace(tzinfo=timezone.utc)
+            return item
         if self._data['page_info']['has_next_page']:
             query_response = self._query(self._data['page_info']['end_cursor'])
             page_index, data = self._page_index, self._data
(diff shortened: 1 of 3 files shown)


instaloader/nodeiterator.py
```diff
@@ -26,7 +27,7 @@
 FrozenNodeIterator.best_before.__doc__ = """Date when parts of the stored nodes might have expired."""
 FrozenNodeIterator.remaining_data.__doc__ = \
     """The already-retrieved, yet-unprocessed ``edges`` and the ``page_info`` at time of freezing."""
-
+FrozenNodeIterator.first_item_timestamp.__doc__ = """Timestamp (UTC) of first item, if an item has been produced."""
```
aandergr, line 30: Am I right in that this serves the purpose of `start_of_first_incomplete_run` I proposed in https://github.com/instaloader/instaloader/issues/1206#issuecomment-867472513? Good idea to save it here in the resume file rather than in latest_stamps, since it is a property more-inherent to the resuming mechanism, and this way it will continue to work if multiple resume prefixes are used concurrently. Also seems an elegant way of avoiding the `copy` method with all its drawbacks.

However, could we store the whole first item rather than just its `date_utc` field? Then the NodeIterator class would stay truly generic, which would be handy if we'd ever want a similar mechanism for objects that do not have that timestamp.

instaloader/nodeiterator.py
```diff
@@ -157,6 +163,11 @@ def magic(self) -> str:
         ).encode())
         return base64.urlsafe_b64encode(magic_hash.digest()).decode()
 
+    @property
+    def first_item_timestamp(self) -> Optional[datetime]:
+        """If this iterator has produced any items, returns the timestamp of that item, in UTC."""
```
aandergr, line 168: This would probably be an appropriate place for the `.. versionadded:: 4.8` directive.

aandergr (commented):
Thanks! On the first glance it looks like a good fix #1206; and unless I miss something, it is better than the solution I proposed in  https://github.com/instaloader/instaloader/issues/1206#issuecomment-867472513.

However, I think it would be a better abstraction to let NodeIterator store the complete first item rather than only its `date_utc` field, and then access `date_utc` in the caller.

instaloader/nodeiterator.py
ekalin: Yes, that is basically the same as `start_of_first_incomplete_run`.

I remember that I considered saving the whole item, but found some difficulty, but I can't recall exactly what was the problem. Looking again, it seems doable, but instead of `NodeIterator[T]` it would be necessary to limit it to `NodeIterator[JsonExportable]`, since the whole stored object must be stored and loaded as a json string. (Or it could still be any type, but it only gets stored if it happens to be `JsonExportable`.)

aandergr: I think we could store the JSON structure that is returned by Instagram directly, then it is not limited to JsonExportable.

ekalin: The type of object must also be stored, so that when the file contents is read, an object of the correct type can be constructed. And we'd also need a series of `if` clauses to call the proper constructor based on the type. Which is basically what `get_json_structure` and `load_structure` already do. Or do you have another way in mind of saving/restoring the item?

aandergr: We can assume that a FrozenNodeIterator is thawn with a matching NodeIterator, more specifically, one that has been passed the same node_wrapper. Hence, we could assign `self._first_item` the raw `node` object we have in `__next__()`, and let the property `first_item` return `self._node_wrapper(self._first_item)`. I think that should work fine.

aandergr: So, (hopefully) more clearly, we'd have (let's call it `first_node` rather than `first_item`)
- a field `_first_node` in NodeIterator (instead of `_first_item_timestamp`)
- a field `first_node` in FrozenNodeIterator (instead of `first_item_timestamp`), which stores the value of `NodeIterator._first_node`
- assign the raw `node` object to `self._first_node` in `__next__()` at the place where now `self._first_item_timestamp` is assigned
- a property in NodeIterator `first_item()` (I call it first_*item* since what it returns is not a raw node anymore) which returns `self._node_wrapper(self._first_node)` (type `T`)

The callers could then use `iterator.first_item.date_utc` in place of `iterator.first_item_timestamp`.

No need to store the type and still everything stays generic. The node_wrappers are all very simple and efficient constructor functions, which do not do any requests or so.

ekalin: This turned out to be simple and elegant. I've just pushed these changes.

aandergr (commented):
When Instaloader loads a resume file that has been created before introducing the new `first_node` field, it errors
```
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/alex/instaloader/instaloader/__main__.py", line 488, in <module>
    main()
  File "/home/alex/instaloader/instaloader/__main__.py", line 466, in main
    _main(loader,
  File "/home/alex/instaloader/instaloader/__main__.py", line 219, in _main
    instaloader.download_profiles(profiles,
  File "/home/alex/instaloader/instaloader/instaloader.py", line 1428, in download_profiles
    self.posts_download_loop(posts_to_download, profile_name, fast_update, post_filter,
  File "/home/alex/instaloader/instaloader/instaloader.py", line 962, in posts_download_loop
    with resumable_iteration(
  File "/usr/lib/python3.9/contextlib.py", line 117, in __enter__
    return next(self.gen)
  File "/home/alex/instaloader/instaloader/nodeiterator.py", line 270, in resumable_iteration
    fni = load(context, resume_file_path)
  File "/home/alex/instaloader/instaloader/structures.py", line 1667, in load_structure_from_file
    return load_structure(context, json_structure)
  File "/home/alex/instaloader/instaloader/structures.py", line 1646, in load_structure
    return FrozenNodeIterator(**json_structure['node'])
TypeError: <lambda>() missing 1 required positional argument: 'first_node'
```
We'd need to change `load_structure()` to set that field to `None` if it is not present.

aandergr:
Looks good and seems to work fine. Thanks a lot for your effort with contributing this nice feature! I released 4.8rc4 with this fix included, and I plan to release 4.8 final in early August if all goes well :)

PR #1223, opened 2021-07-07
Add Proxy support

zabalkansky:
This change adds the ability to specify SOCKS or other proxies to be used by Instaloader when making HTTP requests. 

Here is an example of creating Instaloader object with SOCKS5 proxy:

```
pdef = 'socks5h://%s:%d' % (...)
proxies = { 'http': pdef, 'https': pdef }
L = Instaloader(proxies=proxies, user_agent='Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:81.0) Gecko/20100101 Firefox/81.0')
```

Currently the changes affect only Instaloader and InstaloaderContext objects, not the CLI. The appropriate options should be added.
The documentation also was not updated.

### DIFF
--- a/instaloader.py
+++ b/instaloader.py
@@ -1,0 +1,6 @@
+#!/usr/bin/env python3
+
+from instaloader.__main__ import main
+
+if __name__ == '__main__':
+    main()
--- a/instaloader.py
+++ b/instaloader.py
@@ -1,6 +1,0 @@
-#!/usr/bin/env python3
-
-from instaloader.__main__ import main
-
-if __name__ == '__main__':
-    main()
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -215,11 +215,12 @@ class Instaloader:
                  slide: Optional[str] = None,
                  fatal_status_codes: Optional[List[int]] = None,
                  iphone_support: bool = True,
-                 title_pattern: Optional[str] = None):
+                 title_pattern: Optional[str] = None,
+                 proxies: Optional[dict] = None):
 
         self.context = InstaloaderContext(sleep, quiet, user_agent, max_connection_attempts,
                                           request_timeout, rate_controller, fatal_status_codes,
-                                          iphone_support)
+                                          iphone_support, proxies)
 
         # configuration parameters
         self.dirname_pattern = dirname_pattern or "{target}"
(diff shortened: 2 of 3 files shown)


Thammus (commented):
Thanks a lot for this PR. Proxy support is a feature that has been suggested multiple times by many users. We already had some Pull Requests addressing this feature, e.g. #174 and #1003. Having a proxies attribute may help those for whom requests' proxy support through environment variables is not sufficient.

Looking at this code,
```python3
if self._proxies is not None:
    session.proxies = self._proxies
```
would it not be sufficient to always set `session.proxies = self._proxies` and omit the if-clause?

When setting the proxy inside `get_anonymous_session()`, `load_session_from_file()` and `login()`, as far as I can tell, all relevant requests inside `InstaloaderContext` that are issued got enhanced with the proxy setting. Does this not make it redundant to additionally specify the proxy as an argument when using `session.get()` and `session.post()`?

The target branch for new features should be `upcoming/v4.9` instead of `master`.

zabalkansky:
>would it not be sufficient to always set session.proxies = self._proxies and omit the if-clause?
Yes, thanks for noticing this useless statement. It needs to be removed.

> When setting the proxy inside get_anonymous_session(), load_session_from_file() and login(), as far as I can tell, all relevant requests inside InstaloaderContext that are issued got enhanced with the proxy setting. Does this not make it redundant to additionally specify the proxy as an argument when using session.get() and session.post()?

I tried omitting this argument but the requests were still sent directly despite that the session was initialized with proxy. I don't know, maybe it works with the current version of requests library but this argument was necessary at the time when I was writing this patch.

zabalkansky:
> would it not be sufficient to always set session.proxies = self._proxies and omit the if-clause?

I looked at the code again and now I think this if clause is needed because otherwise proxies may not be picked from the environment variables by requests library. So I would leave the code as it is.

Thammus:
> I tried omitting this argument but the requests were still sent directly despite that the session was initialized with proxy.

If altering `session.proxies` does not have any effect, why did you introduce code that does so? But according to the python-requests documentation `session.proxies.update(proxies)` does configure proxies for the entire session. 
Using the `proxies` argument with request methods like `get()` and `post()` is described as an alternative way of configuring proxies which can be useful if only individual requests need to be modified. Having both methods in our code is redundant and unpleasant. 

> I looked at the code again and now I think this if clause is needed because otherwise proxies may not be picked from the environment variables by requests library. So I would leave the code as it is.

Ah, you are right. The documentation indeed states that Requests only defaults to the proxy configuration defined by standard environment variables if none of the above mentioned ways of altering the configuration is used.

Thammus:
@zabalkansky Are you still interested to work on this PR?

PR #1232, opened 2021-07-21
Pick best video (#1022)

### LINKED ISSUE #1022: App version video quality issue
In some previous PR and commits (#712, #630, etc.), we introduced the feature that we choose to download the media served to app client (instead of the ones served for web), since they typically have higher resolution.

This still remains true, but the app version tends to have slightly lower quality. This was noted before, but the difference was small. Only recently, I noticed that the quality difference sometimes is very huge, especially for videos. 

I started to think, should we really download the app version despite having higher nominal resolution?

Let's take this example: https://www.instagram.com/p/CLHoAQpCI2i/
* The web version is 640x480 875KB.
* The app version is 720x540, but only 240KB.

And if you compare the two side by side, it is clear that the app version is *awful*:

Web:
image
App:
image

(Just to make it clear: it's not always the case. Sometimes the app version is still better; sometimes the two videos are identical. So the whole situation is more complicated and not black and white)

As the one who suggested this into instaloader to begin with, I feel responsible to improve this. Like adding an option, or do it more smartly. But I would like to hear your opinion first.

Thanks!

fireattack:
This fixes #1022.

Also, Instaloader supports to download single post which works with Stories. But the returned data is slightly different and can't be properly handled by existing `Post` class. I patched this part as well. 

They are called `typename` = `StoryImage` and `StoryVideo` (NOT `GraphStoryImage` or `GraphStoryVideo` returned from the Stories timeline as `StoryItem`.)

Weirdly, the `StoryVideo` returned by single post API doesn't behave like `GraphStoryVideo` returned by Stories timeline either. It has no `self._node['video_url']` or `self._node['video_resources']`. So you can only get the video from `self._iphone_struct['video_versions']`.

In long term, I think it's better we either create a base class `Item` which both `StoryItem` and `Post` can inherit from; or just merge the two. This can dramatically reduce the redundancy of code in `structures.py`.

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -528,6 +528,25 @@ class InstaloaderContext:
         :raises ConnectionException: When download repeatedly failed."""
         self.write_raw(self.get_raw(url), filename)
 
+    def head(self, url: str, allow_redirects: Optional[bool] = None) -> requests.Response:
+        """HEAD a URL anonymously.
+
+        :raises QueryReturnedNotFoundException: When the server responds with a 404.
+        :raises QueryReturnedForbiddenException: When the server responds with a 403.
+        :raises ConnectionException: When request failed."""
+        with self.get_anonymous_session() as anonymous_session:
+            resp = anonymous_session.head(url, allow_redirects=allow_redirects)
+        if resp.status_code == 200:
+            return resp
+        else:
+            if resp.status_code == 403:
+                # suspected invalid URL signature
+                raise QueryReturnedForbiddenException("403 when accessing {}.".format(url))
+            if resp.status_code == 404:
+                # 404 not worth retrying.
+                raise QueryReturnedNotFoundException("404 when accessing {}.".format(url))
+            raise ConnectionException("HTTP error code {}.".format(resp.status_code))
+
     @property
     def root_rhx_gis(self) -> Optional[str]:
         """rhx_gis string returned in the / query."""
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -533,7 +533,10 @@ class InstaloaderContext:
 
         :raises QueryReturnedNotFoundException: When the server responds with a 404.
         :raises QueryReturnedForbiddenException: When the server responds with a 403.
-        :raises ConnectionException: When request failed."""
+        :raises ConnectionException: When request failed.
+        
+        .. versionadded:: 4.7.6
+        """
         with self.get_anonymous_session() as anonymous_session:
             resp = anonymous_session.head(url, allow_redirects=allow_redirects)
         if resp.status_code == 200:
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -534,7 +534,7 @@ class InstaloaderContext:
         :raises QueryReturnedNotFoundException: When the server responds with a 404.
         :raises QueryReturnedForbiddenException: When the server responds with a 403.
         :raises ConnectionException: When request failed.
-        
+
         .. versionadded:: 4.7.6
         """
         with self.get_anonymous_session() as anonymous_session:
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -528,7 +528,7 @@ class InstaloaderContext:
         :raises ConnectionException: When download repeatedly failed."""
         self.write_raw(self.get_raw(url), filename)
 
-    def head(self, url: str, allow_redirects: Optional[bool] = None) -> requests.Response:
+    def head(self, url: str, allow_redirects: bool = False) -> requests.Response:
         """HEAD a URL anonymously.
 
         :raises QueryReturnedNotFoundException: When the server responds with a 404.
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -6,6 +6,8 @@ from collections import namedtuple
 from datetime import datetime
 from typing import Any, Dict, Iterable, Iterator, List, Optional, Union
 
+import requests
+
 from . import __version__
 from .exceptions import *
 from .instaloadercontext import InstaloaderContext
@@ -373,14 +375,20 @@ class Post:
     @property
     def video_url(self) -> Optional[str]:
         """URL of the video, or None."""
+        def get_size(url):
+            return requests.head(url).headers.get('Content-Length', 0), url
         if self.is_video:
+            url_candidates = []
+            if 'video_url' in self._node:
+                url_candidates.append(get_size(self._node['video_url']))
             if self._context.is_logged_in:
                 try:
-                    url = self._iphone_struct['video_versions'][0]['url']
-                    return url
+                    self._iphone_struct['video_versions'].sort(key=lambda x: x['height'])
+                    url_candidates.append(get_size(self._iphone_struct['video_versions'][-1]['url']))
                 except (InstaloaderException, KeyError, IndexError) as err:
                     self._context.error('{} Unable to fetch high quality video version of {}.'.format(err, self))
-            return self._field('video_url')
+            url_candidates.sort()
+            return url_candidates[-1][1]
         return None
 
     @property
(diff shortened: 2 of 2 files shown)


aandergr:
> I patched this part as well.

This part feels too risky to me to patch it directly into the master branch and release it without a pre-release cycle. Would it be okay for you to create a separate PR for that, ideally based on the `upcoming/v4.8` branch? Generally, I agree there is a chance that it might be better to somehow merge or abstract `Post` and `StoryItem`.

fireattack:
Split that part to #1238.

instaloader/structures.py
```diff
@@ -373,14 +375,20 @@ def is_video(self) -> bool:
...
         if self.is_video:
+            url_candidates = []
+            if 'video_url' in self._node:
+                url_candidates.append(get_size(self._node['video_url']))
             if self._context.is_logged_in:
                 try:
-                    url = self._iphone_struct['video_versions'][0]['url']
-                    return url
+                    self._iphone_struct['video_versions'].sort(key=lambda x: x['height'])
+                    url_candidates.append(get_size(self._iphone_struct['video_versions'][-1]['url']))
```
aandergr, line 387: To me it seems we can just compare them all and choose the one with the biggest filesize, i.e. the highest bitrate:

```suggestion
                    for version in self._iphone_struct['video_versions']:
                        url_candidates.append(get_size(version['url']))
```

instaloader/structures.py
```diff
@@ -1102,8 +1110,19 @@ def is_video(self) -> bool:
...
+        def get_size(url):
+            return requests.head(url).headers.get('Content-Length', 0), url
         if self.is_video:
-            return self._node['video_resources'][-1]['src']
+            url_candidates = []
+            url_candidates.append(get_size(self._node['video_resources'][-1]['src']))
+            if self._context.is_logged_in:
+                try:
+                    self._iphone_struct['video_versions'].sort(key=lambda x: x['height'])
+                    url_candidates.append(get_size(self._iphone_struct['video_versions'][-1]['url']))
```
aandergr, line 1121: (same as above)

fireattack:
Applied suggested changes (manually, since GitHub's web built-in "apply" feasure does not allow you to modify the suggested changes in anyway).

instaloader/structures.py
```diff
@@ -373,14 +375,20 @@ def is_video(self) -> bool:
...
         if self.is_video:
+            url_candidates = []
+            if 'video_url' in self._node:
+                url_candidates.append(get_size(self._node['video_url']))
             if self._context.is_logged_in:
                 try:
-                    url = self._iphone_struct['video_versions'][0]['url']
-                    return url
+                    for version in self._iphone_struct['video_versions']:
+                            url_candidates.append(get_size(version['url']))
```
aandergr, line 387:
```suggestion
                        url_candidates.append(get_size(version['url']))
```

To fix Pylint's `bad-indentation` warning.

Other than that, it looks good I think.

instaloader/structures.py
```diff
@@ -373,14 +375,20 @@ def is_video(self) -> bool:
     @property
     def video_url(self) -> Optional[str]:
         """URL of the video, or None."""
+        def get_size(url):
+            return requests.head(url).headers.get('Content-Length', 0), url
```
aandergr, line 379: Here we call requests directly and so bypass the HTTP headers and error handling that we usually have when doing a request. I think it would be better to rather have a method in `InstaloaderContext` to do that HEAD request with all the headers set, and ideally wrap HTTP error codes to exceptions. I think of something similar to `InstaloaderContext.get_raw(url: str) -> requests.Response`, with the difference that it does a HEAD request instead of a GET request, and that we use that new method instead of directly calling `requests.head()`. This would also allow to handle the case that a URL candidate yields a non-ok HTTP response.

instaloader/structures.py
```diff
@@ -1102,8 +1110,19 @@ def is_video(self) -> bool:
...
```
aandergr, line 1114: (same as above)

aandergr (commented):
I have just noticed another issue with this PR: It lacks the typical setup of HTTP headers and error handling since it calls the raw `requests.head()` function to do the requests to determine the candidate's file sizes. Sorry for not having noticed that in my last review.

fireattack:
Fixed.

I also exposed `allow_redirects` for `requests.head()` because if not using `allow_redirects=True`, it cannot get real content-length for 302 requests.

302 (basically it redirects you from specific CDNs like `scontent-dfw5-2.cdninstagram.com` to generic domain `scontent.cdninstagram.com`) shouldn't happen anymore after #1237, but I still would like to be able to handle it properly anyway.

instaloader/instaloadercontext.py
```diff
@@ -528,6 +528,25 @@ def get_and_write_raw(self, url: str, filename: str) -> None:
         :raises ConnectionException: When download repeatedly failed."""
         self.write_raw(self.get_raw(url), filename)
 
+    def head(self, url: str, allow_redirects: Optional[bool] = None) -> requests.Response:
```
aandergr, line 531:
```suggestion
    def head(self, url: str, allow_redirects: bool = False) -> requests.Response:
```

Otherwise, we would pass `allow_redirects=None` to `anonymous_session.head()` if no `allow_redirects` parameter is specified, which looks wrong to me.

instaloader/instaloadercontext.py
```diff
@@ -528,6 +528,25 @@ def get_and_write_raw(self, url: str, filename: str) -> None:
...
+        """HEAD a URL anonymously.
+
+        :raises QueryReturnedNotFoundException: When the server responds with a 404.
+        :raises QueryReturnedForbiddenException: When the server responds with a 403.
+        :raises ConnectionException: When request failed."""
```
aandergr, line 536:
```suggestion
        :raises ConnectionException: When request failed.
        
        .. versionadded:: 4.7.6
        """
```

(add *versionadded* directive)

instaloader/structures.py
```diff
@@ -373,14 +373,20 @@ def is_video(self) -> bool:
...
         if self.is_video:
+            url_candidates = []
+            if 'video_url' in self._node:
+                url_candidates.append(get_size(self._node['video_url']))
             if self._context.is_logged_in:
                 try:
-                    url = self._iphone_struct['video_versions'][0]['url']
-                    return url
+                    for version in self._iphone_struct['video_versions']:
+                        url_candidates.append(get_size(version['url']))
```
aandergr, line 385: This would abort the whole candidate selection process if only one of the HEAD requests is answered with a non-ok HTTP response. Maybe it would be better to catch the exceptions inside the loop, to only skip the erroring candidates, and raise an exception only if all candidates yield an error?

instaloader/structures.py
```diff
@@ -1102,8 +1108,19 @@ def is_video(self) -> bool:
...
+        def get_size(url):
+            return self._context.head(url, allow_redirects=True).headers.get('Content-Length', 0), url
         if self.is_video:
-            return self._node['video_resources'][-1]['src']
+            url_candidates = []
+            url_candidates.append(get_size(self._node['video_resources'][-1]['src']))
+            if self._context.is_logged_in:
+                try:
+                    for version in self._iphone_struct['video_versions']:
+                        url_candidates.append(get_size(version['url']))
```
aandergr, line 1119: (same as above)

aandergr (commented):
Thanks for the improvement!

> I also exposed `allow_redirects` for `requests.head()` because if not using `allow_redirects=True`, it cannot get real content-length for 302 requests.

Ah that's a nice finding, thanks!

instaloader/instaloadercontext.py
fireattack: Thanks! I was confused by the requests documentation ad the time, and thought there was a difference between None and False. Turns out it just has different default value (`False`) for `.head()`.

fireattack:
Fixed the error handling in loop thing. I don't know a more elegant way to do it, unfortunately.

aandergr:
Alright, I just pushed my suggestion for the `video_url` property implementation. It now goes through all candidates, prints an error message for every failing candidate, and should requesting the size of all URL candidates fail, it returns the 'default' web version URL. I also added an optimization to not check duplicate URLs twice.

fireattack:
Cool, thanks!

Issue #1145, opened 2021-05-21
Video download slow since 19th May


LaticePrime:
Has anyone else found downloading profile posts unusably slow for the last two days? Scraping stories still just fine!

skrull:
Yes, same here.

LaticePrime:
Thanks! Good to know it's not just me!

aandergr:
Interestingly, just 4 days before this issue we received a bug report ["Instaloader works too fast"](https://github.com/instaloader/instaloader/issues/1141). 

Since July 2020, Instaloader's rate controller can be easily customized without having to touch Instaloader's code. So luckily, if you find Instaloader downloading too slow, too fast or that its perceived download speed varies too much, you can easily adapt it to match your expectations: https://instaloader.github.io/module/instaloadercontext.html#ratecontroller

ekalin:
Same here.

It does not seem to have anything to do with the rate controller. The actual download of media is really slow. I can see the partially downloaded '.temp' files growing very slowly. It seems Instagram has added to throttle on their side.

skrull:
It's something server side. I'm using 4.7.1 and this behavior started a few days ago with video files. It takes around 4 minutes to download each video here.

LaticePrime:
This is exactly what's happened to me; when I view the .temp file it's now growing much more slowly than normal for profile videos only; snatching stories is still really fast. It was fine on Wednesday the 19th so started on the 20th. Has to be instagram side as I've changed nothing and running latest version.

Lex-2008:
Just started using instaloader today, to learn that I was 4 days late :) Have anyone found a workaround? Does being logged in or using a different user-agent help?

Lex-2008:
Note that with decreased speed instaloader _might_ fail to download a complete video file, leaving you with a partially-downloaded one!

It looks like issue is indeed in user-agent string. Probably the one used in instaloader got banned by Instagram. A working **WORKAROUND** is this:
1. visit echo.opera.com with a browser to see your user-agent
2. add `--user-agent` argument to your instaloader command

Alternatively, you can use my user-agent (Microsoft Edge Beta on Linux):
```
--user-agent 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.57 Safari/537.36 Edg/91.0.864.27'
```

skrull:
Uh, much easier than I thought. Thanks.

[...]

PR #1237, opened 2021-07-23
Do not delete referer header

### LINKED ISSUE #1233: Download speed is extremely slow when the default UA is applied and no referer
**Describe the bug**

When downloading the media, the download speed is very slow, as low as 50KB/s.

**To Reproduce**
```sh
instaloader -- -CRjFJnKJVRe
```

**Expected behavior**
The download speed should be relatively fast.

**Instaloader version**

`master`

**Additional context**
After some debugging, I found this problem only happens when the default UA `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.90 Safari/537.36` is used and no referer of `https://www.instagram.com/` is applied.

Removing either the UA entirely or add IG as referer, the download speed instantly becomes 10MB/s (my bandwidth limit).

One can easily test it by adding 

```py
            anonymous_session.headers['user-agent'] = None
```

after L509 at `instaloadercontext.py`:

https://github.com/instaloader/instaloader/blob/8c1129e5563490b3cfe37f8d172c80907452da07/instaloader/instaloadercontext.py#L509-L511

fireattack:
since it causes IG's CDN to limit download speed

Fixes #1233 and #1145.

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -152,7 +152,6 @@ class InstaloaderContext:
         if empty_session_only:
             del header['Host']
             del header['Origin']
-            del header['Referer']
             del header['X-Instagram-AJAX']
             del header['X-Requested-With']
         return header


aandergr:
Thanks! I released this fix in [Version 4.7.5](https://github.com/instaloader/instaloader/releases/tag/v4.7.5).

Issue #1080, opened 2021-04-06
Hashtag.from_name(): KeyError: 'graphql'


Exyss:
Found an error given by using wrong json metadata. 
Tried using different hashtags and to revert to previous versions of Instaloader, but it still gives the problem.
I also checked your built-in code and everything looks fine to me, so I suppose it has something to do with Instagram changing it's metadata format.
Maybe this was already fixed for other types of objects but you forgot to fix it for Hashtags (?)

```
Traceback (most recent call last):
  File "C:\Users\exyss\Desktop\v3.0-beta\src\test.py", line 6, in <module>
    hashtag = instaloader.Hashtag.from_name(_loader.context, "cats")
  File "C:\Users\exyss\AppData\Local\Programs\Python\Python39\lib\site-packages\instaloader\structures.py", line 1302, in from_name
    hashtag._obtain_metadata()
  File "C:\Users\exyss\AppData\Local\Programs\Python\Python39\lib\site-packages\instaloader\structures.py", line 1316, in _obtain_metadata        
    self._node = self._query({"__a": 1})
  File "C:\Users\exyss\AppData\Local\Programs\Python\Python39\lib\site-packages\instaloader\structures.py", line 1311, in _query
    return self._context.get_json("explore/tags/{0}/".format(self.name),
KeyError: 'graphql'
```

MarcoChilese:
I also have this error, an update would be nice

santinmike:
I am using Instaloader 4.7.1. I have the same error. It seems to be a bug due to some recent changes made by Instagram.

aandergr:
Thanks for reporting. Indeed, the structure of the data returned by Instagram for the query that is used there has changed recently.

Luckily though, the alternative described in https://github.com/instaloader/instaloader/issues/874#issuecomment-731561430 still works and can be used instead until this is fixed.

martacosano:
Hello! 
Does the error depends on the account of instagram that I use to download the images?

I try to download images from a hashtag, when I use my personal account of instagram to do it I don’t have that problem. But when I use a new account that I created just for this I have the same problem.
Thank you,

hobzcalvin:
Hi folks, any updates? I have an art project that is dead in the water without hashtag support. I can try digging into the code myself if someone provides guidance on what's happening.

TomaszWychocki:
@hobzcalvin you can temporarily use my code. Just go through JSON and fetch all data you need (line 34). 

```
import instaloader
import settings
[...]

Issue #1129, opened 2021-05-07
Instaloader.get_location_posts(): TypeError: 'NoneType' object is not subscriptable


queensoft666:
**Describe the bug**
ERROR: Failed to execute script __main__

**To Reproduce**
Command line:
`instaloader.exe %1035567819 --login XXXXXXXXXXX -p XXXXXXXXXXX --no-compress-json --no-video-thumbnails`
Account is OK, tested in browser.

**Expected behavior**
Download images for that specific location:
https://www.instagram.com/explore/locations/1035567819/talpa-bascoveni-teleorman/

**Error messages and tracebacks**
```
Loaded session from C:\Users\Queen Soft\AppData\Local\Instaloader\session-XXXXXXXXXXXXXXXXX.
Logged in as XXXXXXXXXXXXXXXXXXX.
Retrieving pictures for location 1035567819...
Traceback (most recent call last):
  File "__main__.py", line 478, in <module>
  File "__main__.py", line 475, in __main
  File "__main__.py", line 445, in main
  File "__main__.py", line 163, in _main
  File "instaloader\instaloader.py", line 64, in call
  File "instaloader\instaloader.py", line 1044, in download_location
  File "instaloader\instaloader.py", line 894, in posts_download_loop
  File "instaloader\instaloader.py", line 1014, in get_location_posts
TypeError: 'NoneType' object is not subscriptable
[8584] Failed to execute script __main__
```

**Instaloader version**
4.7.1

**Additional context**
Windows 10 x64 2004 19041.928
I have full admin rights.

Older version of instaloader (4.6.2) worked a few days ago.
Older version (4.6.2) is also not working right now.

Delete session file - login is OK - but same error afterwards.
Session file is NOT created, but login says OK every time I run the command.
```
Session file does not exist yet - Logging in.
Logged in as XXXXXXXXXXXXXX.
```

Thank you.

aandergr:
Indeed, this does not seem to work anymore. Thanks for this exemplary bug report!



> Session file is NOT created, but login says OK every time I run the command.

Session file is created at the end of Instaloader's execution (unless it terminates due to an error). However, you can use `instaloader --login LOGIN [--password PASSWORD] [--sessionfile SESSIONFILE]` (without any further arguments) to only login and create a sessionfile.

queensoft666:
Yes, I noticed session file is created at the end of the run (profile and other types of download are working fine).

queensoft666:
[...]

PR #1238, opened 2021-07-23
Handle StoryImage & StoryVideo in `download_post`

fireattack:
Instaloader supports to download single post which works with Stories. But the returned data is slightly different and can't be properly handled by existing Post class. I patched this part as well.

They are called typename = StoryImage and StoryVideo (NOT GraphStoryImage or GraphStoryVideo returned from the Stories timeline as StoryItem.)

Also note: weirdly, the StoryVideo returned by single post API doesn't behave like GraphStoryVideo returned by Stories timeline either. It has no self._node['video_url'] or self._node['video_resources']. So you can only get the video from self._iphone_struct['video_versions'].

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -695,12 +695,12 @@ class Instaloader:
                                                             mtime=post.date_local, filename_suffix=suffix)
                 else:
                     downloaded = False
-        elif post.typename == 'GraphImage':
+        elif post.typename in ['GraphImage', 'StoryImage']:
             # Download picture
             if self.download_pictures:
                 downloaded = (not _already_downloaded(filename + ".jpg") and
                               self.download_pic(filename=filename, url=post.url, mtime=post.date_local))
-        elif post.typename == 'GraphVideo':
+        elif post.typename in ['GraphVideo', 'StoryVideo']:
             # Download video thumbnail (--no-pictures implies --no-video-thumbnails)
             if self.download_pictures and self.download_video_thumbnails:
                 with self.context.error_catcher("Video thumbnail of {}".format(post)):
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -105,7 +105,7 @@ class Post:
     @staticmethod
     def supported_graphql_types() -> List[str]:
         """The values of __typename fields that the :class:`Post` class can handle."""
-        return ["GraphImage", "GraphVideo", "GraphSidecar"]
+        return ["GraphImage", "GraphVideo", "GraphSidecar", "StoryImage", "StoryVideo"]
 
     def _asdict(self):
         node = self._node


aandergr:
That's a nice finding - cool to see that it just works!

`Post` seems to already well handle nodes with type `StoryImage`. In the case of a `StoryVideo`, it fails when accessing the properties `video_view_count` and `video_duration`. I think it would be okay to just let it return `None` in this case. Also, `Post.supported_graphql_types()` should then also list `StoryImage` and `StoryVideo` :)

I also tried to put the node returned by that query into `StoryItem`, but that class could not access the fields video_url or expiring_at_timestamp. Seems that `Post` is a better fit.

It could be that StoryItem is actually redundant and all its functionality would be covered by Post with only a few changes.

aandergr:
One problem I noticed here is that the properties `video_view_count` and `video_duration` fail when the Post had been initialized with a StoryVideo, so actually we do not have full compatibility yet.

Also, but this is more a long-term issue: I am not yet sure what a good strategy might be here, since the Post and StoryItem structures may require some refactoring in the near future. I don't like the redundancy that we have now, and also, at least Post needs some work to properly support the new 'sections' format outlined in #1240, needed to fix #1080 and #1129.

fireattack:
Good to know! I will leave this one open for now as we think about how to refactor the structure for the long term.

fireattack:
@aandergr:

Despite the fact this is just a quick workaround instead of full proper implementation/refactoring around `StoryImage` can we merge this in the meantime? 

Currently, downloading a single story with static image (`StoryImage`) straight up does not work (as no image would be downloaded):

```py
import instaloader

my_username = 'yourusername'

L = instaloader.Instaloader()
L.load_session_from_file(my_username)

media_id = 2793728037449719617 # replace with any story that has a static image (no special effect etc.)
shortcode = instaloader.Post.mediaid_to_shortcode(media_id)
```

```
Loaded session from C:\Users\****\AppData\Local\Instaloader\session-****.
Warning: <Post CbFUNknPV9B> has unknown typename: StoryImage
[そういえば新しい トイカメラ買いました 楽しい！] json
```

(It only downloads .json.xz and .txt, not the image.)

aandergr:
I think in its current form (i.e. Post and StoryItem being separate classes), Post is not supposed to handle StoryItems, even though the class seems to work mostly fine with them. I would rather favor to have a method `from_mediaid()` in StoryItem that fetches the StoryItem with the given mediaid, as e.g.

```python
    @classmethod
    def from_mediaid(cls, context: InstaloaderContext, mediaid: int):
        """Create a StoryItem object from a given mediaid"""
        pic_json = context.graphql_query(
            '2b0673e0dc4580674a88d426fe00ea90',
            {'shortcode': Post.mediaid_to_shortcode(mediaid)}
        )
        shortcode_media = pic_json['data']['shortcode_media']
        if shortcode_media is None:
            raise BadResponseException("Fetching StoryItem metadata failed.")
        return cls(context, shortcode_media)
```

fireattack:
Thanks!

Issue #1109, opened 2021-04-22
Post.location: TypeError: 'NoneType' object is not subscriptable


tom-s-ukgwa:
I'd been running Instaloader for a few weeks and it was working really well for me. Great tool!

Then, a few days ago, when I ran the following:

`instaloader profile envagency --geotags --stories --highlights --igtv --login` (with login details)

I got the following error messages:

```
File "/home/work/anaconda3/bin/instaloader", line 8, in <module>
  sys.exit(main())
File "/home/work/anaconda3/lib/python3.8/site-packages/instaloader/__main__.py", line 445, in main
  _main(loader,
File "/home/work/anaconda3/lib/python3.8/site-packages/instaloader/__main__.py", line 211, in _main
  instaloader.download_profiles(profiles,
File "/home/work/anaconda3/lib/python3.8/site-packages/instaloader/instaloader.py", line 1289, in download_profiles
  self.posts_download_loop(profile.get_posts(), profile_name, fast_update, post_filter,
File "/home/work/anaconda3/lib/python3.8/site-packages/instaloader/instaloader.py", line 922, in posts_download_loop
  downloaded = self.download_post(post, target=target)
File "/home/work/anaconda3/lib/python3.8/site-packages/instaloader/instaloader.py", line 670, in download_post
  if self.download_geotags and post.location:
File "/home/work/anaconda3/lib/python3.8/site-packages/instaloader/structures.py", line 554, in location
  loc = self._context.get_json("explore/locations/{0}/".format(location_id),
TypeError: 'NoneType' object is not subscriptable
```

Instaloader downloads stories fine, and the first post after them but displays the above error when attempting to download the second post.

I'm running the latest version (4.7.1)

Any idea how I might fix this? 

Thank you

aandergr:
Thanks for reporting!

> `instaloader profile envagency --geotags --stories --highlights --igtv --login` (with login details)

Also reproducible by downloading the failing post directly: `instaloader --login=LOGIN --geotags -- -CNXwMo1C24i`

faaip:
I have still have this problem using version 4.8.1 with the #1244 fix by @aandergr . Should it be working now?

```
Retrieving pictures for location 1425361787550852...
Traceback (most recent call last):
  File "/Users/frederik/.pyenv/versions/instaloader_cli_wrapper/bin/instaloader", line 8, in <module>
    sys.exit(main())
[...]

PR #1240, opened 2021-07-26
[Partial fix] Fixed the issue for get_location_posts

otanim:
@aandergr, please check this PR,

it fixes the following issue:
https://github.com/instaloader/instaloader/issues/1129https://github.com/instaloader/instaloader/issues/1129

However the fix has a side effect, the API cannot retrieve the page of data via `max_id` that is being retrieved from the payload's `next_max_id`. Basically, it keeps downloading the same page over and over as if it ignores the `max_id` parameter.

Please do not merge this PR until it's not completely fixed, feel free to push new commits to this branch if you figured out the solution.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -1091,11 +1091,36 @@ class Instaloader:
                 params = {'__a': 1, 'max_id': end_cursor}
             else:
                 params = {'__a': 1}
-            location_data = self.context.get_json('explore/locations/{0}/'.format(location),
-                                                  params)['graphql']['location']['edge_location_to_media']
-            yield from (Post(self.context, edge['node']) for edge in location_data['edges'])
-            has_next_page = location_data['page_info']['has_next_page']
-            end_cursor = location_data['page_info']['end_cursor']
+
+            json_data = self.context.get_json('explore/locations/{0}/'.format(location),
+                                              params)
+            location_data = json_data['native_location_data']['recent'] if 'native_location_data' in json_data else \
+                json_data['entry_data']['LocationsPage'][0]['native_location_data']['recent']
+
+            medias = []
+            sections = location_data['sections']
+            for section in sections:
+                medias_per_layout_content = section['layout_content']['medias']
+                for media_per_layout_content in medias_per_layout_content:
+                    media = media_per_layout_content['media']
+                    medias += [media]
+
+            def graphql_patch(data):
+                data_modified = {'code': data['code']}
+                if "image_versions2" in data:
+                    data_modified['__typename'] = "GraphImage"
+                elif "carousel_media" in data:
+                    data_modified['__typename'] = "GraphSidecar"
+                data_modified['is_video'] = "video_versions" in data
+                data_modified['taken_at_timestamp'] = data['taken_at']
+                data_modified['iphone_struct'] = data
+                return data_modified
+
+            medias_graphql_patched = list(map(graphql_patch, medias))
+
+            yield from (Post(self.context, media) for media in medias_graphql_patched)
+            has_next_page = location_data['more_available']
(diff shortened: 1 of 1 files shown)


aandergr:
Hi! Thanks a lot for your effort so far with this PR!

I have noticed that it contains a lot of unrelated changes, such as whitespace changes in otherwise untouched parts of the code and that it does not pass the automatic status checks. We are currently managing a large number of PRs in parallel, and accordingly it would be very helpful if you could solve these superficial problems first before we take a closer look. Also, I am usually hard to convince to add new dependencies.

otanim:
Hey @aandergr, sure thing,
- regarding unrelated changes (whitespaces, etc), those changes happened during code refactorings, I've reverted those back,
- regarding the newly added dependency, I fully understand your concern, I've removed it (it was [pydash](https://github.com/dgilland/pydash), it was a library for simplification of actions related to objects/arrays.
- regarding not addressing the issue "TypeError: 'NoneType' object is not subscriptable", it actually does, the issue here was that the payload that is being received from Instagram, has 2 fields:
   - `graphql` which is a field of type "None",
   - and another field called `native_location_data`, which is an alternated version of the expected and outdated `graphql` field.
   
I hope after the clean-up it's much better/valid for the review, please double-check my newest changes.

otanim:
Hey @aandergr, previously I was wrong regarding my fix for `--geotags`, I've updated the PR with the fix.

instaloader/structures.py
```diff
@@ -553,10 +553,14 @@ def location(self) -> Optional[PostLocation]:
...
-        self._location = PostLocation(location_id, loc['name'], loc['slug'], loc['has_public_page'],
-                                      loc['lat'], loc['lng'])
+            loc_post_requested = self._context.get_json("explore/locations/{0}/".format(location_id),
+                                         params={'__a': 1})['native_location_data']['location_info']
+        loc_name = loc.get("name") or loc_post_requested.get("name")
+        loc_slug = loc.get("slug") or loc_post_requested.get("slug")
+        loc_has_public_page = loc.get("has_public_page") or loc_post_requested.get("has_public_page")
+        loc_lat = loc.get("lat") or loc_post_requested.get("lat")
+        loc_lng = loc.get("lng") or loc_post_requested.get("lng")
+        self._location = PostLocation(location_id, loc_name, loc_slug, loc_has_public_page, loc_lat, loc_lng)
```
aandergr, line 563: Seems to be equivalent to (less code):

```suggestion
            loc.update(self._context.get_json("explore/locations/{0}/".format(location_id),
                                              params={'__a': 1})['native_location_data']['location_info'])
        self._location = PostLocation(location_id, loc['name'], loc['slug'], loc['has_public_page'],
                                      loc['lat'], loc['lng'])
```

aandergr (commented):
Nice, the part fixing #1109 seems to work fine, thanks! I just suggest an equivalent, but simpler, change.

Given that the fix for #1129 is still partial and not yet ready to merge, could you split up these two parts into separate PRs, so that we could already merge and release the fix for #1109? (probably it's easiest to just create a new PR for #1109 and after that remove that part from this PR)

I still did not take a closer look to the other part of the PR yet, unfortunately; but I hope I can get to that soon.

otanim:
@aandergr, done.

aandergr:
Seems the browser does an iphone request (we have implemented this request type in InstaloaderContext.get_iphone_json) to `https://i.instagram.com/api/v1/locations/{id}/sections/`, which may also be the way to request subsequent pages.

It's cool to see a working mapping from the new sections-type post response to 'legacy' Post nodes, as currently required by `Post`. This might also be handy for fixing #1080, where the response has changed to that new format. In the long term, we will probably need to change `Post` so that it can work with all types of Post representations that are currently returned by Instagram.

PR #1114, opened 2021-04-25
Show progress of scraping

ekalin:
**What problem does the pull request solve?** Since profiles are not downloaded in the order specified in the command line (and if one downloads followees, one doesn't even know which profiles will be downloaded), it's not possible to know the progress of the scraping process.

**Changes**: This changes the log lines "Retrieving stories from profile..." and "Downloading profile" to include a counter ##/## showing the current profile number and the total (if known).

**Is it just a proof of concept?** I believe it's complete
**Is the documentation updated (if appropriate)?** Doesn't apply
**Do you consider it ready to be merged or is it a draft?** I believe it's ready
**Can we help you at some point?** It's a small change, but nevertheless feel free to suggest improvements.

PR #1242, opened 2021-07-31
Avoid uninitialized variable when called with :stories

ekalin:
This sould fix the problem identified in #1114.


aandergr:
Looks good, thanks!

PR #1244, opened 2021-08-02
Fix for the issue with --geotags: #1109

otanim:
@aandergr,
I've separated #1109 issue's fix from https://github.com/instaloader/instaloader/pull/1240 pull request, please have a look.


aandergr:
Thanks for finding and implementing this fix!

Issue #859, opened 2020-11-06
Highlight with name "..." and several other problems with dots in names on Linux/Windows


UweLau:
**Describe the bug**
I tried to download a profile where the user choose to name all the highlights "..." (three ASCII dots). 
Problem 1: On Linux the highlights are all downloaded but all in the same directory "..." but at least it gets downloaded
Problem 2: On Windows I get a error:
`FileNotFoundError: [Errno 2] No such file or directory: 'xxxxxxx\\...\\2019-11-20_16-13-30_UTC_cover.jpg.temp'`
Problem 3: When I copy over the "..." directory to Windows it behaves very strangely.
Problem 4: When I download some highlight on Windows with a dot at the end, like f.e. "Mun." it gets truncated to "Mun" because Windows wont allow a dot at the end. Linux on the other hand allows the download but I have to manually rename the copy when I tranfer it to Windows so I can access its content.

**To Reproduce**
instaloader --login xxx --password xxx -highlights XXXX with a user that has above mentioned highlight names. Dont have anything that is public ...

**Expected behavior**
Well the question is not easily answered as there are several approaches for that:
Extend the forbidden characters a bit more to also include combinations, like for example "..." or omit ending dots in general. The would have to be on all platforms so one can transfer the files between. The simpler the better. 
This does not solve the problem with naming two or more the same, it would still all go to one directory and would even go further if two highlights are dumbed down so they end up with the same name.

So my idea would be to use the ID number of the highlight itself. I dont know if there is one that is unique or at least not used with the same user but that would be the best solution. Something like "938317195183-D__bed d_w_ed high__ght" would be perfect, or "D__bed d_w_ed high__ght-938317195183".

**Error messages and tracebacks**
see above for the Windows error message

**Instaloader version**
4.5.3

Thammus:
Problems with dots in highlight names have already been reported in other issues, for example #710. 

Copying folders to Windows that have been downloaded on Linux would still be a problem if all problematic names get sanitized on Windows only. One could implement an optional flag that does rigorous replacements of all problematic strings and could be used on all operating systems to maintain the possibility to interchange the folders. 

I had no idea that one profile can have multiple highlights with the same name. Including a unique ID into the folder name is a nice idea to differentiate between those.

UweLau:
@Thammus 
My apologies. I am afraid I did not see issue #710.
As mentioned in my (draft) pull request I would prefer also as an option but I am not qualified enough to do it myself.
[...]

PR #1252, opened 2021-08-09
Fix the issue with folder names starting with dots

otanim:
If you try to use `instaloader` to retrieve highlights that are grouped by naming it `.` or `..` you'll be having a problem with your OS. Besides, files starting with dots in Linux OS will be considered hidden folders, and you'll have some trouble seeing/moving them unless you use extra commands.

This commit uses the existing folder naming sanitization function and extends its capabilities for the above-mentioned case.

Basically, it will replace folders that start with a dot (".") into another similar one ("․")․

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -146,6 +146,10 @@ class _PostPathFormatter(_ArbitraryItemFormatter):
     def sanitize_path(ret: str) -> str:
         """Replaces '/' with similar looking Division Slash and some other illegal filename characters on Windows."""
         ret = ret.replace('/', '\u2215')
+
+        if ret.startswith('.'):
+            ret = ret.replace('.', '\u2024', 1)
+
         if platform.system() == 'Windows':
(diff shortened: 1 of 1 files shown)


Thammus:
The purpose of the `sanitize_path()` function that your pull request touches is merely to alleviate problems caused by characters that are invalid as part of filenames under Windows. As far as I know folders and filenames starting with a dot do not lead to errors on common OS. However, the described problem that users might get confused by hidden folders is a valid case.
Instaloader has a lot known Issues related to pathnames such as #185, #575, #710 and #859. Like described in https://github.com/instaloader/instaloader/pull/854#issuecomment-727220601, it would be nice to have a flag `--sanitizepaths` to do a proper path sanitation without having to worry about issues with backwards incompatibility.
Nevertheless, I think that substituting leading periods in pathnames by a similar looking character is a change that has more advantages than it might cause problems.

otanim:
You're welcome.

PR #1311, opened 2021-10-15
extend profile data

mehdignu:
- A motivation for this change:
  - extend the profile data to : business_phone_number, mutual_followers, business_email

- The completeness of this change:
  - minor changes ready to merge

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -868,13 +868,16 @@ class Profile:
         else:
             return self._metadata("profile_pic_url_hd")
 
-    def business_email(self) -> str:
-        return self._metadata('business_email')
+    @property
+    def business_email(self) -> Optional[str]:
+        return self._metadata("business_email")
 
-    def business_phone_number(self) -> str:
-        return self._metadata('business_phone_number')
+    @property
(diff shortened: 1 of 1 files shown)


instaloader/structures.py
```diff
@@ -1073,6 +1085,13 @@ def owner_id(self) -> int:
         """The ID of the StoryItem owner."""
         return self.owner_profile.userid
 
+    @property
+    def get_mentions(self) -> List[dict]:
+        """get the user mentions in the story"""
+        tappable_objects = self._node['tappable_objects']
+        mentions = [obj for obj in tappable_objects if obj['__typename'] == 'GraphTappableMention']
+        return mentions
+
```
Thammus, line 1094: - Property names should not start with "get"
- It would be much nicer to get `Profile` instances instead of unspecific dicts
- """Mentioned :class:`Profile`s in the StoryItem""" might then be a better docstring

Thammus (changes requested):
Since this pull request introduces new features instead of fixing a bug, could you please base it on the `upcoming/v4.9` branch instead of `master`?

Issue #1305, opened 2021-10-09
%z in datetime style formatting options doesn't return offset


SpL1Ne:
**Describe the bug**
%z in datetime style formatting options doesn't return offset

For example:
- `--latest-stamps` option for profile download return `2021-10-03T09:49:02.000000+0300` timestamp
- For latest post from same profile `--post-metadata-txt="{date:%Y-%m-%dT%H:%M:%S.%f%z}"` returns `2021-10-03T06:49:02.000000`
- For same post `--post-metadata-txt="{date_local:%Y-%m-%dT%H:%M:%S.%f%z}"` returns `2021-10-03T09:49:02.000000`

And in both cases %z is empty

**To Reproduce**
`instaloader --post-metadata-txt="{date_local:%Y-%m-%dT%H:%M:%S.%f%z}" -- -<post_id>`

**Expected behavior**
For second case `--post-metadata-txt` %z should return +0300

**Error messages and tracebacks**
None

**Instaloader version**
4.8.1

**Additional context**
None.

aandergr:
Indeed, thanks for reporting! `date_utc` and `date_local` currently are naive datetime objects which do not carry timezone information, hence `%z` is empty.

aandergr:
Fixed in [Version 4.8.3](https://github.com/instaloader/instaloader/releases/tag/v4.8.3).

aandergr:
I have postponed the change that date_local returns a timezone-aware datetime object to version 4.9, and date_utc remains a timezone-unaware datetime object, as a compromise, especially to maintain compatibility between minor releases of version 4.8, and still provide a timezone-aware datetime object as of 4.9 as date_local, without breaking backwards compatibility in date_utc comparisons. Accordingly, I have also removed this change from 4.8 and moved it to 4.9.

PR #1316, opened 2021-10-19
datetime style formatting options doesn't return offset

misaelitox:
  - Fixes issue #1305.
  - More general: from `--post-metadata-txt="{date_local:%Y-%m-%dT%H:%M:%S.%f%z}"`  %z in datetime style formatting options doesn't return offset

This is my very first time collaborating with the community, also this is one of my first projects with python, with this implementation what I do is substracting creation date in local time zone vs creation date in UTC, with that offset I can obtain the timezone info.

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -3,7 +3,7 @@ import lzma
 import re
 from base64 import b64decode, b64encode
 from collections import namedtuple
-from datetime import datetime
+from datetime import datetime, timezone, timedelta
 from pathlib import Path
 from typing import Any, Dict, Iterable, Iterator, List, Optional, Tuple, Union
 
@@ -225,16 +225,14 @@ class Post:
     @property
     def date_local(self) -> datetime:
         """Timestamp when the post was created (local time zone)."""
-        return datetime.fromtimestamp(self._node["date"]
-                                      if "date" in self._node
-                                      else self._node["taken_at_timestamp"])
+        timestamp_date = self.get_timestamp_date_created()
+        tzinfo = timezone(timedelta(hours=self.get_timezone_offset(timestamp_date)))
+        return datetime.fromtimestamp(timestamp_date, tzinfo)
 
     @property
     def date_utc(self) -> datetime:
         """Timestamp when the post was created (UTC)."""
-        return datetime.utcfromtimestamp(self._node["date"]
-                                         if "date" in self._node
-                                         else self._node["taken_at_timestamp"])
+        return datetime.utcfromtimestamp(self.get_timestamp_date_created())
 
     @property
     def date(self) -> datetime:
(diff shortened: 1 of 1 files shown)


instaloader/structures.py
```diff
@@ -275,6 +273,17 @@ def mediacount(self) -> int:
...
+    def get_timestamp_date_created(self) -> float:
+        """Timestamp when the post was created"""
+        return (self._node["date"]
+                if "date" in self._node
+                else self._node["taken_at_timestamp"])
+
+    def get_timezone_offset(self, timestamp_date) -> float:
+        """Timestamp offset for a given date"""
+        diff_total_seconds = (datetime.fromtimestamp(timestamp_date) - datetime.utcfromtimestamp(timestamp_date)).total_seconds()
+        return diff_total_seconds / 3600
```
Thammus, line 285: Substracting  `datetime.utcfromtimestamp()` from `datetime.fromtimestamp()` produces a `datetime.timedelta` object which is needed inside the calculation of the `date_local` property to determine the `tzinfo`. Converting it forth and back into hours is not necessary and `get_timezone_offset()` could return the `timedelta` directly.

Since `get_timezone_offset()` does not use `self` and only uses the`timestamp_date` parameter to calculate the offset, it does not need to be a class method.

Thammus (changes requested):
Thanks for your suggested code changes and the pull request.
After a short research this indeed seems to be the most reliable and clean way of calculating and attaching timezone info for a local datetime.

Although the attached issue #1305 only refers to posts, it would be nice to also have these aware datetime objects for stories. The story class also has properties `date_local` and `date_utc` which use a similar logic. Would you be so kind as to adapt these properties too?

instaloader/structures.py
misaelitox: I have just applied feedback as per the comments above, thank you!

Thammus:
Thank you for this fix and for applying the suggested changes 🎉

misaelitox:
> Thank you for this fix and for applying the suggested changes 🎉

My pleasure!

ekalin:
It seems that the objects returned by `time_utc` still don't have the timezone information, as `datettime.utcfromtimestamp()` returns a naive object: https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp

aandergr:
> It seems that the objects returned by `time_utc` still don't have the timezone information, as `datettime.utcfromtimestamp()` returns a naive object: https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp

Thanks for that comment! I changed Post.date_utc and StoryItem.date_utc such that they return aware datetime objects now.

aandergr:
I have postponed the change that date_local returns a timezone-aware datetime object to version 4.9, and date_utc remains a timezone-unaware datetime object, as a compromise, especially to maintain compatibility between minor releases of version 4.8, and still provide a timezone-aware datetime object as of 4.9 as date_local, without breaking backwards compatibility in date_utc comparisons. Accordingly, I have also removed this change from 4.8 and moved it to 4.9.

PR #1321, opened 2021-10-26
Fix KeyError for video_url, and better workflow

### LINKED ISSUE #1320: `StoryVideo` has no `video_url` field and thus causes KeyError when parsing
**Describe the bug**
As mentioned in #1238, the StoryVideo returned by single post API doesn't behave like GraphStoryVideo returned by Stories timeline either. It has no self._node['video_url'] or self._node['video_resources']. So you can only get the video from self._iphone_struct['video_versions'].

Previously you can still parse it successfully if you logged in and does not disable iPhone API, because it can still pull the videos from the iPhone API. However, after changing the line at 
https://github.com/instaloader/instaloader/blob/f09bd51ed416710b47226b3e91faefa14c85e91a/instaloader/structures.py#L380
it will errors out due to such key does not exist.

**To Reproduce**

1. find any story video and copy the URL as `url`
```python
import instaloader
url = "{put URL here}"
username = '{your username}'
if m := re.search(r'https://www.instagram.com/stories/[^/]+?/(\d+?)/', .url):
    shortcode = instaloader.Post.mediaid_to_shortcode(int(m[1])) 
else:
    raise Exception(f'Cannot parse URL {url}!')
L = instaloader.Instaloader()
L.load_session_from_file(username)
post = instaloader.Post.from_shortcode(L.context, shortcode)
L.download_post(post, target='')

```

fireattack:
Fixes #1320.

Also improves the workflow a little bit.

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -377,20 +377,23 @@ class Post:
     def video_url(self) -> Optional[str]:
         """URL of the video, or None."""
         if self.is_video:
-            version_urls = [self._field('video_url')]
+            version_urls = []
+            try:
+                version_urls.append(self._field('video_url'))
+            except (InstaloaderException, KeyError, IndexError) as err:
+                self._context.error(f"Warning: Unable to fetch video from graphql of {self}: {err}")
             if self._context.iphone_support and self._context.is_logged_in:
                 try:
                     version_urls.extend(version['url'] for version in self._iphone_struct['video_versions'])
                 except (InstaloaderException, KeyError, IndexError) as err:
-                    self._context.error(f"Unable to fetch high-quality video version of {self}: {err}")
-                    return version_urls[0]
-            else:
+                    self._context.error(f"Unable to fetch high-quality video version of {self}: {err}")                    
+            version_urls = list(dict.fromkeys(version_urls))
+            if len(version_urls) == 0:
+                    return None
+            if len(version_urls) == 1:
                 return version_urls[0]
             url_candidates: List[Tuple[int, str]] = []
             for idx, version_url in enumerate(version_urls):
-                if any(url_candidate[1] == version_url for url_candidate in url_candidates):
-                    # Skip duplicates
-                    continue
                 try:
                     url_candidates.append((
                         int(self._context.head(version_url, allow_redirects=True).headers.get('Content-Length', 0)),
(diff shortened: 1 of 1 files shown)


SpyHoodle:
Attribute `video_url` in object `StoryItem` also needs to be fixed, creates the same errors

fireattack:
`StoryItem` actually uses `video_resources`, and does not suffer from this problem (graphql will return URLs). But it doesn't hurt to copy the workflow there.

instaloader/structures.py
```diff
@@ -377,20 +377,23 @@ def is_video(self) -> bool:
...
+            except (InstaloaderException, KeyError, IndexError) as err:
+                self._context.error(f"Warning: Unable to fetch video from graphql of {self}: {err}")
             if self._context.iphone_support and self._context.is_logged_in:
                 try:
                     version_urls.extend(version['url'] for version in self._iphone_struct['video_versions'])
                 except (InstaloaderException, KeyError, IndexError) as err:
                     self._context.error(f"Unable to fetch high-quality video version of {self}: {err}")
-                    return version_urls[0]
-            else:
+            version_urls = list(dict.fromkeys(version_urls))
```
aandergr, line 390: This is to remove duplicates, right? It might be a little cleaner to use a `set` for `version_urls` rather than a list. It does not store the order of items, so we will loose the enumeration in the warning message "Video URL candidate 1/4 for ..."; but neither does `list(dict.fromkeys(...))` ensure that the order is kept. But that's okay anyway, there is probably little value in printing the index of the URL that fails.

instaloader/structures.py
```diff
@@ -1129,20 +1132,23 @@ def is_video(self) -> bool:
...
```
aandergr, line 1145: (same here)

aandergr (commented):
Hi! Sorry for my late response, I have been quite busy the last few weeks.

The change looks very good, I have only inline-commented a little suggestion.

I have noticed that one of the Actions failed, fixed that, and rebased your branch on the now-current master.

instaloader/structures.py
fireattack: Yeah, this is to remove duplicates. I prefer this way than `set` exactly because of the order preservation (even though not very important here), and IIRC since Py 3.5 dicts are ordered, so this operation will keep the order of items.

aandergr: Ah interesting, did not know that. Then it is fine. Thanks!

pafke2:
I face a lot of "JSON Query to graphql/query: 429 Too Many Requests" with a table when trying to download Igtv, stories and posts don't have such issue. All of that with disabled iphone-support

Issue #1143, opened 2021-05-18
Same comment being scraped hundreds of times in some posts


ashramen:
**Describe the bug**
Scraping from certain posts using get.comments() scrapes some of the same comments repeatedly.

**To Reproduce**
`
L = instaloader.Instaloader()

L.login("USERNAME", "PASSWORD")    

post = instaloader.Post.from_shortcode(L.context, "CN_BVM1hXgO")
for comment in post.get_comments():
    print(comment.text)
`
This is an example of a post where this occurs with shortcode CN_BVM1hXgO

**Expected behavior**
Each comment is scraped and printed once.

**Error messages and tracebacks**
The same few comments are scraped hundreds of times until the account receives a 429 Too Many Requests error

**Instaloader version**
4.7.1

**Additional context**
This error seems to only occur for some posts. I have scraped hundreds of other posts without any problems.

miriq:
Can confirm this happens sometimes. And it's a server-side bug, if you try to load comments using the browser it also shows the same comment over and over again.

Here is how I fixed it, check if two consecutive comments have the same id and abort:

```diff
diff --git a/instaloader/structures.py b/instaloader/structures.py
index 9059656..396dc1c 100644
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -460,24 +460,34 @@ class Post:
         def _postcomment(node):
             return PostComment(*_postcommentanswer(node),
                                answers=_postcommentanswers(node))
-        if self.comments == 0:
-            # Avoid doing additional requests if there are no comments
-            return []
 
-        comment_edges = self._field('edge_media_to_comment', 'edges')
-        answers_count = sum([edge['node'].get('edge_threaded_comments', {}).get('count', 0) for edge in comment_edges])
+        def inner(self):
+            if self.comments == 0:
+                # Avoid doing additional requests if there are no comments
+                return []
 
-        if self.comments == len(comment_edges) + answers_count:
-            # If the Post's metadata already contains all parent comments, don't do GraphQL requests to obtain them
-            return [_postcomment(comment['node']) for comment in comment_edges]
-        return NodeIterator(
-            self._context,
-            '97b41c52301f77ce508f55e66d17620e',
-            lambda d: d['data']['shortcode_media']['edge_media_to_parent_comment'],
-            _postcomment,
-            {'shortcode': self.shortcode},
-            'https://www.instagram.com/p/{0}/'.format(self.shortcode),
-        )
[...]

Issue #910, opened 2020-12-07
Endless loop when receiving comments


djigit0s:
**Endless loop when receiving comments for some links**
For example, the link is https://www.instagram.com/p/CGh4slpJ0IL/

**To Reproduce**
```
count = 0
for item in post_info.get_comments():
    count +=1
    print('count')
    print(item)


```

**Instaloader version**
`instaloader --version` output.
4.5.4

**Additional context**
The cycle goes on forever and the count increases, although the post only has 340 comments
![Снимок](https://user-images.githubusercontent.com/3939174/101407794-7a496600-38ec-11eb-8710-bf9b1f7682c5.JPG)

aandergr:
Thanks for reporting this. Indeed, I can reproduce that some comments are looped endlessly for this post.

Thammus:
This endless loop bug seems to be fixed by Instagram.

PR #1347, opened 2021-11-23
Endless loop when receiving comments

### LINKED ISSUE #1017: Post.get_comments() creates an infinite loop, fetching same comment
**Describe the bug**
I'm using Post.get_comments() function in a for loop and it had been working well but I found an exception with one particular comment in a post.
The for loop keeps fetching the same comment indefinitely. I can't see what is particular to this comment that is causing this behaviour, the comment that is been fetched continously is an answer of a comment, it has hashtags and that's all.

**To Reproduce**
```
>>> import instaloader
>>> username = 'xxxxx'
>>> L = instaloader.Instaloader()
>>> try:
...     L.load_session_from_file(username)
... except instaloader.BadCredentialsException:
...     L.login(username, 'mypassword')
...     L.save_session_to_file()
... except instaloader.TwoFactorAuthRequiredException:
...     L.two_factor_login()
...     L.save_session_to_file()
... except Exception as e:
...     print(e)
>>> the_post_with_issues = instaloader.Post.from_shortcode(L.context, 'CKznw1jDJle')
>>> for comment in the_post_with_issues.get_comments():
...     print(comment.text)
```

**Expected behavior**
Fetch all comments, without duplicates

No error message, just keeo fetching the same comment

**Instaloader version**
4.6.1

JeremyKj87:
- **A motivation for this change, e.g.**
  - Fixes #1017 #1143 #910 
  - More general: 
There is currently a problem with Instagram that sporadically the comments do not come to an end and can be scraped endlessly.
The has_next_page-flag (inside the request response) is always true although the comments are repeated.
This error does not always occur, but mainly with larger profiles and the consequence is that you quickly run into an HTTP 429 Error (Too Many Requests).
The problem can also be reproduced in the browser

Test data: https://www.instagram.com/p/CVyEdhbIXPG/

- **The changes proposed in this pull request**
-  change the nodeiterator __next__ method to check if the response is the same than before

  - Is it just a proof of concept? It's just a poc 
  - Can we help you at some point?

### DIFF
--- a/instaloader/nodeiterator.py
+++ b/instaloader/nodeiterator.py
@@ -133,14 +133,15 @@ class NodeIterator(Iterator[T]):
             return item
         if self._data['page_info']['has_next_page']:
             query_response = self._query(self._data['page_info']['end_cursor'])
-            page_index, data = self._page_index, self._data
-            try:
-                self._page_index = 0
-                self._data = query_response
-            except KeyboardInterrupt:
-                self._page_index, self._data = page_index, data
-                raise
-            return self.__next__()
+            if self._data['edges'] != query_response['edges']:
+                page_index, data = self._page_index, self._data
+                try:
+                    self._page_index = 0
+                    self._data = query_response
+                except KeyboardInterrupt:
+                    self._page_index, self._data = page_index, data
+                    raise
+                return self.__next__()
         raise StopIteration()
 
     @property


Thammus:
The change looks good and works fine 🌈. Many thanks for fixing this recurring bug 🙂

PR #1372, opened 2022-01-02
Remove unnecessary conversion to timezone aware timestamp

ekalin:
Since `Post.date_local` now already returns a timestamp with the appropriate timezone, it's not necessary to add timezone information to timestamps when comparing dates for `LatestStamps`.


aandergr:
Thanks!

aandergr:
I have postponed the change that date_local returns a timezone-aware datetime object to version 4.9, and date_utc remains a timezone-unaware datetime object, as a compromise, especially to maintain compatibility between minor releases of version 4.8, and still provide a timezone-aware datetime object as of 4.9 as date_local, without breaking backwards compatibility in date_utc comparisons. Accordingly, I have also removed this change from 4.8 and moved it to 4.9.

Issue #580, opened 2020-04-19
How to specify directory for grouped highlights?


renox92:
If I download them to `{profile}/highlights` then they all get bunched up in that folder, if I download to `{profile}/highlights/{target}`, then I end up with a path like
```
profile1/highlights/profile1/highlight_group1/
profile1/highlights/profile1/highlight_group2/
profile1/highlights/profile1/highlight_group3/
...
```

It doesn't look like highlight group can be used by itself in a dirname to get a `profile1/highlights/highlight_groupN/` path, so maybe it's possible to change .py to not include profile name when the target is a highlight?

PR #1380, opened 2022-01-16
Fix --dirname-pattern {target} placeholder (#580)

FractaIism:
This PR fixes issue #580

The issue is that when downloading highlights using --dirname-pattern, the {target} placeholder was being expanded to {profile}/&lt;highlight-title&gt;, with no way to separate these two components from the command line.

The changes in this PR make it so that {target} is replaced only with &lt;highlight-title&gt;, while changing the default --dirname-pattern to {profile}/{target} for backward compatibility.


aandergr:
> This PR fixes issue #580

You cannot 'fix' a question.

We cannot merge this PR, it introduces backwards incompatible changes.

PR #1413, opened 2022-02-16
check if profile is private when downloading stories

### LINKED ISSUE #1402: Attempting to download stories from a private account not followed using --no-posts doesn't result in error
**Describe the bug**
While attempting to download stories with an account which does not follow a particular user, I don't get an error message if I use the --no-posts flag.

**To Reproduce**
Type this command, with private_account_not_followed and your_account replaced with the appropriate accounts:
instaloader private_account_not_followed --login=your_account --no-posts --stories

**Expected behavior**
I expect there to be an error message like if I try to download posts from this user with the same account. If I try to download with posts, I get the message private but not followed.

**Instaloader version**
4.8.3

0xwho7:

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -1394,7 +1394,7 @@ class Instaloader:
                     self.save_metadata_json(json_filename, profile)
 
                 # Catch some errors
-                if tagged or igtv or highlights or posts:
+                if tagged or igtv or highlights or posts or stories:
                     if (not self.context.is_logged_in and
                             profile.is_private):
                         raise LoginRequiredException("--login=USERNAME required.")


GowthamChowdary:
#1402

aandergr:
Thanks for attempting to fix this issue. Unfortunately there are cases where this change tremendously slows down the process of downloading stories. The reason is that Instaloader generally attempts to use the minimum number of requests necessary, and the profile instances given to download_profiles() can be sparse instances, not carrying the full metadata and especially not having the fields `is_private` and `followed_by_viewer`. When these fields are accessed, Instaloader will do an extra query to fetch the full metadata structure of the given profiles. This is not a problem when downloading tagged, igtv, highlights or regular posts, as in these cases, the full metadata structure is required anyway. However, for only downloading stories, Instaloader does not need to have the complete metadata present. 

This can be reproduced by invoking Instaloader to download only the stories of followees, which is a case were incomplete Profile instances are used:
```
instaloader --stories --no-posts --no-profile-pic --login USER @USER 
```
This invocation is slowed down significantly with the here-proposed change.

wzup:
Why does it require login for public stories?
Documentation mentions that login is required only for private stories [link](https://instaloader.github.io/cli-options.html#login-download-private-profiles)
```
$ instaloader --no-posts --no-videos --no-profile-pic --stories appletv
Login is required to download stories.
Stored ID 6672060127 for profile appletv
```

PR #1434, opened 2022-03-08
Hack: rename webp images to jpeg

### LINKED ISSUE #1414: Images are being downloaded as .webp instead of .jpg
**Describe the bug**
Images are being downloaded as `.webp` files instead of the usual `.jpg` files.

**To Reproduce**
I'm using the Python module.
`L = instaloader.Instaloader(download_geotags=False, compress_json=False, filename_pattern='{mediaid}')`
I first saved thousands of shortcodes from a particular hashtag by iterating through:
```
post_iterator = instaloader.NodeIterator(
    L.context, "9b498c08113f1e09617a1703c22b2f32",
    lambda d: d['data']['hashtag']['edge_hashtag_to_media'],
    lambda n: instaloader.Post(L.context, n),
    {'tag_name': hashtagname},
    f"https://www.instagram.com/explore/tags/{hashtagname}/"
)
```
Then I downloaded the post using:
```
post = instaloader.Post.from_shortcode(L.context, 'shortcode')
L.download_post(post, target='target')

```
This is one of the posts that saved as `.webp`:
`'CZbzQ5vL0tB'`

**Expected behavior**
Save the image as a `.jpg` instead of `.webp`

N/A

**Instaloader version**
4.8.4

**Additional context**
I had not tried scraping posts since about two weeks ago so it seems like this is something new. I'm only getting `.webp` files for posts from this year (although not all). I downloaded 417 posts from this month (February) and the image files I got were 638 jpgs and 106 webps.
I'm not expecting this to be fixed if this is considered an non-issue but I thought I should bring it up

miriq:
Fixes #1414

For some reason Instagram incorrectly sets the file extension to webp, even though the image is a jpg. This pull request assumes that webp images are always jpgs, that is verified using the magic header bytes "ffd8ff". If for some reason the header bytes do not match, it will print an error and exit.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -314,6 +314,12 @@ class Instaloader:
         Returns true, if file was actually downloaded, i.e. updated."""
         urlmatch = re.search('\\.[a-z0-9]*\\?', url)
         file_extension = url[-3:] if urlmatch is None else urlmatch.group(0)[1:-1]
+        webp_hack = False
+        if file_extension == "webp":
+            # https://github.com/instaloader/instaloader/issues/1414
+            # Instagram incorrectly sets the file extension to webp, even though the image is a jpg
+            file_extension = "jpg"
+            webp_hack = True
         if filename_suffix is not None:
             filename += '_' + filename_suffix
         filename += '.' + file_extension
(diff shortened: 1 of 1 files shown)


afterdelight:
so if it was a real webp it will reject the webps to be downlaoded?

miriq:
In that case it will download the webp but save it as jpg, and print an error. This way was easier to implement: assume that instagram always returns jpeg, and if in the future that's not true then hopefully somebody can fix it. Or maybe even remove this hack, because it looks like a bug on instragram's side.

afterdelight:
OK

aandergr:
Hi! Thanks for your effort and for sharing your ideas on how to fix this. I prefer the approach outlined in #1447 to use the Content-Type header to determine an appropriate file extension.

miriq:
Ok no worries, I had been using this workaround for a few days and I opened this pull request because nobody else had a fix, now that it is properly fixed this hack is not needed anymore.

PR #1440, opened 2022-03-13
Add --sanitize-paths option

canh:
This PR attempts to fix #854
- UnitTests are passing and the code seems to be working as intended.
- cli-options.rst is updated.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,7 +20,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install Dependencies
         run: |
-          python -m pip install pipenv==2020.11.15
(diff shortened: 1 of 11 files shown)


aandergr (commented):
Hi! Thanks a lot for this pull request!

It extends the windows path sanitization logic to handle reserved device names (fixes #185), and dots at the end of the filename (fixes #710), and adds a flag `--sanitize-paths` to always enable that logic (to fix #859). This looks pretty good overall and very well integrated into the code base. Even if there are any remaining cases of Windows-incompatible path names that are not yet handled, the logic can be easily extended.

Currently it is based on `master` branch. Would you be so kind to rebase it to `upcoming/v4.9`?

canh:
I messed up git :) Will send another PR.

PR #1447, opened 2022-03-15
Using header to determine file extension

### LINKED ISSUE #1414: Images are being downloaded as .webp instead of .jpg

fireattack:
Fixes #1414.

Please consider this WIP since I need some input about how to implement certain things, particularly about how to detect existing local file.

My idea is to:

1. detect with apparent extension (.webp) first before making the request; 
2. detect with potential *real* extension as well -- but this has to be hard-coded as ".jpg". (*not implemented yet*)
3. and then detect again once we got the correct filename (that means after we made the request, but haven't started downloading). If we already did 2, I don't know if this really is necessary, though (*not implemented yet*).

What do you think?

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -312,15 +312,12 @@ class Instaloader:
                      filename_suffix: Optional[str] = None, _attempt: int = 1) -> bool:
         """Downloads and saves picture with given url under given directory with given timestamp.
         Returns true, if file was actually downloaded, i.e. updated."""
-        urlmatch = re.search('\\.[a-z0-9]*\\?', url)
-        file_extension = url[-3:] if urlmatch is None else urlmatch.group(0)[1:-1]
         if filename_suffix is not None:
             filename += '_' + filename_suffix
-        filename += '.' + file_extension
         if os.path.isfile(filename):
             self.context.log(filename + ' exists', end=' ', flush=True)
             return False
-        self.context.get_and_write_raw(url, filename)
+        filename = self.context.get_and_write_raw(url, filename, get_extension=True)
         os.utime(filename, (datetime.now().timestamp(), mtime.timestamp()))
         return True
 
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -325,7 +325,9 @@ class Instaloader:
             header_extension = '.' + resp.headers['Content-Type'].split(';')[0].split('/')[-1]
             header_extension = header_extension.lower().replace('jpeg', 'jpg')
             filename += header_extension
-        if os.path.isfile(filename):
+        else:
+            filename = nominal_filename
+        if filename != nominal_filename and os.path.isfile(filename):
             self.context.log(filename + ' exists', end=' ', flush=True)
             return False
         self.context.write_raw(resp, filename)
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -320,6 +320,7 @@ class Instaloader:
         if os.path.isfile(nominal_filename):
             self.context.log(nominal_filename + ' exists', end=' ', flush=True)
             return False
+
         resp = self.context.get_raw(url)
         if 'Content-Type' in resp.headers and resp.headers['Content-Type']:
             header_extension = '.' + resp.headers['Content-Type'].split(';')[0].split('/')[-1].lower().replace('jpeg', 'jpg')
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -314,8 +314,13 @@ class Instaloader:
         Returns true, if file was actually downloaded, i.e. updated."""
         if filename_suffix is not None:
             filename += '_' + filename_suffix
-        if os.path.isfile(filename):
-            self.context.log(filename + ' exists', end=' ', flush=True)
+        urlmatch = re.search('\\.[a-z0-9]*\\?', url)
+        file_extension = url[-3:] if urlmatch is None else urlmatch.group(0)[1:-1]
+        if filename_suffix is not None:
+            filename += '_' + filename_suffix
+        nominal_filename = filename + '.' + file_extension
+        if os.path.isfile(nominal_filename):
+            self.context.log(nominal_filename + ' exists', end=' ', flush=True)
             return False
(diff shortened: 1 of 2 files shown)


instaloader/instaloader.py
```diff
@@ -312,15 +312,17 @@ def download_pic(self, filename: str, url: str, mtime: datetime,
                      filename_suffix: Optional[str] = None, _attempt: int = 1) -> bool:
         """Downloads and saves picture with given url under given directory with given timestamp.
         Returns true, if file was actually downloaded, i.e. updated."""
+        if filename_suffix is not None:
+            filename += '_' + filename_suffix
```
aandergr, line 316: This seems to be done twice now.

instaloader/instaloader.py
```diff
@@ -312,15 +312,17 @@ def download_pic(self, filename: str, url: str, mtime: datetime,
...
             filename += '_' + filename_suffix
-        filename += '.' + file_extension
-        if os.path.isfile(filename):
-            self.context.log(filename + ' exists', end=' ', flush=True)
+        nominal_filename = filename + '.' + file_extension
+        if os.path.isfile(nominal_filename):
+            self.context.log(nominal_filename + ' exists', end=' ', flush=True)
             return False
-        self.context.get_and_write_raw(url, filename)
+        filename = self.context.get_and_write_raw(url, filename, get_extension=True)
```
aandergr, line 325: I think a good way to go for `download_pic()` would be:

1. Keep the logic that determines the filename with the extension retrieved from the URL and checks whether that file already exists
2. Determine an appropriate filename extension using the Content-Type
3. Check whether that file already exists, return False if so
4. Save the file, update its utime and return True

instaloader/instaloadercontext.py
```diff
@@ -521,13 +521,18 @@ def get_raw(self, url: str, _attempt=1) -> requests.Response:
...
         """Downloads and writes anonymously-requested raw data into a file.
 
         :raises QueryReturnedNotFoundException: When the server responds with a 404.
         :raises QueryReturnedForbiddenException: When the server responds with a 403.
         :raises ConnectionException: When download repeatedly failed."""
-        self.write_raw(self.get_raw(url), filename)
+        resp = self.get_raw(url)
+        if get_extension and 'Content-Type' in resp.headers and resp.headers['Content-Type']:
+            extension = '.' + resp.headers['Content-Type'].split(';')[0].split('/')[-1].lower().replace('jpeg', 'jpg')
+            filename += extension
```
aandergr, line 533: It would be probably much easier to have this in `download_pic()` and then let it call `get_raw()` and `write_raw()` rather than `get_and_write_raw()`, since we'd have to check whether a file with the appropriate extension already exists.

instaloader/instaloader.py
aandergr: > would requests to image server count into the quota that triggers "too many requests"?

No, they do not count into the quota.

Thanks! :)

fireattack:
Thanks, I'm done I believe.

instaloader/instaloader.py
```diff
@@ -312,15 +312,23 @@ def download_pic(self, filename: str, url: str, mtime: datetime,
...
             filename += '_' + filename_suffix
-        filename += '.' + file_extension
+        urlmatch = re.search('\\.[a-z0-9]*\\?', url)
+        file_extension = url[-3:] if urlmatch is None else urlmatch.group(0)[1:-1]
+        nominal_filename = filename + '.' + file_extension
+        if os.path.isfile(nominal_filename):
+            self.context.log(nominal_filename + ' exists', end=' ', flush=True)
+            return False
+        resp = self.context.get_raw(url)
+        if 'Content-Type' in resp.headers and resp.headers['Content-Type']:
```
aandergr, line 324: If the condition of this `if` evaluates to False, `filename` lacks the file extension, and later the file is saved without any valid extension.

instaloader/instaloader.py
```diff
@@ -312,15 +312,23 @@ def download_pic(self, filename: str, url: str, mtime: datetime,
...
+            header_extension = '.' + resp.headers['Content-Type'].split(';')[0].split('/')[-1]
+            header_extension = header_extension.lower().replace('jpeg', 'jpg')
+            filename += header_extension
         if os.path.isfile(filename):
             self.context.log(filename + ' exists', end=' ', flush=True)
             return False
```
aandergr, line 330: This is actually only needed if a different file extension was determined using the Content-Type header.

fireattack: I thought about it; but if the file with the same extension already existed, we have already returned before. 

So while is kinda redundant, it would help in some rare cases the file became existing during the first check and after the request is made. 

What do you think?

fireattack: Anyway, I modified it so it will only check if the ext is different.

aandergr:
Thanks!

PR #1452, opened 2022-03-16
Add --sanitize-paths option

canh:
This PR attempts to fix #854

- UnitTests are passing and the code seems to be working as intended.
- cli-options.rst is updated.

### DIFF
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -239,6 +239,13 @@ How to Download
 
    .. versionadded:: 4.8
 
+.. option:: --sanitize-paths
+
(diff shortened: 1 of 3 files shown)


aandergr:
Just for the record ;) This fixes #185, fixes #859, fixes #710 and supersedes #1440.

One day we might also have a flag `--restrict-filenames`, that rigorously restrict filenames to only ASCII characters, and avoids "&" and spaces in filenames; in addition to our now-new `--sanitize-paths` that solves path name related problems the way Instaloader's users are accustomed to.

Issue #1446, opened 2022-03-15
Emoji in hashtags


MiguelX413:
**Describe the bug**
Some Japanese hashtags aren't processed correctly. Like [#はなすたぐらむ🌺](https://www.instagram.com/explore/tags/はなすたぐらむ🌺/) in https://instagram.com/p/Ca4Rc0XLcQP/ being returned as 'はなすたぐらむ' and the [#黒髪ロング](https://www.instagram.com/explore/tags/黒髪ロング/) in https://instagram.com/p/CbHXJ4Gv3dM/ being returned as '黒髪ロンク'.
I am using .caption_hashtags

**To Reproduce**
Use .caption_hashtags on these posts

**Expected behavior**
I expect 'はなすたぐらむ🌺' and '黒髪ロング' to be returned.

**Instaloader version**
4.8.4

**Additional context**
https://github.com/MiguelX413/IgTgBot/tree/779ac663a28e43db81d1740bf8b73ceb5547037c

MiguelX413:
The emoji issue is still not resolved by my fix so this issue is only partially solved now.

MiguelX413:
I think this should probably be reopened.

Issue #978, opened 2021-01-16
Regex for parsing non-English hashtags fails for certain character sets


bluedrops:
**Describe the bug**
I spent a few hours trying to fix this myself but was unable to as I'm less experienced with regex, so I'll log a bug report with my findings.

I've noticed that most foreign languages are parsed correctly as far as I can tell, such as Chinese, Arabic, Korean, and most Japanese.
However, I noticed that certain Japanese characters appear to be treated as delimiters rather than continuing to match.

**To Reproduce**
```
def test():
    input_str = '''
    #プレ #フレブル #プレイヤー #プレ好き #ぷれ #プレゼント
    '''
    # Pasted from Instaloader source code for simplicity
    hashtag_regex = re.compile(r"(?:#)(\w(?:(?:\w|(?:\.(?!\.))){0,28}(?:\w))?)")
    hashtags = re.findall(hashtag_regex, input_str.lower())

    print(hashtags)
test()
```

**Expected behavior**
```
['プレ', 'フレブル', 'プレイヤー', 'プレ好き', 'ぷれ', 'プレゼント']
```

**Actual behavior**
```
['プレ', 'フレブル', 'フ', 'フ', 'ふ', 'プレゼント']
```

aandergr:
Interesting, though probably more related to Python's unicode handling than to Instaloader.

The hashtag regex matches contiguous letters after a `#`. `プ` is two characters (need to backspace twice to erase it) and contains a katakana-hiragana semi-voiced sound mark (which is not a letter), whereas `プ`, despite looking equal, is a single character interpreted as a letter. So it fails matching `#プレイヤー`, matching it only up to `フ`, whereas `#プレゼント` is matched completely.

How does the Instagram app handle these two hashtags?

bluedrops:
That makes sense! I can see why the hashtag regex breaks on these.

Interestingly, the Instagram app handles the hashtags as expected.

PR #1453, opened 2022-03-16
Add unicode normalization of captions for improved support of other…

MiguelX413:
… languages and their hashtags

This attempts to partially fix https://github.com/instaloader/instaloader/issues/1446 with support for hashtags in other scripts.
It is a very minor change.
Captions will be unicode normalized and hashtags and mentions will be better captured


aandergr:
Interesting finding. Indeed, this does seem to fix a subset of the problems mentioned in #1446 and #978, and seems to not break the comparison operator for strings unaffected by this issue, including ones that contain emojis.

Issue #1455, opened 2022-03-17
Japanese following a mention without spacing is detected as part of the mention


MiguelX413:
**Describe the bug**
When Japanese follows a mention without spaces, it is detected as part of the mention, so a post that has “@MiguelX413は” would return 'miguelx413は' in caption_mentions.

**Expected behavior**
I expect just 'miguelx413' to be returned instead.

**Instaloader version**
4.8.4

**Additional context**
[Add any other context about the problem here.](https://github.com/MiguelX413/IgTgBot/tree/3bf05852755848ae302335c9ffd42582f5fabbe2)

PR #1456, opened 2022-03-17
Mention regex will now be restricted to ASCII

MiguelX413:
This attempts to fix https://github.com/instaloader/instaloader/issues/1455 by restricting mentions to only being ASCII.


aandergr:
Thanks!

PR #1464, opened 2022-03-23
Add ujson support

MiguelX413:
This commit adds support for [ujson](https://pypi.org/project/ujson/) if it is available.


aandergr:
I am not convinced of the benefit of this change. Instaloader's time spent for parsing JSONs is marginal.

Issue #1422, opened 2022-02-24
Missing comments collected due 429 error


pvita:
**Describe the bug**
I'm triyng to collect almost 5k comments from a single post (I have the shortcode).
After almost 2 hours (3000, 35000 i read in the CMD) I'm encountering 429 error repeatedly, so I can't to anything else. I can only stop the collection. I tried to wait a bit, but the collection doesn't restart.

Anyway, when I stop the collection (ctrl+c), i lose also the comments collected yet (3000, 3500). The folder is empty.

**To Reproduce**
instaloader --login XXX --comments --no-pictures --no-videos --no-captions --no-metadata-json --max-connection-attempts 0 -- **-shortcode**

**Expected behavior**
I would save the comments collected before the error, at least.

**Instaloader version**
4.8.4

**Additional context**
I tried also with different profiles logged in.

aandergr:
Does this problem still occur with version 4.9?

Also, is there any shortcode where it occurs reliably?

pvita:
> Does this problem still occur with version 4.9?
> 
> Also, is there any shortcode where it occurs reliably?

I didn't try with the new update. I will let you know.
If you want a test shortcode try this one: CaAGxjMqI45

pvita:
> Does this problem still occur with version 4.9?
> 
> Also, is there any shortcode where it occurs reliably?

The problem stlil occurres. When you stop the collection (ctrl+c after 429) it doesn't close the file, losing all the already collected comments.

aandergr:
> When you stop the collection (ctrl+c after 429) it doesn't close the file, losing all the already collected comments.

The current state of downloaded comments is saved with ctrl+c or with AbortDownloadException together with a resume file. This works fine for me, I do not understand the bug that you are reporting.

PR #1465, opened 2022-03-23
added fix for infinity loading next query in nodeiterator if query_re…

JeremyKj87:
…sponse has next_page but _data['edges'] is size 0

  - Fixes  #1347 as a better/improved solution (and probably also #1422).

  - What problem does the pull request solve?
The last fix in #1347 was a partially great solution unless instagram did change the response logic for some queries. 
Currently they return empty "edges arrays" despite the presence of "next_page: true" in query response content.

This leads to an endless loop in the nodeiterator and coherently to the 429 error (this is probably also a part of the error in #1422 ).

You can reproduce this behavior by saving posts with comments from: 

https://www.instagram.com/laura_isabell.gntm22.official/

(particulary this post:

https://www.instagram.com/p/CbFSldpriVy/

will lead to an endless loop respectively to the 429 error)

  - Is it just a proof of concept?
  No.

  - Do you consider it ready to be merged or is it a draft?
It is considered ready to be merged.

### DIFF
--- a/instaloader/nodeiterator.py
+++ b/instaloader/nodeiterator.py
@@ -133,7 +133,7 @@ class NodeIterator(Iterator[T]):
             return item
         if self._data['page_info']['has_next_page']:
             query_response = self._query(self._data['page_info']['end_cursor'])
-            if self._data['edges'] != query_response['edges']:
+            if self._data['edges'] != query_response['edges'] and len(query_response['edges']) > 0:
                 page_index, data = self._page_index, self._data
                 try:
                     self._page_index = 0


aandergr:
> You can reproduce this behavior by saving posts with comments from:
> 
> https://www.instagram.com/laura_isabell.gntm22.official/
> 
> (particulary this post:
> 
> https://www.instagram.com/p/CbFSldpriVy/
> 
> will lead to an endless loop respectively to the 429 error)

I could not reproduce that behavior. Downloading the comments of that particular post works fine for me, without the here-presented change.

Instagram will usually not return blank pages. However, it does happen from time to time that responses from Instagram have errors due to irregularities on Instagram's part. Since this pull request simply catches such an error without causing any foreseeable side effects, we can merge it.

Issue #1471, opened 2022-04-01
TypeError: normalize() argument 2 must be str, not None


queensoft666:
**Describe the bug**
Retrieve images for a speciific location.
First 10 images downloaded OK, then error.

**To Reproduce**
`instaloader.exe %114304211920174 --login xxxxxxxxxxxxx -p xxxxxxxxxxxxxxx --no-compress-json -C --no-video-thumbnails`

**Expected behavior**
Retrieve images for a speciific location

**Error messages and tracebacks**```

[ 10] %114304211920174\2022-04-01_16-40-56_UTC.jpg [😘] comments json
[ 11] Traceback (most recent call last):
  File "__main__.py", line 505, in <module>
  File "__main__.py", line 502, in __main
  File "__main__.py", line 471, in main
  File "__main__.py", line 170, in _main
  File "instaloader\instaloader.py", line 81, in call
  File "instaloader\instaloader.py", line 1156, in download_location
  File "instaloader\instaloader.py", line 1040, in posts_download_loop
  File "instaloader\instaloader.py", line 750, in download_post
  File "string.py", line 161, in format
  File "string.py", line 165, in vformat
  File "string.py", line 205, in _vformat
  File "string.py", line 270, in get_field
  File "instaloader\instaloader.py", line 124, in get_value
  File "instaloader\structures.py", line 371, in caption
TypeError: normalize() argument 2 must be str, not None
[1928] Failed to execute script '__main__' due to unhandled exception!
```

**Instaloader version**
4.9

queensoft666:
I have downloaded latest relese (4.9), which is older ( Mar 26, 2022) than this fix - bug is still present.
I have downloaded the source code (a few days ago), compiled for Windows - bug is still present.
Am I doing something wrong?
`Traceback (most recent call last):
  File "__main__.py", line 505, in <module>
  File "__main__.py", line 502, in __main
  File "__main__.py", line 471, in main
  File "__main__.py", line 170, in _main
  File "instaloader\instaloader.py", line 81, in call
  File "instaloader\instaloader.py", line 1156, in download_location
  File "instaloader\instaloader.py", line 1040, in posts_download_loop
  File "instaloader\instaloader.py", line 750, in download_post
  File "string.py", line 161, in format
  File "string.py", line 165, in vformat
  File "string.py", line 205, in _vformat
  File "string.py", line 270, in get_field
  File "instaloader\instaloader.py", line 124, in get_value
[...]

PR #1475, opened 2022-04-04
Checks if caption is not None before normalization

MiguelX413:
This attempts to fix https://github.com/instaloader/instaloader/issues/1471 by correcting an oversight in https://github.com/instaloader/instaloader/pull/1453.


aandergr:
Thanks!

Issue #1480, opened 2022-04-08
Write output into log file


youngtrashbag:
**Provide us a use case of the feature**
When trying to scrape an Account with many Posts, the rate limit is exceeded, so I am trying to run this as a cronjob.
However in doing so, I am only able to see the output after the program has exited.

**Describe the solution you'd like**
There should be a new parameter when calling instaloader  e.g. `--log="/var/log/instaloader.log"`.
The exact name of the parameter is not relevant, `--log` is the first best example I thought of.

Instaloader should then append its output to that file, instead of writing it into `stdout`, maybe it can even do both.

**Describe alternatives you've considered**
I tried piping it into a log file, but after manual testing I found out that it will only write the output when the process exited.

```bash
python3 -m instaloader
	--login "${USERNAME}" --password "${PASSWORD}" \
	username \
	>> "/var/log/instaloader.log"
```

**If the feature request is accepted, would you be willing to submit a pull request?**
Yes, I would be happy to contribute!

aandergr:
> **Describe alternatives you've considered**
> I tried piping it into a log file, but after manual testing I found out that it will only write the output when the process exited.
> 
> ```shell
> python3 -m instaloader
> 	--login "${USERNAME}" --password "${PASSWORD}" \
> 	username \
> 	>> "/var/log/instaloader.log"
> ```

It works fine with [`script`](https://man7.org/linux/man-pages/man1/script.1.html), like
```shell
script -af /var/log/instaloader.log -c "instaloader --login ${USERNAME} username"
```

Hence, I do not see any need to add that feature to Instaloader.

youngtrashbag:
I want to run it in the background, which I managed to do with `nohup instaloader.sh &`, but it will then create a file called `nohup` and write the output into that file. I can then interactively listen with `tail -f nohup`, and exit to do something else when I want to.

However I would want to send that output directly into a file I specified, which, as far as I know, is not possible with `nohup`.
Trying this with `script` and appending `&` to run it in the background blocks the shell.

victoryy2003:
Is there an equivalent way to use on Windows?

PR #1492, opened 2022-04-17
add --output parameter

youngtrashbag:
As mentioned in Issue #1480, I added a `--output` option, which will write to a file, instead of `stdout`.

I created a de-facto interface, from which the most important methods are inherited.

There are three Writer-classes:
- `DefaultWriter`
    Prints to `stdout`
- `FileWriter`
    append to a file, regardless if it exists or not (but user is warned that file exists if it does)
- `QuietWriter`
    does nothing at all

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -15,6 +15,7 @@ from . import (AbortDownloadException, BadCredentialsException, Instaloader, Ins
 from .instaloader import (get_default_session_filename, get_default_stamps_filename)
 from .instaloadercontext import default_user_agent
 from .lateststamps import LatestStamps
+from .util.output import StandardWriter, FileWriter, QuietWriter
 
 
 def usage_string():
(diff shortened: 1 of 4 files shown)


youngtrashbag:
I could only select master as the base branch, and not e.g. `upcoming/v4.10` as is suggested in [`contributing.md`](https://github.com/instaloader/instaloader/blob/218d88fc6d7e981676e5ce0c0ea47ffd99ee27fb/docs/contributing.rst#writing-code-or-improving-the-documentation)

youngtrashbag:
I chose `master` as base branch, because I could not select `upcoming/v4.10`, as suggested in [`contributing.md`](https://github.com/instaloader/instaloader/blob/218d88fc6d7e981676e5ce0c0ea47ffd99ee27fb/docs/contributing.rst#writing-code-or-improving-the-documentation)

Issue #1494, opened 2022-04-17
FileNotFoundError when --latest-stamps is used and there's no config directory yet


canh:
**Describe the bug**
Program dies with FileNotFoundError when the config/instaloader directory doesn't exist and --latest-stamps is used. 

**To Reproduce**
Run this on a fresh system after installing instaloader:

```
root@26f683a6238b:/# instaloader --login=<REDACTED> --latest-stamps --fast-update apple
Using latest stamps from /root/.config/instaloader/latest-stamps.ini.
Session file does not exist yet - Logging in.
Enter Instagram password for <REDACTED>:
Logged in as <REDACTED>.
Traceback (most recent call last):
  File "/usr/local/bin/instaloader", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/site-packages/instaloader/__main__.py", line 471, in main
    _main(loader,
  File "/usr/local/lib/python3.10/site-packages/instaloader/__main__.py", line 182, in _main
    profile = instaloader.check_profile_id(target, latest_stamps)
  File "/usr/local/lib/python3.10/site-packages/instaloader/instaloader.py", line 1357, in check_profile_id
    latest_stamps.save_profile_id(profile.username, profile.userid)
  File "/usr/local/lib/python3.10/site-packages/instaloader/lateststamps.py", line 46, in save_profile_id
    self._save()
  File "/usr/local/lib/python3.10/site-packages/instaloader/lateststamps.py", line 28, in _save
    with open(self.file, 'w') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/root/.config/instaloader/latest-stamps.ini'
```

**Expected behavior**
The config directory should have been created and instaloader should have kept running.

**Error messages and tracebacks**
If applicable, add error messages and tracebacks to help explain your problem.

**Instaloader version**
4.9

ekalin:
Duplicate of #1467.

canh:
fyi, I fixed this on https://github.com/canh/instaloader/commit/f88f965272b9f3a42ebe98b416692e76ecc4537f

I'll make it a PR once upcoming/v4.10 branch is created.

PR #1500, opened 2022-04-21
Make latest-stamps create intermediate directories

### LINKED ISSUE #1467: Application doesn't create lastest-stamps.ini on its own
**Describe the bug**
Application doesn't create lastest-stamps.ini on it's own

**To Reproduce**
```
@echo off
cd %~dp0
instaloader.exe (about 35 accounts to download) --no-compress-json --stories --latest-stamps --login xyz
```

**Expected behavior**
In my opinion app should create AppData\Local\Instaloader\latest-stamps.ini on its own.

Screenshot_1

**Instaloader version**
I checked v4.8.4 & v4.9

**Additional context**
After creating the directory and file manually everything works fine.
Running as administrator has the same effect.

canh:
before writing the file

Fixes #1467 and #1494

### DIFF
--- a/instaloader/lateststamps.py
+++ b/instaloader/lateststamps.py
@@ -1,6 +1,8 @@
 import configparser
 from datetime import datetime, timezone
 from typing import Optional
+from os.path import dirname
+from os import makedirs
 
 
 class LatestStamps:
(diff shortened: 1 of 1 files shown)


Thammus:
Thanks!

PR #1507, opened 2022-04-25
Update code formatting to be cleaner and closer to 3.6+ standards.

MiguelX413:
This pull request aims to update the writing of the code by updating string formatting to f-strings, removing u-strings, removing unnecessary parentheses, and updating the way in which variables are typed.
These commits were made in consideration of Python 3.6 and newer, as that is the oldest version advertised by the project.

### DIFF
--- a/deploy/windows/create_exe.py
+++ b/deploy/windows/create_exe.py
@@ -51,7 +51,7 @@ commands = ["pip install pipenv==2022.1.8",
 for command in commands:
     print()
     print('#' * (len(command) + 6))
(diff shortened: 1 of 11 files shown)


MiguelX413:
I believe the quotes style should be normalized to primarily use either single or double quotes throughout all files.

MiguelX413:
Should I do this for 4.10 instead?

MiguelX413:
Replaced by https://github.com/instaloader/instaloader/pull/1513

PR #1513, opened 2022-04-27
Update code formatting to be cleaner and closer to 3.6+ standards.

MiguelX413:
Replacing https://github.com/instaloader/instaloader/pull/1507

This pull request aims to update the writing of the code by updating string formatting to f-strings, removing u-strings, removing unnecessary parentheses, and updating the way in which variables are typed.
These commits were made in consideration of Python 3.6 and newer, as that is the oldest version advertised by the project.

### DIFF
--- a/deploy/windows/create_exe.py
+++ b/deploy/windows/create_exe.py
@@ -51,7 +51,7 @@ commands = ["pip install pipenv==2022.1.8",
 for command in commands:
     print()
     print('#' * (len(command) + 6))
-    print('## {} ##'.format(command))
+    print(f'## {command} ##')
     print('#' * (len(command) + 6))
     print(flush=True)
     err = subprocess.Popen(command).wait()
@@ -65,7 +65,7 @@ with open('dist/instaloader.exe', 'rb') as f:
         hash_md5.update(chunk)
 
 with open('dist/instaloader.exe.md5', 'w+') as f:
-    f.write('{} *instaloader.exe\n'.format(hash_md5.hexdigest()))
+    f.write(f'{hash_md5.hexdigest()} *instaloader.exe\n')
 
 # Create ZIP file
 shutil.make_archive('instaloader-{}-windows-standalone'.format(os.getenv('VERSION_TAG')), 'zip', 'dist')
--- a/deploy/windows/create_exe.py
+++ b/deploy/windows/create_exe.py
@@ -68,4 +68,4 @@ with open('dist/instaloader.exe.md5', 'w+') as f:
     f.write(f'{hash_md5.hexdigest()} *instaloader.exe\n')
 
 # Create ZIP file
-shutil.make_archive('instaloader-{}-windows-standalone'.format(os.getenv('VERSION_TAG')), 'zip', 'dist')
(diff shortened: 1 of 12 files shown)


aandergr:
Thanks for going through the code and changing it in order to make it cleaner and better comprehensible!

Please leave out `docs/sphinx_autodoc_typehints.py`, as this file is copied from another project and changing it makes it harder to keep it up to date with upstream when necessary.

I noticed that despite your changes, several strings are still built with `.format()`. Is there a rationale behind when to use f-strings and when to use `.format()`?

MiguelX413:
> Thanks for going through the code and changing it in order to make it cleaner and better comprehensible!
> 
> Please leave out `docs/sphinx_autodoc_typehints.py`, as this file is copied from another project and changing it makes it harder to keep it up to date with upstream when necessary.

Got it!

> 
> I noticed that despite your changes, several strings are still built with `.format()`. Is there a rationale behind when to use f-strings and when to use `.format()`?

I felt that in cases like https://github.com/instaloader/instaloader/blob/e4ba49c2be429780a7a371046d3376e6cee5de65/instaloader/__main__.py#L23-L28 or when methods are being used within, it would probably decrease readability more to change some strings to f-strings.

MiguelX413:
Also, I would like to propose that https://github.com/psf/black be used for this project.

MiguelX413:
I would also like to reformat the namedtuples to use Typing NamedTuple in the new class format
But I'm not entirely sure what the types of the attributes should be

aandergr:
I have also thought about Black, and find it practical in principle to automate the formatting of the code by a software. But the problem is that we can't introduce such a change, because it makes it very difficult for downstream users who use a modified Instaloader version to merge our updates and merge them with their modifications. Unfortunately, the same is true for reformatting the strings, as the code is changed in very many places.

MiguelX413:
> I have also thought about Black, and find it practical in principle to automate the formatting of the code by a software. But the problem is that we can't introduce such a change, because it makes it very difficult for downstream users who use a modified Instaloader version to merge our updates and merge them with their modifications. Unfortunately, the same is true for reformatting the strings, as the code is changed in very many places.

Do you mean the change to f-strings etc too?

MiguelX413:
Do you mind PostComment being made a subclass of PostCommentAnswer in https://github.com/instaloader/instaloader/commit/6ceac282553d738bf7509dcc71c894b46813adb3?
Also, do you have any idea why the docs wont build for it? @aandergr

MiguelX413:
What do you think of the types in https://github.com/instaloader/instaloader/commit/943666b48f4d2446e1120863eb78c55d81de250b? @aandergr

MiguelX413:
Why is namedtuple a func instead of class in the docs btw?

MiguelX413:
What do you think of the types in https://github.com/instaloader/instaloader/commit/0c08405e5a71fbf8125233e5652fff30ccf93d00? @aandergr

MiguelX413:
Do you think we could incorporate any of the changes in this PR on their own if not all of them? @aandergr

MiguelX413:
I moved a few things into #1610.

Issue #1530, opened 2022-05-15
"X has changed its name to X"


NicoWeio:
**Describe the bug**
For me, this appears together with #1523, but it might be a more general issue:
Instaloader reports that a user has changed their name, while in the output, the old and new names are identical.

**To Reproduce**
- Extract a session from Firefox
- `instaloader --login=[…] --stories --fast-update […]`

**Expected behavior**
A clear and concise description of what you expected to happen (even if it seems obvious).

**Error messages and tracebacks**
One relevant line: `JSON Query to […]/feed/: Could not find "window._sharedData" in html response.`

<details>
<summary>Full output</summary>

```
Loaded session from /home/nicolai/.config/instaloader/session-[…].
Logged in as […].
JSON Query to […]/feed/: Could not find "window._sharedData" in html response. [retrying; skip with ^C]
JSON Query to […]/feed/: Could not find "window._sharedData" in html response. [retrying; skip with ^C]
Trying to find profile […] using its unique ID […].
Profile […] has changed its name to […].
JSON Query to […]/feed/: Could not find "window._sharedData" in html response. [retrying; skip with ^C]
JSON Query to […]/feed/: Could not find "window._sharedData" in html response. [retrying; skip with ^C]
Profile […] does not exist.
The most similar profiles are: […].
Trying again anonymously, helps in case you are just blocked.
Warning: […] will be downloaded anonymously ("Profile […] does not exist.
The most similar profiles are: […].").
Downloading anonymously: […]
[1/1] Downloading profile […]
[…]/2020-05-02_10-12-06_UTC_profile_pic.jpg already exists
Retrieving posts from profile […].
[ 1/40] […]/2022-05-13_05-55-51_UTC.jpg [[…]] json 
[ 2/40] […]/2022-01-09_18-04-40_UTC.jpg exists [[…]] unchanged json 
Saved session to /home/nicolai/.config/instaloader/session-[…].

Errors or warnings occurred:
Profile […] has changed its name to […].
Warning: […] will be downloaded anonymously ("Profile […] does not exist.
The most similar profiles are: […].").
```

</details>

**Instaloader version**
4.9

Showmaster3000:
Is it possible that the profile you download blocked you?

NicoWeio:
I don't think so. The profile loads fine in the Instagram app, with the username unchanged.

Showmaster3000:
[...]

PR #1536, opened 2022-05-19
Fix incorrect error messages when certain requests fail

NicoWeio:
This PR aims to fix the symptoms described in #1530, where error messages contradict themselves. It does _not_ fix the underlying problem (“Could not find "window._sharedData" in html response.“), though. Still, I think this is worth it, since profile loading might break in a similar fashion some time in the future.

I'm new to this codebase, so I do lack the “bigger picture”.

---

### a) handle case where `Profile.from_id` works, but `Profile.from_username` doesn't

https://github.com/instaloader/instaloader/blob/385c6c8a353458b8e1f5d1eacd98a714a41e9fc0/instaloader/instaloader.py#L1326-L1337

These are the local variables of interest:
```py
profile_id == 1234
profile == None
profile_name == 'somename'
newname == 'somename'
```
And here's what happens:
- `Profile.from_username` raises `ProfileNotExistsException`
- It's caught afterwards
- `profile_id` is read from the file
- `profile is None` since it wasn't found earlier
- > Trying to find profile somename using its unique ID 1234.
- `Profile.from_id` succeeds
- **Instaloader falsely concludes that the profile has changed its name, since it could not be found using the old name, but using the ID. In reality, `Profile.from_username` failed for some other reason, and the username remained unchanged.**

---

### b) Handle case where `QueryReturnedNotFoundException` lies and `similar_profiles` contains the exact username

https://github.com/instaloader/instaloader/blob/385c6c8a353458b8e1f5d1eacd98a714a41e9fc0/instaloader/structures.py#L764-L772

> Profile somename does not exist.
> The most similar profile is: somename.

Here, Instaloader tries to help by suggesting similar usernames, but doesn't notice that it's actually the one we wanted. It just wasn't able to load it, but since we catch a `QueryReturnedNotFoundException`, it's plausible to assume that the profile really does not exist.

The exception (in my case) is thrown here:
https://github.com/instaloader/instaloader/blob/385c6c8a353458b8e1f5d1eacd98a714a41e9fc0/instaloader/instaloadercontext.py#L356
An alternative approach would be to raise another kind of exception there. 🤷🏼‍♂️

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -1335,7 +1335,8 @@ class Instaloader:
                 profile_from_id = Profile.from_id(self.context, profile_id)
                 newname = profile_from_id.username
                 if profile_name == newname:
-                    self.context.error(f"Warning: Profile {profile_name} could not be retrieved by its name, but by its ID.")
+                    self.context.error(
+                        f"Warning: Profile {profile_name} could not be retrieved by its name, but by its ID.")
                     return profile_from_id
                 self.context.error("Profile {0} has changed its name to {1}.".format(profile_name, newname))
                 if latest_stamps is None:
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -1334,6 +1334,9 @@ class Instaloader:
                                                                                                   profile_id))
                 profile_from_id = Profile.from_id(self.context, profile_id)
                 newname = profile_from_id.username
+                if profile_name == newname:
+                    self.context.error("Profile {0} could not be retrieved by its name, but by its ID.".format(profile_name))
+                    return profile_from_id
                 self.context.error("Profile {0} has changed its name to {1}.".format(profile_name, newname))
                 if latest_stamps is None:
                     if ((format_string_contains_key(self.dirname_pattern, 'profile') or
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -1335,7 +1335,7 @@ class Instaloader:
                 profile_from_id = Profile.from_id(self.context, profile_id)
                 newname = profile_from_id.username
                 if profile_name == newname:
-                    self.context.error("Profile {0} could not be retrieved by its name, but by its ID.".format(profile_name))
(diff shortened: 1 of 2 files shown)


instaloader/structures.py
```diff
@@ -771,6 +771,8 @@ def _obtain_metadata(self):
             top_search_results = TopSearchResults(self._context, self.username)
             similar_profiles = [profile.username for profile in top_search_results.get_profiles()]
             if similar_profiles:
+                if self.username in similar_profiles:
+                    raise ProfileNotExistsException("Profile {} seems to exist, but could not be loaded.".format(self.username))
```
aandergr, line 775:
```suggestion
                    raise ProfileNotExistsException(f"Profile {self.username} seems to exist, but could not be loaded.")
```

instaloader/instaloader.py
```diff
@@ -1334,6 +1334,9 @@ def check_profile_id(self, profile_name: str, latest_stamps: Optional[LatestStam
                                                                                                   profile_id))
                 profile_from_id = Profile.from_id(self.context, profile_id)
                 newname = profile_from_id.username
+                if profile_name == newname:
+                    self.context.error("Profile {0} could not be retrieved by its name, but by its ID.".format(profile_name))
```
aandergr, line 1338:
```suggestion
                    self.context.error(f"Warning: Profile {profile_name} could not be retrieved by its name, but by its ID.")
```

- prefix the output with "Warning:", since the reported problem does not have any impact on the final result
- new f"" format string

aandergr (commented):
Thanks for sharing your findings and proposing a better way to report to the user! This looks fine, and I have only two minor change suggestions.

NicoWeio:
Thanks for your review!

NicoWeio:
Oops, seems like I didn't lint myself. I used autopep8 to conform to the maximum line length of 120.
Just let me know if you'd like to have this rebased before merging. :)

aandergr:
Thanks!

PR #1540, opened 2022-05-23
Fix mention and hashtag regexes and normalize profile biographies

MiguelX413:
This pull request fixes the totally incorrect regex for hashtags and also makes sure profile biographies are normalized.
The hashtag regex was totally incorrect in both the maximum length of 30 and inclusion of periods.
Profile biographies were not being normalized until now.

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -382,9 +382,8 @@ class Post:
         """List of all lowercased hashtags (without preceeding #) that occur in the Post's caption."""
         if not self.caption:
             return []
-        # This regular expression is from jStassen, adjusted to use Python's \w to support Unicode
-        # http://blog.jstassen.com/2016/03/code-regex-for-instagram-username-and-hashtags/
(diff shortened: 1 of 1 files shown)


instaloader/structures.py
```diff
@@ -394,9 +393,10 @@ def caption_mentions(self) -> List[str]:
             return []
         # This regular expression is modified from jStassen, adjusted to use Python's \w to
         # support Unicode and a word/beginning of string delimiter at the beginning to ensure
-        # that no email addresses join the list of mentions.
+        # that no email addresses join the list of mentions, and adjusted to ensure usernames
+        # only consisting of numbers are excluded.
```
aandergr, line 397: `@433` should match `433`, as that seems to be a legal username (that user exists).

aandergr, line 397: `@433` should match `433`, as that seems to be a legal username (that user exists).

MiguelX413:
I believe this and $1541 are ready now @aandergr

MiguelX413:
Should I squash these myself?

aandergr:
Closing, as I prefer #1541 over this.

Issue #1526, opened 2022-05-08
#อนิเมะ isn't detected properly.


MiguelX413:
**Describe the bug**
Because #อนิเมะ has combining characters that can't be normalized into composed characters, the current regex fails to capture the hashtag entirely.

**To Reproduce**
Apply the regex to a caption with the hashtag.

**Expected behavior**
'อนิเมะ' should be in the hashtags of the post.

**Instaloader version**
`4.9`

aandergr:
Thanks for reporting!

Similar to #1446 in that it refers to a problem with the regex for hashtags when certain characters are involved.

MiguelX413:
> Thanks for reporting!
> 
> Similar to #1446 in that it refers to a problem with the regex for hashtags when certain characters are involved.

Indeed, it wasn't fixed by #1453 because this hashtag contains conbining characters with no composite equivalent @aandergr

PR #1542, opened 2022-05-23
Feature/better regex

MiguelX413:
This is based on https://github.com/instaloader/instaloader/pull/1541.
Add support for [regex](https://github.com/mrabarnett/mrab-regex) as an optional dependency to improve hashtag Unicode support.
Documentation not adjusted accordingly yet. Can't pass checks yet.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -14,6 +14,8 @@ psutil = "*"
 typing-extensions = "*"
 types-requests = "*"
 typed-ast = "*"
+regex = "*"
+types-regex = "*"
 
 [packages]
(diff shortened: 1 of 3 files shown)


MiguelX413:
This would solve both https://github.com/instaloader/instaloader/issues/1446 and https://github.com/instaloader/instaloader/issues/1526.

Nasipuk (commented):
#5141

MiguelX413:
Could we do something along these lines in the future? @aandergr

Issue #1379, opened 2022-01-15
post-filter datetime - Filter evaluation failed


kygger:
----------------------------------------------------------------------------------------------------------------------------------

**Describe the bug**

windows console execution file(instaloader.exe) arguments
--post-filter="date_utc < datetime(2022,1,14)"

Filter evaluation failed: can't compare offset-naive and offset-aware datetimes

it's not possible to compare datetime
but version 4.8.2 is possible.

----------------------------------------------------------------------------------------------------------------------------------

**To Reproduce**

instaloader.exe romi.fit --dirname-pattern "folder" --fast-update -l "id"
-f "x:\xxx\session-id" --post-filter="date_utc < datetime(2022,1,14)" --no-video-thumbnails --no-captions --no-metadata-json --no-compress-json --filename-pattern "{date_utc} {owner_username} {shortcode}"

Only download posts with property "date_utc < datetime(2022,1,14)".
Loaded session from x:\xxx\session-id.
Logged in as udjoa10.
Stored ID 1078396896 for profile romi.fit.
[1/1] Downloading profile romi.fit
romi.fit 피트 신새롬 275\romi.fit_2022-01-06_09-30-40_UTC_profile_pic.jpg
Retrieving posts from profile romi.fit.
[  1/354] <Post CYu27w2P3GP> skipped. Filter evaluation failed: can't compare offset-naive and offset-aware datetimes
[  2/354] <Post CYsMWiGvgq6> skipped. Filter evaluation failed: can't compare offset-naive and offset-aware datetimes

----------------------------------------------------------------------------------------------------------------------------------

**Expected behavior**

I dont't get it.

----------------------------------------------------------------------------------------------------------------------------------

**Error messages and tracebacks**

no trackbacks

----------------------------------------------------------------------------------------------------------------------------------

**Instaloader version**
4.8.3

----------------------------------------------------------------------------------------------------------------------------------

aandergr:
Thanks for reporting! This is fixed in version 4.8.2.

PR #1546, opened 2022-05-27
[Post.date_utc]: return utc timezone aware datetime object

anshulsri12:
Please refer: https://github.com/python/cpython/issues/85290

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -4,7 +4,7 @@ import re
 from base64 import b64decode, b64encode
 from collections import namedtuple
 from contextlib import suppress
-from datetime import datetime
+from datetime import datetime, timezone
 from itertools import islice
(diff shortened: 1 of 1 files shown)


anshulsri12:
From datetime library's docs:

>**Warning**
Because naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing a specific timestamp in UTC is by calling datetime.fromtimestamp(timestamp, tz=timezone.utc).

MiguelX413 (approved):
Looks good to me!

aandergr:
Changing the datetimes to offset-aware is not backwards compatible since offset-aware and offset-naive datetimes cannot be compared (#1379).

PR #1548, opened 2022-05-28
Make docs explicit about importing multiple argument files

ziemek99:
Just a minor addition to the docs, stating explicitly that you can import more than one argument file.


aandergr:
Looks good, thanks!

PR #1558, opened 2022-06-02
Handle StoryImage in StoryItem.url

fireattack:
A further fix about issue mentioned in #1238.

After @aandergr 's commit https://github.com/instaloader/instaloader/commit/c0e5d0475bff1507c930d4ceefb5d66686caee5d, a single story can be properly handled; however, the original issue still somewhat exists: higher quality image from iPhone API won't be fetched if a single Story is feeded in and then `.download_storyitem` is called, due to its type being "StoryImage" instead of "GraphStoryImage". 

This fixes it.
Thanks!


aandergr:
Thanks!

Issue #787, opened 2020-08-31
Rate Limiting when downloading media?


itisyeetimetoday:
Does Instaloader automatically rate limit when scraping post and stories? I know that it auto rate limits when scraping followers, but I'm not sure if it automatically rate limits when scraping media.

MrC0D3:
Yes it does. Check https://github.com/instaloader/instaloader/issues/774

Instagram also has its own rate limits.

Issue #774, opened 2020-08-15
429 error on first post retrieval from logged in account, successive attempts also fail


MrC0D3:
**Describe the bug**
429 error on first or second attempt at post retrieval from logged in account, even after waiting several minutes or hours between requests. No 429 error when logged out of the account, but account login is necessary to download media from private profiles and for high quality images.

**To Reproduce**
instaloader <target> --no-captions --no-profile-pic --no-metadata-json --no-video-thumbnails --login=xxx --post-filter="date_utc >= datetime(2020, 1, 1)"

Same 429 error if I use my local IP address or switch to a VPN for the account.

**Expected behavior**
Successful download of media that meets criteria

**Error messages and tracebacks**
This error was displayed on the first run from the logged in account. 
```
Loaded session from /Users/xxx/.config/instaloader/session-xxx.
Logged in as xxx.
Stored ID xxx for profile xxx.
Retrieving posts from profile xxx.
[  1/286] <target>/2020-08-15_xxx_UTC.mp4 
[  2/286] JSON Query to api/v1/media/xxx/info/: 429 Too Many Requests [retrying; skip with ^C]
HTTP error code 429 was returned because too many queries occurred in
the last time. Please do not use Instagram in your browser or run
multiple instances of Instaloader in parallel.
The request will be retried in 666 seconds, at 11:54.
GraphQL requests:
  xxx
    last 10 minutes: 1 requests
    last 11 minutes: 1 requests
    last 15 minutes: 1 requests
    last 20 minutes: 1 requests
    last 30 minutes: 1 requests
    last 60 minutes: 1 requests
  other
    last 10 minutes: 1 requests
    last 11 minutes: 1 requests
    last 15 minutes: 1 requests
    last 20 minutes: 1 requests
    last 30 minutes: 1 requests
    last 60 minutes: 1 requests
  iphone
    last 10 minutes: 1 requests
    last 11 minutes: 1 requests
    last 15 minutes: 1 requests
    last 20 minutes: 1 requests
    last 30 minutes: 1 requests
    last 60 minutes: 1 requests
^C[skipped by user]
JSON Query to api/v1/media/xxx/info/: 429 Too Many Requests Unable to fetch high quality image version of <Post xxx>.
```

**Instaloader version**
$ instaloader --version
4.5

**Additional context**

[...]

PR #1581, opened 2022-06-13
Allow construction of instaloader with rate controller from command line

cecileae:
Many users have complained in existing issues such as #787, #774, etc for rate limiting functionality. I've added usage of the preexisting RateController class for users using instaloader as a standalone command line tool.

The documentation is updated and it is ready to be merged.


cecileae:
Functionality already exists, closing

Issue #1584, opened 2022-06-15
Skip pinned posts


hardestus:
When "--fast-update" is used and the first post is already downloaded, the process stops. Sometimes, the first post is not the most recent one since instagram allows about three posts to be pinned.

The downloading process should continue, even if the pinned posts are downloaded, until it finds an already-downloaded regular post.

Otherwise, you could miss some posts thinking you are up to date. In that case, you will need to omit "--fast-update" option and download all posts whenever you think you missed some.

Thanks

nzec:
Related to #1582

Issue #1582, opened 2022-06-14
Make the ordering of the returned iterator of 'Profile.get_posts()' be chronological


1155159410:
About a week ago, Instagram announced a new function, _'Pin to profile'_. (reference: https://twitter.com/instagram/status/1534227704802729985)

Before this function was announced, the returned iterator of `Profile.get_posts()` is always in chronological order (from most recent to oldest). But now, if there are pinned posts in the profile, the iterator will no longer always be in chronological order.

Chronological ordering is helpful when we want to update the downloaded profile. We can stop iterating when the post's date is earlier than the latest downloaded one. For example:
```
latest_date_local = datetime.fromisoformat("2022-06-11T18:09:22+08:00")  # assume this datetime is the datetime of newest downloaded post
posts = instaloader.Profile(instaloader_.context, "username").get_posts()
for post in posts:
    if post.date_local <= latest_date_local:
        break  # stop iterating
    else:
        ... # download the post
```
But the above implementation way does not work now since the iterator may not be in chronological order (when there are pinned posts in the profile).

There is an easy way to order the `Post` nodes chronologically, but the overhead is really large. When there are many posts in the profile, it costs lots of time, and easily causes `'Too Many Requests'` errors. Here is the implementation way:
```
posts = instaloader.Profile(instaloader_.context, "username").get_posts()
posts = sorted(posts, key=lambda _: _.date_local, reverse=True)
```
Therefore, I would like to ask, if there is any better way to solve the _posts pinning problem_, that causes the returned iterator of `Profile.get_posts()` not be in chronological order.

1155159410:
> Is there anything in the post json that indicates it's a pinned post? If so, pinned posts could be ignored in the `post.date_local <= latest_date_local` comparison.

I have tried to use `get_json_structure()` to inspect the json of `Profile` and `Post`, to find sth like `num_of_pinned_posts` (profile class) or `is_pinned` (post class). But seems there is no such related key.

May I suggest adding _class property_ such as, `pinned_count: int` in `Profile` class or `is_pinned: bool` in `Post` class if feasible.

hardestus:
> Is there anything in the post json that indicates it's a pinned post? If so, pinned posts could be ignored in the `post.date_local <= latest_date_local` comparison.

Yes. "pinned_for_users" contains the user's details if the post is pinned and it is empty if not.

1155159410:
[...]

PR #1586, opened 2022-06-15
Handle pinned posts with --fast-update and --latest-stamps

ekalin:
IG now allows posts to be pinned, and they are return first by `Profile.get_posts()`. This makes `--fast-update` and `--latest-stamp` stop too early (as the first post is considered already downloaded) and posts newer than the pinned post are not downloaded. See #1582 and #1584.

- **The completeness of this change**: It's complete, hopefully
- **Is it just a proof of concept?**: No
- **Is the documentation updated (if appropriate)?** I think no changes are necessary
- **Do you consider it ready to be merged or is it a draft?**: Implementation is complete, but further tests would be useful
- **Can we help you at some point?** Yes

The ideal solution would be to return pinned posts in their proper chronological order. This would require storing pinned posts in a "pending" list and add some logic in the `__next__` function to return either a real post or one of the pinned posts according to the post time. But this would require saving (and restoring) a lot more state when the process is interrupted.

Instead I'm thinking of just treating pinned posts differently:
* With `--fast-update`, pinned posts are not considered when deciding whether to stop downloading. They'll be skipped because they already exist, but download will only stop when reaching a regular post that has been downloaded.
* Similarly, with `--latest-stamps`, pinned posts will be skipped if older than the latest saved post, but downloads continue and only stop when a regular post that is older than the last saved post is reached.

### DIFF
--- a/instaloader/nodeiterator.py
+++ b/instaloader/nodeiterator.py
@@ -129,7 +129,9 @@ class NodeIterator(Iterator[T]):
                 raise
             item = self._node_wrapper(node)
             if self._first_node is None:
-                self._first_node = node
+                pinned = 'pinned_for_users' in node and bool(node['pinned_for_users'])
+                if not pinned:
+                    self._first_node = node
             return item
         if self._data['page_info']['has_next_page']:
             query_response = self._query(self._data['page_info']['end_cursor'])
--- a/instaloader/nodeiterator.py
+++ b/instaloader/nodeiterator.py
@@ -179,7 +179,7 @@ class NodeIterator(Iterator[T]):
         function as the `is_first` parameter when creating the class.
 
         .. versionadded:: 4.8
-        .. versionchanged:: 4.10
+        .. versionchanged:: 4.9.2
            What is considered the first item can be overridden.
         """
         return self._node_wrapper(self._first_node) if self._first_node is not None else None
--- a/instaloader/nodeiterator.py
+++ b/instaloader/nodeiterator.py
@@ -76,7 +76,8 @@ class NodeIterator(Iterator[T]):
                  node_wrapper: Callable[[Dict], T],
                  query_variables: Optional[Dict[str, Any]] = None,
                  query_referer: Optional[str] = None,
-                 first_data: Optional[Dict[str, Any]] = None):
+                 first_data: Optional[Dict[str, Any]] = None,
+                 is_first: Optional[Callable[[T], bool]] = None):
         self._context = context
         self._query_hash = query_hash
         self._edge_extractor = edge_extractor
@@ -91,6 +92,7 @@ class NodeIterator(Iterator[T]):
         else:
             self._data = self._query()
         self._first_node: Optional[Dict] = None
+        self._is_first = is_first
 
     def _query(self, after: Optional[str] = None) -> Dict:
         pagination_variables = {'first': NodeIterator._graphql_page_length}  # type: Dict[str, Any]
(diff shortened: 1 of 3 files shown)


ekalin:
On further reflection, saving a list of pending posts (actually their json structures) is basically what's included in a frozen iterator, so it would increase the size of the saved iterators, but it should not require extra logic for saving/restoring, nor should it translate to more queries when resuming. But it would be nice to have confirmation from those that understand instaloader better than me.

I gues `NodeIterator.__next__` could be changed to return the pinned posts in chronological order with the following algorithm:
1. Get what IG considers the next post (what `__next__` currently does).
2. If it's a pinned post, store it in a `pending` list, and go back to 1.
3. If it's not a pinned post, and there are no pending posts, return it normally.
4. Otherwise, compare it to the newest pending post. If it's newer, return it normally.
5. If a pending post is newer, than return the pending post (removing it from the list of pending posts), and store the post obtained in 1 in the pending list.
6. If 1 returned no more posts, start returning posts from the pending list until there are no more pending posts, only then iteration ends.

The list of pending posts would be saved and restored when freezing/thawing the iterator.

Comments?

ekalin:
Naturally if there's a way (some query param, maybe?) to make IG return the posts in strict chronological order, ignoring pinning status, that would be the easiest way to solve the problem. But I'm not counting on it.

ekalin:
The changes ended up being quite simple.

For `--fast-update`, pinned posts don't cause downloads to stop. The pinned post will be checked every run but will be skipped because they're already downloaded, and downloads will stop once the first non-pinned already downloaded post is reached.

For `--latest-stamps`, pinned posts are downloaded if newer than the saved timestamp, otherwise completely ignored. Download stops once a non-pinned post older than the saved timestamp is reached. When updating the timestamp, the newest non-pinned post is considered, pinned posts are not considered.

One situation could lead to redownloads with `--latest-stamps`: if between two runs an user adds only a pinned post, the saved timestamp is not updated, and so on subsequent runs the pinned post will be downloaded again until a non-pinned post is added. But no posts should be missed and incorrectly skipped.

I believe the implementation is complete, but I didn't have many users with pinned posts to test.

instaloader/nodeiterator.py
```diff
@@ -129,7 +129,9 @@ def __next__(self) -> T:
                 raise
             item = self._node_wrapper(node)
             if self._first_node is None:
-                self._first_node = node
+                pinned = 'pinned_for_users' in node and bool(node['pinned_for_users'])
```
aandergr, line 132: I would prefer that this condition becomes a parameter of the class, such that the NodeIterator implementation does not need to know anything about the structure it is yielding.

ekalin: I introduced a callback as a parameter to the class constructor, if passed this callback is used to determine what constitutes the first item.

As a bonus, the problem that could lead to redownloads (described in https://github.com/instaloader/instaloader/pull/1586#issuecomment-1157674263) was fixed: instead of skipping pinned posts and storing the time of the first non-pinned post, the callback now can identify the newest post in the feed, pinned or not, even if it's not the first that instagram returns.

instaloader/nodeiterator.py
```diff
@@ -168,7 +174,13 @@ def first_item(self) -> Optional[T]:
         """
         If this iterator has produced any items, returns the first item produced.
 
+        It is possible to override what is considered the first item (for example, to consider the
+        newest item in case items are not in strict chronological order) by passing a callback
+        function as the `is_first` parameter when creating the class.
+
         .. versionadded:: 4.8
+        .. versionchanged:: 4.10
```
aandergr, line 182:
```suggestion
        .. versionchanged:: 4.9.2
```

instaloader/structures.py
```diff
@@ -644,6 +644,13 @@ def location(self) -> Optional[PostLocation]:
                                       loc.get('lat'), loc.get('lng'))
         return self._location
 
+    @property
+    def is_pinned(self) -> bool:
+        """True if this Post has been pinned by at least one user.
+
+        .. versionadded: 4.10"""
```
aandergr, line 651:
```suggestion
        .. versionadded: 4.9.2"""
```

aandergr (commented):
Thanks!

Issue #1146, opened 2021-05-21
Story Filter for reposted stories


MrPrajwalB:
As we know, people can repost someone's stories to their own stories they are tagged in it.  Is it possible to filter out those stories which are reshared like this while downloading stories using instaloader? Also, I would like some details on the storyitem.typename object as what it means? (GraphStoryImage vs GraphStoryVideo). Thank you.

PR #1609, opened 2022-07-02
Added `is_reposted`, `original_poster_profile` in `StoryItem`

StamKaly:
A lot of times the "owner" is reposting a story he was tagged into. I am calling the person who tagged the "owner" as "original_poster".

The GraphQL query from such a reposted story has information inside the `tappable_objects` key of each `StoryItem`.
For example for a reposted story it looks like this:

```
"tappable_objects": [
                        {
                            "__typename": "GraphTappableMention",
                            "x": 0.49967163378806606,
                            "y": 0.5150924938267181,
                            "width": 0.9216036185381431,
                            "height": 0.9216036185381431,
                            "rotation": 0.0,
                            "custom_title": null,
                            "attribution": null,
                            "username": "username",
                            "full_name": "User Name",
                            "is_private": false
                        }
                    ]
```

And for a story that was NOT reposted there is nothing inside `tappable_objects` with the `__typename` of `GraphTappableMention`.

The functions I added are pretty self explanatory and I followed the structure the rest of the project has.

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -1292,7 +1292,7 @@ class StoryItem:
             url_candidates.sort()
             return url_candidates[-1][1]
         return None
-    
+
     @property
     def is_reposted(self) -> bool:
         """True if the StoryItem is reposted by other profile."""
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -1123,7 +1123,7 @@ class StoryItem:
         self._context = context
         self._node = node
         self._owner_profile = owner_profile
-	self._original_poster_username = None
+        self._original_poster_username = None
         self._iphone_struct_ = None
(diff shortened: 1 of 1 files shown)


StamKaly:
Resolves #1146

StamKaly:
It seems like this creates a loop because `GraphTappableMention` also appears with the `owner`'s information on the `original_poster`'s story. So it would flag the "original story" also as reposted.

Looking for a workaround..

aandergr:
Just by the logic of the code, `is_reposted` does not make a lot of sense, since it is equivalent to `original_poster_profile is not None`.

Hence, I would prefer a function that returns all tappable mention profiles.

StamKaly:
That's true, I'll rewrite that function.

PR #1610, opened 2022-07-03
Improve typing syntax and type NamedTuples

MiguelX413:
This PR changes type hints to use the PEP 526 typing syntax and replaces untyped `collections.namedtuple`s with `typing.NamedTuple`s.


aandergr:
Thanks!

Issue #1676, opened 2022-08-03
Getting more frequent 400 error


christoarmani:
I am aware of Instagrams and Instaloaders rate limitations, but I have noticed that I am getting a 400 error more frequently and having to unlock my account more often. I am curious to see if others have been noticing the same thing. 

For context, I use the +args.txt method in the command line to download. Within my file I am calling the following arguments:

```
--login=username001
--password=pasw*rd
--abort-on=400
--abort-on=403
--abort-on=429
--no-metadata-json
--no-profile-pic
--no-caption
--filename-pattern={profile}_{filename}
```

I wonder if I should be using the [session login](https://instaloader.github.io/cli-options.html#cmdoption-sessionfile) or if Instagram has been further cracking down on downloads.

Showmaster3000:
propably duplicate of #1633.
also incomplete bug report!

afterdelight:
you can write it like --abort-on=400,403,429

Issue #1658, opened 2022-07-25
Requesting iphone versions of media seems to be triggering blocks


fbck:
Requesting iphone versions of media seems to be triggering request blocks and account blocks.
Difference in quality between iphone versions is there, but it's sometimes negligible, and sometimes, the original media isn't even reaching up to regular media dimensions, let alone iphone sized ones.
Most immediate solution might be to just not request iphone versions of media, by using --no-iphone option.

Showmaster3000:
and what is your bug?

fleescree:
I also got detected since yesterday, two of my accounts have asked for phone verification and password reset.
`instaloader --sessionfile="$HOME/.Instagram/session" --latest-stamps="$HOME/Instagram/latest-stamps.ini" --user-agent="\"$USER_AGENT\""  --stories --fast-update [user_list] `


```
Downloading stories
Download stories: 400 Bad Request
Saved session to /home/ubuntu/.Instagram/session

Errors or warnings occurred:
400 Bad Request Unable to fetch high quality profile pic.
400 Bad Request Unable to fetch high quality profile pic.
400 Bad Request Unable to fetch high quality profile pic.
Download stories: 400 Bad Request

```

ikuchan122:
Yes, you are absolutely correct. This issue has been occurring since July 15. Several accounts of mine have been temporarily disabled. It used to occur only when Instaloader tried to fetch reposted post/story. However, since three days ago, it occurred whenever Instaloader calls the iphone api. Meanwhile, the web api works fine. 

You may wish to see the discussions in the below thread.
https://github.com/instaloader/instaloader/issues/1633#issuecomment-1186261689
https://github.com/instaloader/instaloader/issues/1633#issuecomment-1193169496

fbck:
> and what is your bug?

This feature requests media in a way that constantly results in blocked accounts, and it shouldn't be the default behavior that requires an extra flag to disable if it's currently broken.

AyluinReymaer:
I actually didn't know that it was due to instaloader retrieving the iphone versions of media.

I'm using the python module of instaloader so I didn't have the `--no-iphone` argument.

However, I've initialized the `Instaloader` class object with `iphone_support=False` (which is set to True by default) according to the docs: https://instaloader.github.io/module/instaloader.html?highlight=iphone_support

I'll report back tomorrow at latest to see if the issue is still ongoing.

AyluinReymaer:
[...]

PR #1681, opened 2022-08-06
Don't use iPhone api when merely comparing profile_pic URL with existing one

fireattack:
### A motivation for this change, e.g.
Reduce iPhone API calls. Helps issues like #1676, #1658, etc.

### The changes proposed in this pull request
Don't try to use the iPhone API when merely comparing the profile pic URL with the local stamp.
    
This saves us some API calls when using `--latest-stamps`.

From what I can see, if you are updating lots of profiles, and most of them have no new posts (so no time wasted on checking posts), it would call this API for checking each profile pic way too fast and will got your account suspended (especially if you're using `--no-sleep`). This isn't needed because all the variant, including the web one, has the same remote URL (sans the query parameters). We only need to fetch from iPhone API when the avatar is actually changed. 

### The completeness of this change
Completed, ready to merge.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -562,7 +562,7 @@ class Instaloader:
         if latest_stamps is None:
             self.download_profilepic(profile)
             return
-        profile_pic_basename = profile.profile_pic_url.split('/')[-1].split('?')[0]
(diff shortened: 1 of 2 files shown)


aandergr:
Looks good, thanks!

PR #1696, opened 2022-08-14
download-limit arg added to limit the downloads

Hassibayub:
Motivation:
The count arg should limit the downloads, but it is not working as intended, it was something I needed in my project, and I once saw people talking about this problem which is still not resolved.

Solution:
I added a new arg "-dl, --download-limit" to limit the download. 

>Documentation is not yet updated, I want to you to check and approve the changes first.

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -79,7 +79,9 @@ def _main(instaloader: Instaloader, targetlist: List[str],
           fast_update: bool = False,
           latest_stamps_file: Optional[str] = None,
           max_count: Optional[int] = None, post_filter_str: Optional[str] = None,
-          storyitem_filter_str: Optional[str] = None) -> None:
+          storyitem_filter_str: Optional[str] = None,
+          download_limit: Optional[int] = None) -> None:
+
     """Download set of profiles, hashtags etc. and handle logging in and session files if desired."""
     # Parse and generate filter function
     post_filter = None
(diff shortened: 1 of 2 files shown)


Hassibayub:
Please write feedback. I am looking forward to it.

ekalin:
Wouldn't it make more sense to fix `--count` than to add another option that does the same thing?

Hassibayub:
The --count doesn't work.

afterdelight:
what does this command do?

Hassibayub:
You can specify the number of downloads, i.e., if you want to download 10 contents (video, pictures, stories), you can specify  `--download-limit 10`, and the app will stop after 10 downloads.

While in `--count 10`, I don't see the app stop after 10 downloads

MustafaBaypara:
how can i use this version?

MustafaBaypara:
how can i use on python ?

Hassibayub:
> how can i use on python ?

`python instaloader.py <you args> --download-limit 10` to limit downloads to 10 only.

MustafaBaypara:
> > how can i use on python ?
> 
> `python instaloader.py <you args> --download-limit 10` to limit downloads to 10 only.

limits not working on python, when i try on terminal it is work

aandergr:
> Motivation: The count arg should limit the downloads, but it is not working as intended,

The `--count` argument applies to #hashtag, %location_id, :feed, and :saved, as described in the documentation. It is working as intended.

> Solution: I added a new arg "-dl, --download-limit" to limit the download.

It is quite trivial for the user to only download a certain count of posts of a profile, e.g. by calling download_post() for each post of an islice()'d profile.get_posts(). There is no need to include that in the CLI.

afterdelight:
> > how can i use on python ?
> 
> `python instaloader.py <you args> --download-limit 10` to limit downloads to 10 only.

ok, thank you. this PR is very useful for me since instagram randomly lock accounts when downloading many medias.

MustafaBaypara:
@Hassibayub hey what happened to "-dl"

MustafaBaypara:
ok i get back with ur pull requests ty

dagolinuxoid:
> Motivation: The count arg should limit the downloads, but it is not working as intended, it was something I needed in my project, and I once saw people talking about this problem which is still not resolved.
> 
> Solution: I added a new arg "-dl, --download-limit" to limit the download.
> 
> > Documentation is not yet updated, I want to you to check and approve the changes first.

Actually it's a good idea. I was looking for the option as well. Honestly it would be better to have the option included into instaloader.

Issue #1706, opened 2022-08-19
command line: instaloader new version update


bachig26:
**Provide us a use case of the feature**
How could the user invoke the new function?
- _`instaloader.exe -U` or `instaloader.exe --update`_

Which problem would it solve?
- _this function is to solve the instaloader update via command line easily instead of downloading the new version from the GitHub every time._

If new information is obtained, how would it be further processed?
- _for example, this function is available in the [yt-dlp](https://github.com/yt-dlp/yt-dlp) {**Self-updater**: The releases can be updated using `yt-dlp -U`}, which I use every time to update and easy for the user._

**If the feature request is accepted, would you be willing to submit a pull request?**
No, I can't do programming. 
possibly add `help wanted` label if this feature request pr needs to be done by someone other than you.

bachig26:
@aandergr @Thammus @Hassibayub I want to ask if this is possible and what do you think about this?

Hassibayub:
You are good at thinking out of the box. I'll check on this in my free time today 👍👍

Hassibayub:
I'll possibly complete this request tomorrow. @bachig26

bachig26:
> I'll possibly complete this request tomorrow. @bachig26 

Thank you. Looking forward to try it once got the release.

Hassibayub:
Feature added and pull request created. It's the basic implementation that just works. I will add wait animation and stylize terminal output in future, meanwhile I am occupied with my piles of projects.

Thank you @bachig26 for suggesting this update.

aandergr:
Running `pip3 install --upgrade instaloader` is in no way more complicated than something like `instaloader --upgrade` would be. Hence, I do not see a benefit in this feature suggestion.

PR #1714, opened 2022-08-26
Arg -U --update added for update app

Hassibayub:
# Motivation
Resolve feature requested at #1706

It's the basic implementation and will add wait animation and terminal output recreation in the future.  

The documentation md file not updated


lx30011:
Has no error handling

Hassibayub:
I don't think it requires any error handling; we run the subprocess, which returns the error if the internet connection isn't established.

aandergr:
This simply calls pip to update Instaloader and as such does not add any benefit. The user can instead directly call pip.

It also causes trouble if Instaloader was installed by other means than pip.

PR #1717, opened 2022-08-28
Fixes incorrect latest stamp when download is resumed (#1701)

### LINKED ISSUE #1701: Incorrect latest stamp saved when download is interrupted and resumed
**Describe the bug**
When `--latest-stamps` is used and the download is interrupted and then resumed, the saved post timestamp is not that of the newest post in the profile, but of the newest post downloaded in the latest run.

**To Reproduce**
1. `instaloader --latest-stamps some-profile`
2. Interrupt the download with Control-C.
3. Re-run instaloader and let it finish downloading the profile.
4. Check latest stamps file, it'll have the timestamp of the newest post of the second run.

**Expected behavior**
The saved timestamp should be that of the newest post of the whole profile.

**Instaloader version**
4.9.3, but should happen since 4.9.2.

**Additional context**
This happens because of the new check for pinned posts. When the `NodeIterator` is created it is passed a closure that stores the newest date seen and uses this date to determine if a post is the newest (because with pinned posts the first post is not necessarily the newest). But this date is not saved between runs.

ekalin:
**Motivation**: Fixes #1701.

Not only this fixes the problem, but it makes the code simpler. However, the type of the `is_first` callback has been changed. But sine this is a new feature (and very specific), I don't think this change will be a problem.

**Do you consider it ready to be merged or is it a draft?** Ready
**Can we help you at some point?** This is a simple change and should be complete, but I'm open to suggestions

### DIFF
--- a/instaloader/nodeiterator.py
+++ b/instaloader/nodeiterator.py
@@ -92,7 +92,6 @@ class NodeIterator(Iterator[T]):
         else:
             self._data = self._query()
         self._first_node: Optional[Dict] = None
-        self._first_item: Optional[T] = None
         self._is_first = is_first
 
     def _query(self, after: Optional[str] = None) -> Dict:
@@ -132,13 +131,11 @@ class NodeIterator(Iterator[T]):
                 raise
             item = self._node_wrapper(node)
             if self._is_first is not None:
-                if self._is_first(item, self._first_item):
+                if self._is_first(item, self.first_item):
                     self._first_node = node
-                    self._first_item = item
             else:
                 if self._first_node is None:
                     self._first_node = node
-                    self._first_item = item
             return item
         if self._data['page_info']['has_next_page']:
             query_response = self._query(self._data['page_info']['end_cursor'])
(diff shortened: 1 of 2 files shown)


instaloader/nodeiterator.py
```diff
@@ -131,11 +132,13 @@ def __next__(self) -> T:
                 raise
             item = self._node_wrapper(node)
             if self._is_first is not None:
-                if self._is_first(item):
+                if self._is_first(item, self._first_item):
```
aandergr, line 135: I would prefer to pass `self._node_wrapper(self._first_node)` here as the second parameter, rather than storing `_first_item` besides `_first_node`. The node wrapper functions are all pure and quick, hence I think we should prefer the solution that requires less state managing.

ekalin: Fair point. And it makes the code simpler, since there's already a function for converting the `_first_node` with `_node_wrapper`, with proper `None`-checking.

ekalin:
Hmm... works for me here. The error looks like one file has been updated, but the other is still at the previous verson. Could you double check that the working copy is 100% up-to-date?

If it is, I guess I'll need the command line, it's probably some combination of parameters that I haven't tried.

afterdelight:
What is the file name did you mean? Okay, ill try to update to latest version first

ekalin:
The error looks like `nodeiterator.py` has been updated, but `structures.py` has not. Which would be very unusual.

afterdelight:
yeah i replaced the file manually from your branch to an older version of insta loader

aandergr:
Thanks!

Issue #1728, opened 2022-09-06
Highlights make the script retrieve errors


Muggeren:
**Describe the bug**
It retrieves an error

**To Reproduce**
"./instaloader --login=USER --highlights --no-video-thumbnails --no-posts TARGET"

**Expected behavior**
To download all highlights from target

**Error messages and tracebacks**

[  5/ 49] Warning: Unable to fetch video from graphql of <StoryItem 2771602286366031741>: list index out of range
Unable to fetch high-quality video version of <StoryItem 2771602286366031741>: 'video_versions'
Traceback (most recent call last):
  File "__main__.py", line 505, in <module>
  File "__main__.py", line 502, in __main
  File "__main__.py", line 471, in main
  File "__main__.py", line 220, in _main
  File "instaloader\instaloader.py", line 1456, in download_profiles
  File "instaloader\instaloader.py", line 81, in call
  File "instaloader\instaloader.py", line 960, in download_highlights
  File "instaloader\instaloader.py", line 887, in download_storyitem
  File "instaloader\instaloader.py", line 97, in call
  File "instaloader\instaloader.py", line 340, in download_pic
  File "re.py", line 201, in search
TypeError: expected string or bytes-like object
[23852] Failed to execute script '__main__' due to unhandled exception!

**Instaloader version**
4.9.3

**Additional context**
Everytime i use the +args feature, the script seems to crash too.

Showmaster3000:
Duplicate of #1718.
And same thing like in #1460. But this issue is closed and looks unfixed...

fireattack:
So here is how it happens.

In the highlight (https://www.instagram.com/stories/highlights/17962767154067076/), there is one story (fifth last, the one said "sideline work" that is a `StoryItem` with `.is_video == True`, but it doesn't actually have a video_url in neither graphql and iPhone API. You can see that if you open it in browser too, it just has a still image, no video or whatsoever.

We currently don't handle this case well: we assume we will always at least got the video_url from graphql. `StoryItem.video_url` *can* return `None` without erroring out, but `Instaloader.download_pic()` don't properly handle `url=None`).

I'm trying to come up with a fix, but not exactly sure what is the best way to do it.

Muggeren:
Ah okay. Thanks for answering the post, and the knowledge  (:

aandergr:
Fixed in version 4.9.4.

Issue #1718, opened 2022-08-30
Weird bug when running


Muggeren:
**Describe the bug**
After running for like 20 min, this occurs:

"[  6/ 10] Warning: Unable to fetch video from graphql of <StoryItem 2675921249948319648>: list index out of range
Unable to fetch high-quality video version of <StoryItem 2675921249948319648>: 'video_versions'
Traceback (most recent call last):
  File "__main__.py", line 505, in <module>
  File "__main__.py", line 502, in __main
  File "__main__.py", line 471, in main
  File "__main__.py", line 220, in _main
  File "instaloader\instaloader.py", line 1456, in download_profiles
  File "instaloader\instaloader.py", line 81, in call
  File "instaloader\instaloader.py", line 960, in download_highlights
  File "instaloader\instaloader.py", line 887, in download_storyitem
  File "instaloader\instaloader.py", line 97, in call
  File "instaloader\instaloader.py", line 340, in download_pic
  File "re.py", line 201, in search
TypeError: expected string or bytes-like object
[17556] Failed to execute script '__main__' due to unhandled exception!"

**To Reproduce**
happens when i use this
--login=username
--password=password
--no-metadata-json
--no-captions
--no-video-thumbnails
--tagged
--highlights
PROFILE
PROFILE

**Expected behavior**
To download all requested from above

**Error messages and tracebacks**
Just did

**Instaloader version**
v4.9.3
-->

Showmaster3000:
potentially same thing like in #1460. But this issue is closed and looks unfixed?

aandergr:
Duplicate of #1460

IljaM13:
As workaround: run with --highlights and --fast-update
This will skip already downloaded highlights an won't run into exception.
Maybe it will crash with one of the following highlights. Just retry until you gott them all. :-)

PR #1732, opened 2022-09-07
download_pic: return if url is empty

fireattack:
### A motivation for this change, e.g.
  - Fixes # 1460, #1718, #1728, etc.

### What problem does the pull request solve?
 
When parsing very old `StoryItem` from highlights, there is a chance that despite the metadata says `.is_video == True`, it actually does not have video URL in neither graphql nor iPhone API (server side issue, you can even see that with official web). 

This will cause `StoryItem.video_url` to return `None` and `download_pic` does not handle that well.

### The changes proposed in this pull requests
There are multiple ways to fix it: 
1) only attempt to call `download_pic` if url is valid; 
2) handle None `url` in `download_pic`. 

I personally think method 2 is better since it will deal with potential similar issues in other classes like `Post`.

Also a note: one may think we can just mark the item as `.is_video = False` if graphql does not return video URL. This is WRONG because as mentioned before, if you fetch story directly by mediaid or shortcode, it will not have `video_resources` URLs, but you can still get it from iPhone API.

### The completeness of this change
  - Is the documentation updated (if appropriate)? N/A
  - Do you consider it ready to be merged or is it a draft? Ready to be merged
  - Can we help you at some point? Yes

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -337,6 +337,8 @@ class Instaloader:
         Returns true, if file was actually downloaded, i.e. updated."""
         if filename_suffix is not None:
             filename += '_' + filename_suffix
+        if not url:
+            return False
         urlmatch = re.search('\\.[a-z0-9]*\\?', url)
         file_extension = url[-3:] if urlmatch is None else urlmatch.group(0)[1:-1]
         nominal_filename = filename + '.' + file_extension


fireattack:
One thing though is that do we want to fallback to download the thumbnail if this happens, even when the user says "--no-video-thumbnails`.

This is the case where just marking item as not video works really well; but we can achieve the same by changing the routine a little bit.

instaloader/instaloader.py
```diff
@@ -337,6 +337,8 @@ def download_pic(self, filename: str, url: str, mtime: datetime,
         Returns true, if file was actually downloaded, i.e. updated."""
         if filename_suffix is not None:
             filename += '_' + filename_suffix
+        if not url:
+            return False
```
aandergr, line 341: I am against adding this, since download_pic should not receive `None` as `url` parameter (and it cannot according to how it is typed).

aandergr (commented):
Thanks for this change!

I think it is a good way to handle a video whose video_url_fetch_failed as a picture (as opposed to a video), and as such, download its URL regardless of whether download_video_thumbnails is set.

instaloader/instaloader.py
fireattack: Agreed. It's no longer needed

PR #1765, opened 2022-10-08
Fixed login

### LINKED ISSUE #1758: Login error
Hi,

I got error `KeyError: "name='csrftoken', domain=None, path=None"`,  when trying to login

instaloder ver : 4.9.5

IL2C:
- **A motivation for this change, e.g.**

  - Fixes #1758 

- **The changes proposed in this pull request**

  - Use a user agent that belongs to Instagram software _(versions don't need updating)_.

- **The completeness of this change**

  - Is the documentation updated (if appropriate)? N/A
  - Do you consider it ready to be merged or is it a draft? Ready to merge

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -32,8 +32,9 @@ def copy_session(session: requests.Session, request_timeout: Optional[float] = N
 
 
 def default_user_agent() -> str:
-    return 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 ' \
-           '(KHTML, like Gecko) Chrome/92.0.4515.131 Safari/537.36'
+    return 'Mozilla/5.0 (Linux; Android 13) AppleWebKit/537.36 ' \
+            '(KHTML, like Gecko) Chrome/106.0.5249.79 Mobile Safari/537.36 ' \
+            'Instagram 255.1.0.17.102'
 
 
 class InstaloaderContext:


movAX13h:
If you hardcode the user agent like this, IG will trigger the phishing alert if the browser you logged in with has a different UA. To avoid that you can use the user agent of the browser you logged in with + the instagram part `Instagram 255.1.0.17.102`. I recommend passing the user agent to the `instaloader.Instaloader()` call so you can easily change it if required.

IL2C:
> If you hardcode the user agent like this, IG will trigger the phishing alert if the browser you logged in with has a different UA. To avoid that you can use the user agent of the browser you logged in with + the instagram part `Instagram 255.1.0.17.102`. I recommend passing the user agent to the `instaloader.Instaloader()` call so you can easily change it if required.

1. I have yet to have my account blocked that's logged in through a browser and the app. We'd need more feedback on this.
2. The user agent must use Android or iPhone followed by the Instagram version to access the `csrftoken` cookie on login.
3. This PR doesn't make it any more difficult in changing the user agent.

sadegh:
I applied this change locally and it fixed the error.

movAX13h:
It works temporarily ... but runs into the same issues after some time.

davebiffuk:
This patch on top of instaloader 4.9.1 works fine for me, fixing the `KeyError: "name='csrftoken', domain=None, path=None"` issue with typical usage like this: `instaloader --login MYACCOUNT --no-iphone ACCOUNTTODOWNLOAD`

koenvanderveen:
this fixed it for me as well 👍

Thammus:
For me it feels somehow wrong to use a user agent of the app to access web endpoints and to mimic the behavior of the browser. When I tried the user agent suggested here and tested it with Instaloader, I was indeed able to log in again, but after downloading about 20 posts, all further requests were answered with a 403 error code and all active sessions on my account were invalidated and I was forced to confirm my identity by phone number the next time I logged in

PR #1771, opened 2022-10-14
FIX Login - user_agent change

Shinoby92:
### fixes login error due wrong user_agent
needs to be iOS user agent to work.


movAX13h:
Did you use iOS when you created the user? 
It seems to me that these custom user agents are not working for everyone.

Thammus:
Given the similarity to #1765, the arguments, experiences, and discussions there probably apply to this suggested changes as well. Also, this PR and #1765 are conflicting, i.e. we can only use one of the two proposed user agents

top31:
> fools

Issue #1773, opened 2022-10-18
File "/usr/local/lib/python3.10/site-packages/requests/cookies.py", line 413, in _find_no_duplicates     raise KeyError(f"name={name!r}, domain={domain!r}, path={path!r}")


Drjacky:
```bash
Traceback (most recent call last):
  File "/usr/local/bin/instaloader", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/site-packages/instaloader/__main__.py", line 471, in main
    _main(loader,
  File "/usr/local/lib/python3.10/site-packages/instaloader/__main__.py", line 121, in _main
    instaloader.interactive_login(username)
  File "/usr/local/lib/python3.10/site-packages/instaloader/instaloader.py", line 1572, in interactive_login
    self.login(username, password)
  File "/usr/local/lib/python3.10/site-packages/instaloader/instaloader.py", line 634, in login
    self.context.login(user, passwd)
  File "/usr/local/lib/python3.10/site-packages/instaloader/instaloadercontext.py", line 265, in login
    session.headers.update({'X-CSRFToken': login.cookies['csrftoken']})
  File "/usr/local/lib/python3.10/site-packages/requests/cookies.py", line 334, in __getitem__
    return self._find_no_duplicates(name)
  File "/usr/local/lib/python3.10/site-packages/requests/cookies.py", line 413, in _find_no_duplicates
    raise KeyError(f"name={name!r}, domain={domain!r}, path={path!r}")
KeyError: "name='csrftoken', domain=None, path=None"
```

Version: 4.9.5

There was 2FA enabled but when I inserted the code, it said: `2FA error: This code is no longer valid. Please request a new one.`. but it was valid and new. So, I disabled the 2FA and I got the above error.

Drjacky:
I tried with an account without 2FA, and got this:
```bash
Fatal error: Login: Checkpoint required. Point your browser to https://www.instagram.com/challenge/action/AXFOb52balbalblablaban8tQZ/ - follow the instructions, then retry.
```

And when I click on it, it just asks if it was you or not and change your password. That's it.

Drjacky:
I used this fork https://github.com/instaloader/instaloader/issues/1758#issuecomment-1272618465 and it work but after a couple of downloads, it displayed 403. and when I logged in again, I get 400 and now the account is disabled!

HubbaBubba666:
I use instaloader.exe and since today I always get such an error message:

Traceback (most recent call last):
[...]

Issue #1217, opened 2021-07-02
2FA not working


soneji:
```
Session file does not exist yet - Logging in.
Enter Instagram password for _____:
Enter 2FA verification code: XXXXXX
2FA error: This code is no longer valid. Please request a new one.
Enter 2FA verification code:
```

This happens again and again, I even tried setting up a 2FA app, but this did not work (I read the issue and you guys dont support it)

```
instaloader --version
4.7.4
```
Any help would be appreciated, thank you!

neutrongenious:
facing same issue +1

aandergr:
Thanks for reporting.

valdearg:
Anyone find an issue for this? Been happening for a few weeks.

jamesbaber:
Same issue here

lucabinotti:
I have the same issue.

Soneji:
No point spamming, just leave a 👍 on my original post, I'm sure the team are working on it as you can see

> Thanks for reporting.

Soneji:
@aandergr is there any update on this? 16+ people having this issue

aandergr:
We will track all updates related to this issue here, I recommend to use the subscribe or watch feature of github to get informed about updates. 

Until this is fixed, the workaround described in https://instaloader.github.io/troubleshooting.html#login-error can be used.

riccardogabellone:
> 
> 
> We will track all updates related to this issue here, I recommend to use the subscribe or watch feature of github to get informed about updates.
> 
> Until this is fixed, the workaround described in https://instaloader.github.io/troubleshooting.html#login-error can be used.

I have the same issue, but in Android using **Termux**, .py script cannot be used if phone is not rooted (I figured out this after trying to use `-c` parameter and get cookies from `/data/data/org.mozilla.firefox/files/mozilla/xxxxxxxx.default/ `as mentioned [here](https://support.mozilla.org/it/questions/1040912))

Dan1jel:
Don't work for me either...

bitwon:
Same here.

benjaminu:
Same issue here. Tried both cli (`--login`) and code versions; same error in both cases.
```
    loader = instaloader.Instaloader()
    try:
        loader.login(username, password)
    except TwoFactorAuthRequiredException:
        two_factor_code = input("Enter 2FA Code: ")
        loader.two_factor_login(two_factor_code)
```
```
[...]

PR #1784, opened 2022-10-26
Fixed bug that was asking for 2FA code again and again.

### LINKED ISSUE #1565: Cannot login using the 2FA
**Describe the bug**

**To Reproduce**
1. type the command: instaloader --login [your_username]
2. give the account's password3. 
3. give the code you receive on your mobile device (or in case, generated on the authenticator app)
4. you get the following error: 2FA error: This code is no longer valid. Please request a new one5. 
5. if you try again with a new code you get: 2FA error: feedback_required

**Expected behavior**
after giving the 2FA code, I expect to be authenticated by Instagram.

1. This code is no longer valid. Please request a new one5. 2. 
2. 2FA error: feedback_required3. 

**Instaloader version**
4.9

sakisdog:
This PR fixes the bug that caused the 2FA code to be asked again and again even though it was typed correctly. Fixes #1565, #1773, #1758, #1565, #1217 and many more.
No need to import cookies from any browser now.

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -208,6 +208,8 @@ class InstaloaderContext:
                                 'ig_vw': '1920', 'ig_cb': '1', 'csrftoken': '',
                                 's_network': '', 'ds_user_id': ''})
         session.headers.update(self._default_http_header())
+        #Getting mid cookie value
+        session.get("https://i.instagram.com/api/v1/web/login_page/")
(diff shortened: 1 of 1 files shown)


aandergr:
The 2FA code is no longer requested repeatedly, so it is an improvement, but unfortunately it seems to me that no functioning session is established afterwards.

arunjohnraphi (approved):

PR #1819, opened 2022-12-03
download_profile add max_count param

zhlhuang:
Add param  --count to download_profile,  do not attempt to download more than COUNT posts


ekalin:
How will this be different from the already existing `--count` option?

Thammus:
Like discussed in #1696 (and multiple other issues) we do not like to change or expand the behavior or functionality of the `--count` param.

PR #1825, opened 2022-12-07
fix datetime locale issue

### LINKED ISSUE #1814: ValueError from datetime formatting
**Describe the bug**
When I download a profile, including the profile picture, the program raises a ValueError, complaining about a datetime that supposedly does not match a specified format. Downloading the profile with `--no-profile-pic` works fine.

**To Reproduce**

    $ instaloader <profile>

**Expected behavior**
Download the profile picture and posts without error.

    $ instaloader bbc

    Hint: Use --login to download higher-quality versions of pictures.
    [1/1] Downloading profile bbc
    File "/usr/bin/instaloader", line 33, in <module>
        sys.exit(load_entry_point('instaloader==4.9.5', 'console_scripts', 'instaloader')())
    File "/usr/lib/python3.10/site-packages/instaloader/__main__.py", line 471, in main
        _main(loader,
    File "/usr/lib/python3.10/site-packages/instaloader/__main__.py", line 220, in _main
        instaloader.download_profiles(profiles,
    File "/usr/lib/python3.10/site-packages/instaloader/instaloader.py", line 1427, in download_profiles
        self.download_profilepic_if_new(profile, latest_stamps)
    File "/usr/lib/python3.10/site-packages/instaloader/instaloader.py", line 563, in download_profilepic_if_new
        self.download_profilepic(profile)
    File "/usr/lib/python3.10/site-packages/instaloader/instaloader.py", line 574, in download_profilepic
        self.download_title_pic(profile.profile_pic_url, profile.username.lower(), 'profile_pic', profile)
    File "/usr/lib/python3.10/site-packages/instaloader/instaloader.py", line 97, in call
        return func(instaloader, *args, **kwargs)
    File "/usr/lib/python3.10/site-packages/instaloader/instaloader.py", line 529, in download_title_pic
        date_object = datetime.strptime(http_response.headers["Last-Modified"], '%a, %d %b %Y %H:%M:%S GMT')
    File "/usr/lib/python3.10/_strptime.py", line 568, in _strptime_datetime
        tt, fraction, gmtoff_fraction = _strptime(data_string, format)
    File "/usr/lib/python3.10/_strptime.py", line 349, in _strptime
        raise ValueError("time data %r does not match format %r" %
    ValueError: time data 'Wed, 20 Oct 2021 06:41:49 GMT' does not match format '%a, %d %b %Y %H:%M:%S GMT'

**Instaloader version**
4.9.5

**Additional context**
I am particularly confused about this issue, since the mentioned datetime seems to match that format perfectly.
I am using Manjaro Linux. This issue happens both when installed from the AUR or from pip.

dwy6626:
Fixes #1814

## Cause

The `%a` in `strptime` depends on locale: https://docs.python.org/3.10/library/datetime.html#strftime-and-strptime-format-codes.

In https://github.com/instaloader/instaloader/blob/master/instaloader/instaloadercontext.py#L145, the request header specify the response locale to be `en_US`. If the client locale is not `en_US`, there could be value error in the datetime parsing like #1814. 

Base on [this stackoverflow](https://stackoverflow.com/questions/48270191/), Python depends on the underlying `C` locale. So we cannot use envvar `LC_ALL` to overwrite the time format locale. Therefore, I decide to update the locale in runtime.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -14,6 +14,7 @@ from io import BytesIO
 from pathlib import Path
 from typing import Any, Callable, IO, Iterator, List, Optional, Set, Union, cast
 from urllib.parse import urlparse
+import locale
 
 import requests
 import urllib3  # type: ignore
@@ -526,6 +527,7 @@ class Instaloader:
         http_response = self.context.get_raw(url)
         date_object = None  # type: Optional[datetime]
         if 'Last-Modified' in http_response.headers:
+            locale.setlocale(locale.LC_TIME, 'en_US')  # use the locale specified in the request header for parsing datetime
             date_object = datetime.strptime(http_response.headers["Last-Modified"], '%a, %d %b %Y %H:%M:%S GMT')
             date_object = date_object.replace(tzinfo=timezone.utc)
             pic_bytes = None
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -527,7 +527,8 @@ class Instaloader:
         http_response = self.context.get_raw(url)
         date_object = None  # type: Optional[datetime]
         if 'Last-Modified' in http_response.headers:
-            locale.setlocale(locale.LC_TIME, 'en_US')  # use the locale specified in the request header for parsing datetime
+            # use the locale specified in the request header for parsing datetime
+            locale.setlocale(locale.LC_TIME, 'en_US')
             date_object = datetime.strptime(http_response.headers["Last-Modified"], '%a, %d %b %Y %H:%M:%S GMT')
             date_object = date_object.replace(tzinfo=timezone.utc)
             pic_bytes = None


aandergr:
In https://docs.python.org/3/library/locale.html#background-details-hints-tips-and-caveats the documentation says

> Initially, when a program is started, the locale is the `C` locale, no matter what the user’s preferred locale is.

Hence, strptime should work correctly in parsing the weekday names as required per HTTP protocol, regardless of the language settings.

That article goes on

> It is generally a bad idea to call [setlocale()](https://docs.python.org/3/library/locale.html#locale.setlocale) in some library routine, since as a side effect it affects the entire program. Saving and restoring it is almost as bad: it is expensive and affects other threads that happen to run before the settings have been restored.

dwy6626:
1. When the `C` locale is not `en_US`, python program will use wrong locale and the parsing will fail.
2. agree with you, thanks for point out the bad idea of use setlocale. I will try to come up with another solution and close this PR.

PR #1841, opened 2022-12-22
Add copy proxies to new session

ainyava:
Hi, I just needed to use proxies with this library and say an OpenPR on adding this functionality.

I figured there are some reviews needed to be resolved but minimal support to coping proxies to new session would solve my problem.

I've been setting proxies in my code like:
``` python
L.context._session.proxies = dict(http='[PROXY]', https='[PROXY]')
```  
then I saw the copy_session function used under the hood

I think this quick solution would be nice :+1:


aandergr:
This change, at least in itself, has no function.

Instaloader already supports proxies by setting environment variables. The `sessions.proxies` value is overridden by those same environment variables.

A meaningful proxy implementation would have to have added value over the existing one. It could do this, for example, by not using the proxies set by environment variables. One approach to this would be to pass the proxy parameter in the requests call instead of through the session object.

PR #1879, opened 2023-02-04
Detect redirects to https://i.instagram.com/accounts/login, fixes redirect loop

ekalin:
In some cases (possibly when requests are made to `i.instagram.com`), the redirect to login uses host `i.instagram.com` instead of `api.instagram.com`:

```
HTTP redirect from https://i.instagram.com/api/v1/users/web_profile_info/?username=profile to https://i.instagram.com/accounts/login/?next=/api/v1/users/web_profile_info/
```

This is not detected by instaload as a login request, and the application seems to enter a redirect loop, which results in getting 429 responses.

This change just adds the new URL to the detection of login required, letting the existing logic take over.

**Is it just a proof of concept?**: No
**Do you consider it ready to be merged or is it a draft?** Ready

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -331,7 +331,8 @@ class InstaloaderContext:
             while resp.is_redirect:
                 redirect_url = resp.headers['location']
                 self.log('\nHTTP redirect from https://{0}/{1} to {2}'.format(host, path, redirect_url))
-                if redirect_url.startswith('https://www.instagram.com/accounts/login'):
+                if (redirect_url.startswith('https://www.instagram.com/accounts/login') or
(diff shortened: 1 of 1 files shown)


ekalin:
This happened again, and I was able to confirm that now the redirect is properly identified.

Unfortunately it does not prevent the loop and 429 errors. I've opened #1882 for that issue, which is believe is best treated separately.

Nevertheless, with this new code I only got two 429 errors (and a total wait of 30 minutes) before for whatever reason the requests were allowed to proceed (and eventually fail). Previously it took seven 429's and 90 minutes to fail, but this might just haven been some particularity of Instagram on the different days.

aandergr:
Thank you. That looks like a reasonable change. It's plausible that when redirecting to the i.instagram.com login, the handle_429 logic should also be called if already logged in.

Issue #1871, opened 2023-01-27
Trying to download with location id, but it failed on version 4.9.5


Not4GoodBoyThou:
**Describe the bug**
I was trying to download few posts with location id, and its showing an error. The location id is the same as that of in the unittest
`ID = 362629379`

**To Reproduce**
`instaloader --login=username --sessionfile=sessionpath %362629379 -c 15`

**Expected behavior**
Downloading Posts with the given location id

**Error messages and tracebacks**
Loaded session from sessionpath 
Logged in as username .
Retrieving pictures for location 362629379...
JSON Query to explore/locations/362629379/: Expecting value: line 1 column 1 (char 0) [retrying; skip with ^C]
JSON Query to explore/locations/362629379/: Expecting value: line 1 column 1 (char 0) [retrying; skip with ^C]
%362629379: JSON Query to explore/locations/362629379/: Expecting value: line 1 column 1 (char 0)
Saved session to sessionpath.

Errors or warnings occurred:
%362629379: JSON Query to explore/locations/362629379/: Expecting value: line 1 column 1 (char 0)

**Instaloader version**
`4.9.5`

**Additional context**
The location id is valid
https://www.instagram.com/explore/locations/362629379/

JoaquinBenitez:
Same error is happening to me, I've tried downgrading to other versions, but the error is happening too (on versions 4.9.4, 4.9.3 and 4.9.2)

PR #1881, opened 2023-02-05
Fixed downloading with location id

SayanthD:
* Closes https://github.com/instaloader/instaloader/issues/1871
* Also fixed accessing 'location' of the Post


aandergr:
Fixes both downloading by location ID and geotags. Thanks! :)

PR #1899, opened 2023-02-23
Added property "entities" to Profile class

jamesphoenixcode:
Added a property for more metadata hoarding.

This property first will test if the target profile has entities, if not it will return the string:
`No sponsors list.`

Otherwise it will return a single `string\n` or a list of strings i.e:
`Sponsor1`
`Sponsor2`

This was an idea when i was creating a script for better and pretty organization of the data downloaded using IL lib.
Any advice for better coding, comment or related is appreciate, i am starting.

Edit 1, 2 and 3:
Using
`Python 3.11.2`
`Instaloader 4.9.6`

Here is the repo of that idea: [Repo](https://github.com/jamesphoenixcode/prettyIL)

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -877,6 +877,21 @@ class Profile:
         return self._metadata('biography')
 
     @property
+    def entities(self) -> str:
+        '''This property will return a list or single ``string`` of the sponsored accounts/entities showed in the target profile
+        or ``No sponsors list.`` string.
+        '''
+        # Check if the target has entities.
+        if len(self._metadata('biography_with_entities')["entities"]) == 0:
+            return "No sponsors list."
+        else:
+            # If so, create the list/single string and return it.
+            sponsorsList = ""
+            for item in self._metadata('biography_with_entities')["entities"]:
+                sponsorsList = sponsorsList + f'{item["user"]["username"]}\n'
+            return sponsorsList
+
+    @property
     def blocked_by_viewer(self) -> bool:
         return self._metadata('blocked_by_viewer')
 
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -878,9 +878,7 @@ class Profile:
 
     @property
     def entities(self) -> str:
-        '''This property will return a list or single ``string`` of the sponsored accounts/entities showed in the target profile
-        or ``No sponsors list.`` string.
-        '''
+	'''This return a ``str`` of the sponsored accounts/entities showed in the target profile otherwise ``No sponsors list.`` string.'''
         # Check if the target has entities.
         if len(self._metadata('biography_with_entities')["entities"]) == 0:
             return "No sponsors list."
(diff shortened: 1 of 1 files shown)


instaloader/structures.py
```diff
@@ -876,6 +876,19 @@ def business_category_name(self) -> str:
     def biography(self) -> str:
         return self._metadata('biography')
 
+    @property
+    def entities(self) -> str:
```
aandergr, line 880: Could we make it return a List rather than a `str`? Would be more user-friendly.

aandergr:
- The return type should be typed as `List[...]` rather than `list`
 - The types of the items of the list should preferably be `Profile` instances rather than `str`. The `Profile` class is typically able to handle these profile stubs, similar as in e.g. `get_likes()`.
 - If the list is empty an empty list should be returned, rather than a string
 - It should be based on `upcoming/v4.10` as it adds a new feature, also a `versionadded` directive should be added to the docstring, similar to as it is in other methods.
 - The name of the property should reflect that it has something to do with sponsors in the biography.

jamesphoenixcode:
> * The return type should be typed as `List[...]` rather than `list`
> * The types of the items of the list should preferably be `Profile` instances rather than `str`. The `Profile` class is typically able to handle these profile stubs, similar as in e.g. `get_likes()`.
> * If the list is empty an empty list should be returned, rather than a string
> * It should be based on `upcoming/v4.10` as it adds a new feature, also a `versionadded` directive should be added to the docstring, similar to as it is in other methods.
> * The name of the property should reflect that it has something to do with sponsors in the biography.

What do you think?

aandergr:
Cool, looks a lot better.
- `Profile.from_username` would do an additional request which would be unnecessary in this case, we should use the simpler default Profile constructor directly
- I liked that it previously was a `@property`, and I think it should be one, named e.g. simply `sponsors`. 
- The docstring part `wich you can iterate over` is redundant, since a list can always be iterated over; however, this is of course a bit of a matter of taste, I just like things short like that.
- The `if` is redundant, since if the list was empty, the for loop would also produce an empty list.
- The expression could also be narrowed down into a simple  `return` statement.

So all in all, I could imagine something like this

```python
    @property
    def sponsors(self) -> List['Profile']:
        """
        Return a list of :class:`Profile` sponsors.

        .. versionadded:: 4.10
        """
        return [Profile(self._context, sponsor['user'])
                for sponsor in self._metadata('biography_with_entities')['entities']]
```

instaloader/structures.py
```diff
@@ -876,6 +876,20 @@ def business_category_name(self) -> str:
     def biography(self) -> str:
         return self._metadata('biography')
 
+    def get_sponsors(self) -> List['Profile']:
+        """
+        Return a list of :class:`Profile` sponsors wich you can iterate over.
+
+        .. versionadded:: 4.10
```
aandergr, line 883: We released `4.10`, this should go into `4.11` then.

```suggestion
        .. versionadded:: 4.11
```

PR #1900, opened 2023-02-23
Abort the download when redirected to login

### LINKED ISSUE #1882: Redirect to login when fetching profile info leads to 429 error
**Describe the bug**
For whatever reason, IG is relatively nice and did not temporarily block the account, it is just asking for a new login. Moreover, it let the initial login request succeed, but once one asks for profile data, just then it asks for a login.

This leads to messages such as:

    JSON Query to api/v1/users/web_profile_info/?username=profile: Redirected to login [retrying; skip with ^C]

Unfortunately, it seems that it just retries the query, leading to a 429 error:
```
   d6f4427fbe92d846298cf93df0b937d3:    1    1    1    1    1    1
 *                           iphone:   19   19   19   19   19   19
is running.
The request will be retried in 29 minutes, at 01:45.
```

Eventually, for whatever reason, it got past this requests, and was able to attempt the actual download, which failed with a 403 ("Please log back in") and caused the program to abort (because I set `--abort-on`).

**To Reproduce**
Unfortunately this is very IG dependent. It has happened twice in the last couple days, but who knows what triggers this specific IG behavior?

**Expected behavior**
If it's a simple request to login, then (if login was requested in the command line) instaloader should attempt to login again. If login succeeds, then it should attempt to continue what it was doing before. If it fails, the program should abort.

If login was not requested, then it should abort. I believe this is what happens in this case.

**Instaloader version**
4.9.5

**Additional context**
I'd be willing to submit a PR to fix this, but I'm not sure how to best solve this.

ekalin:
This fixes #1882: instead of trying again, it aborts the process. If it happens during download, it should save the iterator for the next run, but for me this happens when fetching the profiles, before anything was downloaded. (During the download IG instead returns a 403 error instead.)

**Do you consider it ready to be merged or is it a draft?** It's ready
**Can we help you at some point?** I don't see the need, but feel free to suggest another message.

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -335,8 +335,8 @@ class InstaloaderContext:
                     redirect_url.startswith('https://i.instagram.com/accounts/login')):
                     if not self.is_logged_in:
                         raise LoginRequiredException("Redirected to login page. Use --login.")
-                    # alternate rate limit exceeded behavior
-                    raise TooManyRequestsException("Redirected to login")
+                    raise AbortDownloadException("Redirected to login page. You've been logged, please wait " +
+                                                 "some time, recreate the session and try again")
                 if redirect_url.startswith('https://{}/'.format(host)):
                     resp = sess.get(redirect_url if redirect_url.endswith('/') else redirect_url + '/',
                                     params=params, allow_redirects=False)


instaloader/instaloadercontext.py
```diff
@@ -335,8 +335,8 @@ def get_json(self, path: str, params: Dict[str, Any], host: str = 'www.instagram
                     redirect_url.startswith('https://i.instagram.com/accounts/login')):
                     if not self.is_logged_in:
                         raise LoginRequiredException("Redirected to login page. Use --login.")
-                    # alternate rate limit exceeded behavior
-                    raise TooManyRequestsException("Redirected to login")
+                    raise AbortDownloadException("Redirected to login page. You've been logged, please wait " +
```
aandergr, line 338:
```suggestion
                    raise AbortDownloadException("Redirected to login page. You've been logged out, please wait " +
```

Should read "You've been logged **out**", I guess?

aandergr (commented):
This looks good to me, as discussed in #1882. Just found a small typo in the message of the exception.

ekalin:
Sure, it was a silly typo, but has been fixed

aandergr:
Thanks!

PR #1901, opened 2023-02-23
Log the response when an abort status is received

ekalin:
This is a small change that adds the contents of the response to the log message when the download is aborted with one of the error codes specified with the command-line option `--abort-on`.

I find the contents to be useful, sometimes it just says "You've been logged out", sometimes something like "Challenge required". And it would help in the case of other errors as well.

**Do you consider it ready to be merged or is it a draft?** It's ready
**Can we help you at some point?** Not necessary

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -372,8 +372,11 @@ class InstaloaderContext:
             resp = sess.get('https://{0}/{1}'.format(host, path), params=params, allow_redirects=False)
             if resp.status_code in self.fatal_status_codes:
                 redirect = " redirect to {}".format(resp.headers['location']) if 'location' in resp.headers else ""
-                raise AbortDownloadException("Query to https://{}/{} responded with \"{} {}\"{}:{}".format(
-                    host, path, resp.status_code, resp.reason, redirect, resp.text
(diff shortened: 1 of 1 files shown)


ekalin:
An HTML page would be bad, I agree.

Running my version with this change, however, I only get JSON errors, which are much smaller (and useful).

Maybe a check could be added and only output the body if the content type is JSON, but it seems unnecessary.

EricWu91:
What about saving the content in a html _file_ instead of dumping it in the console output?

aandergr:
> What about saving the content in a html _file_ instead of dumping it in the console output?

I think that would be a good and useful solution.

ekalin:
Saving it to a file would probably require a command-line option to specify the destination (and enable the option, as not everyone might want it), and so seems overly complicated.

I've added a check to only include the body it it's a JSON (the exception that might print the body is raised in a method called `get_json`, but I've added an extra check), and it's also truncated it it's too long (I used 500 characters, but this can be changed - in my experience, the longest I've seen is about 300 characters). Hopefully this can be merged.

aandergr:
Yes this looks good and I also think it can be merged. Thanks!

PR #1967, opened 2023-05-07
[new] Added Get Followed Hashtags.

M1ndo:
Retrieve a list of hashtags followed by a given profile since it hasn't be implemented.
Updated documentation and updated test file.

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -1103,6 +1103,7 @@ class Profile:
         To use this, one needs to be logged in and private profiles has to be followed.
 
         :rtype: NodeIterator[Hashtag]
+
         .. versionadded:: 4.10
         """
         if not self._context.is_logged_in:
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -1097,6 +1097,26 @@ class Profile:
     def _make_is_newest_checker() -> Callable[[Post, Optional[Post]], bool]:
         return lambda post, first: first is None or post.date_local > first.date_local
 
+    def get_followed_hashtags(self) -> NodeIterator['Hashtag']:
+        """
+        Retrieve list of hashtags followed by given profile.
+        To use this, one needs to be logged in and private profiles has to be followed.
+
+        :rtype: NodeIterator[Hashtag]
+        .. versionadded:: 4.10
+        """
+        if not self._context.is_logged_in:
+            raise LoginRequiredException("--login required to get a profile's followers.")
+        self._obtain_metadata()
+        return NodeIterator(
+            self._context,
+            'e6306cc3dbe69d6a82ef8b5f8654c50b',
+            lambda d: d["data"]["user"]["edge_following_hashtag"],
+            lambda n: Hashtag(self._context, n),
+            {'id': str(self.userid)},
+            'https://www.instagram.com/{0}/'.format(self.username),
+        )
+
     def get_followers(self) -> NodeIterator['Profile']:
         """
         Retrieve list of followers of given profile.


instaloader/structures.py
```diff
@@ -1061,6 +1061,20 @@ def get_followers(self) -> NodeIterator['Profile']:
...
+        To use this, one needs to be logged in and private profiles has to be followed.
+        """
+        if not self._context.is_logged_in:
+            raise LoginRequiredException("--login required to get a profile's followers.")
+        self._obtain_metadata()
+        yield from (Hashtag(self._context, edge["node"]) for edge in
+            self._context.graphql_query("e6306cc3dbe69d6a82ef8b5f8654c50b",
+                                        {"id": str(self.userid)},
+                                        "https://www.instagram.com/{0}/"
+                                        .format(self.username))["data"]["user"]["edge_following_hashtag"]["edges"])
```
aandergr, line 1076: This should rather return a NodeIterator, which does all the pagination handling and also supports resuming. If it's unclear how to create one, see the other invocations of NodeIterator or feel free to ask.

docs/as-module.rst
```diff
@@ -82,6 +82,9 @@ metadata of a Profile. :class:`Profile` instances can be created with:
 - :meth:`Profile.get_followers`
    Profiles that follow given user.
 
+- :meth:`Profile.get_followed_hashtags`
+   Hashtags that given user has followed.
+
```
aandergr, line 87: This list starts with

> With the `Profile` class, Instaloader also makes it easy to access metadata of a Profile. `Profile` instances can be created with:

However, `Profile.get_followed_hashtags()` returns `Hashtag` instances. It should not be listed in this list.

instaloader/structures.py
```diff
@@ -1061,6 +1061,20 @@ def get_followers(self) -> NodeIterator['Profile']:
             'https://www.instagram.com/{0}/'.format(self.username),
         )
 
+    def get_followed_hashtags(self) -> Iterator['Hashtag']:
+        """
+        Retrive list of hashtags followed by given profile.
+        To use this, one needs to be logged in and private profiles has to be followed.
+        """
```
aandergr, line 1068: We should also base this change on `upcoming/v4.10`, and add a `.. versionadded:: 4.10` directive in this docstring.

aandergr (commented):
Thanks, this looks like a nice feature to me and seems well implemented. I just have a few minor inline comments.

instaloader/structures.py
M1ndo: That would be the case if the returned object supports pagination but its doesn't so it raises *KeyError* It's just 1 big list of nodes that contains a hashtag object.
image

**Still** i haven't came across a testing account with more than >55 followed hashtags but i'm certain it doesn't support pagination unlike returning an object of followers or followings (with 25 items)...

docs/as-module.rst
M1ndo: Removed it, since it will included in `structures.rst`

instaloader/structures.py
aandergr: Indeed, thanks for noting. In `upcoming/v4.10` I just pushed a fix to let the NodeIterator handle single paged responses like this one and it does not raise that KeyError anymore.

M1ndo: Thanks for adding a fix, i have switched to branch upcoming/v4.10 and everything should be resolved now.

instaloader/structures.py
```diff
@@ -1097,6 +1097,26 @@ def get_igtv_posts(self) -> NodeIterator[Post]:
     def _make_is_newest_checker() -> Callable[[Post, Optional[Post]], bool]:
         return lambda post, first: first is None or post.date_local > first.date_local
 
+    def get_followed_hashtags(self) -> NodeIterator['Hashtag']:
+        """
+        Retrieve list of hashtags followed by given profile.
+        To use this, one needs to be logged in and private profiles has to be followed.
+
+        :rtype: NodeIterator[Hashtag]
+        .. versionadded:: 4.10
```
aandergr, line 1106:
```suggestion
        :rtype: NodeIterator[Hashtag]

        .. versionadded:: 4.10
```

Here a blank line is needed.

M1ndo: Should be resolved now.

PR #1977, opened 2023-05-15
Add Post.translated_caption property

SayanthD:

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -452,6 +452,25 @@ class Post:
             return None
 
     @property
+    def translated_caption(self) -> Optional[str]:
+        """Translated caption of the post, if caption is available.
+
+        .. versionadded:: 4.10"""
+        if not self._context.iphone_support:
+            raise IPhoneSupportDisabledException("iPhone support is disabled.")
+        if not self._context.is_logged_in:
+            raise LoginRequiredException("--login required to access iPhone media info endpoint.")
+
+        if not self.caption:
+            return None
+        try:
+            pk = self._iphone_struct['caption']['pk']
+            data = self._context.get_iphone_json(path=f'api/v1/language/bulk_translate/?comment_ids={pk}', params={})
+            return data['comment_translations'][0]['translation']
+        except KeyError:
+            return None
+
+    @property
     def tagged_users(self) -> List[str]:
         """List of all lowercased users that are tagged in the Post."""
         try:
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -408,6 +408,21 @@ class Post:
                     yield PostSidecarNode(is_video=is_video, display_url=display_url,
                                           video_url=node['video_url'] if is_video else None)
 
+    def get_translated_caption(self) -> Optional[str]:
+        """Retrieve translation of the caption.
+
+        .. versionadded:: 4.10"""
+        if self.caption and self._context.iphone_support and self._context.is_logged_in:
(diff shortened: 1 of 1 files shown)


aandergr:
`Post.translated_caption` always makes an API request, even when called multiple times in a row for the same post. This goes somewhat against the intuition of a property. If we stick with it, we should rather call it `Post.get_translated_caption()`.

The translation endpoint seems to simply get passed an ID `pk` and returns the translation? Maybe this could be abstracted and a method could be provided that generally returns the translation for a `pk`, and also stores this in a simple dict in the InstaloaderContext and takes it from this dict when requested repeatedly. Then I would find it also as property `translated_caption` okay.

SayanthD:
@aandergr You meant like rewriting the present caption in the fake_node to dict like `'caption' : {'id': <pk value>, 'text': '...', 'translation': '...'}`

or this, adding a key called 'translation' with `'translation' : {'id': <pk value>, 'text': '...'}` in fake_node ?

And then adding the method to fetch translated text from api with using the 'id'?

aandergr:
> @aandergr You meant like rewriting the present caption in the fake_node to dict like `'caption' : {'id': <pk value>, 'text': '...', 'translation': '...'}`
> 
> or this, adding a key called 'translation' with `'translation' : {'id': <pk value>, 'text': '...'}` in fake_node ?

`fake_node` in the `from_iphone_struct()` classmethod is a translation from the iphone struct to the 'regular' graphql structure needed in the `Post` class. Changing its caption would break the code.

> And then adding the method to fetch translated text from api with using the 'id'?

The id `pk` is already present in the iphone struct (`caption.pk`), so no need to store it additionally somewhere. `translated_caption` would call a method that returns the translation for a `pk`.

SayanthD:
@aandergr Sorry for the delayed response. Please review this

instaloader/structures.py
```diff
@@ -408,6 +408,21 @@ def get_sidecar_nodes(self, start=0, end=-1) -> Iterator[PostSidecarNode]:
...
+        .. versionadded:: 4.10"""
+        if self.caption and self._context.iphone_support and self._context.is_logged_in:
+            try:
+                pk = self._iphone_struct['caption']['pk']
+                data = self._context.get_iphone_json(path=f'api/v1/language/bulk_translate/?comment_ids={pk}',
+                                                    params={})
+                self._node["translated_caption"] = data['comment_translations'][0]['translation']
+                return data['comment_translations'][0]['translation']
+            except (InstaloaderException, KeyError, IndexError) as err:
+                self._context.error(f"Unable to get translation: {err}")
```
aandergr, line 423: `pk` should be generically passed as a parameter, the function should be moved in the context since it can serve more than only the Post, the translation should not be stored in the node, rather the translation strings could be cashed with a dict by `pk`.

instaloader/structures.py
```diff
@@ -408,6 +408,21 @@ def get_sidecar_nodes(self, start=0, end=-1) -> Iterator[PostSidecarNode]:
                     yield PostSidecarNode(is_video=is_video, display_url=display_url,
                                           video_url=node['video_url'] if is_video else None)
 
+    def get_translated_caption(self) -> Optional[str]:
+        """Retrieve translation of the caption.
+
+        .. versionadded:: 4.10"""
```
aandergr, line 414: We released `4.10`, this should go into `4.11` then.

```suggestion
        .. versionadded:: 4.11
```

PR #1998, opened 2023-06-04
[new] Get comment likes.

M1ndo:
Return an a list profiles that liked a comment.

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -45,11 +45,11 @@ PostCommentAnswer.likes_count.__doc__ = "Number of likes on comment."
 
 class PostComment:
     def __init__(self, context: 'InstaloaderContext', node: Dict[str, Any],
-                 answers: Iterator['PostCommentAnswer'], shortcode: str):
+                 answers: Iterator['PostCommentAnswer'], post: 'Post'):
         self._context = context
         self._node = node
         self._answers = answers
-        self._shortcode = shortcode
+        self._post = post
 
     @property
     def id(self) -> int:
@@ -86,7 +86,7 @@ class PostComment:
         """
         Iterate over all likes of a comment. A :class:`Profile` instance of each like is yielded.
 
-        .. versionadded:: 4.10
+        .. versionadded:: 4.11
         """
         if self.likes_count != 0:
             return NodeIterator(
@@ -95,12 +95,12 @@ class PostComment:
                 lambda d: d['data']['comment']['edge_liked_by'],
                 lambda n: Profile(self._context, n),
                 {'comment_id': self.id},
-                'https://www.instagram.com/p/{0}/'.format(self._shortcode),
+                'https://www.instagram.com/p/{0}/'.format(self._post.shortcode),
             )
         return []
 
     def __repr__(self):
-        return f'<PostComment {self.id} of {self._shortcode}>'
+        return f'<PostComment {self.id} of {self._post.shortcode}>'
 
 class PostLocation(NamedTuple):
     id: int
@@ -630,7 +630,7 @@ class Post:
 
         def _postcomment(node):
             return PostComment(context=self._context, node=node,
-                               answers=_postcommentanswers(node), shortcode=self.shortcode)
+                               answers=_postcommentanswers(node), post=self)
         if self.comments == 0:
             # Avoid doing additional requests if there are no comments
             return []
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -44,12 +44,12 @@ PostCommentAnswer.likes_count.__doc__ = "Number of likes on comment."
 
 
 class PostComment:
-    def __init__(self, node: Dict[str, Any], answers: Iterator['PostCommentAnswer'],
-                 context: 'InstaloaderContext', shortcode: str):
+    def __init__(self, context: 'InstaloaderContext', node: Dict[str, Any],
+                 answers: Iterator['PostCommentAnswer'], shortcode: str):
+        self._context = context
         self._node = node
         self._answers = answers
-        self._context = context
-        self.shortcode = shortcode
+        self._shortcode = shortcode
 
     @property
     def id(self) -> int:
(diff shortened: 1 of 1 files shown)


instaloader/structures.py
```diff
@@ -49,6 +49,7 @@ class PostComment(NamedTuple):
     text: str
     owner: 'Profile'
     likes_count: int
+    likes: Iterator['Profile']
```
aandergr, line 52: Docstring for likes is missing.

The docstrings for PostComment are described in currently line 56 - 58 of this file.

instaloader/structures.py
```diff
@@ -583,9 +584,20 @@ def _postcommentanswers(node):
...
+                    lambda d: d['data']['comment']['edge_liked_by'],
+                    lambda n: Profile(self._context, n),
+                    {'comment_id': node['id']},
+                    'https://www.instagram.com/p/{0}/'.format(self.shortcode),
+                )
+
         def _postcomment(node):
             return PostComment(*_postcommentanswer(node),
-                               answers=_postcommentanswers(node))
+                               answers=_postcommentanswers(node), likes=_postcommentlikes(node))
```
aandergr, line 600: This will trigger an additional GraphQL query for every comment that has a like, even if the likes are not accessed by the user.

One way to solve that problem would be to convert PostComment into a real class with a `@property` `likes()` that does the query only if `.likes` is actually called.

M1ndo: Yes totally missed that, thanks for letting me know, Also it's not needed to convert  `PostComment`  into a real class, for `@property` `likes()` to work. But if the code seems awkward or something, i could push a change to convert it into a real class.

aandergr: Indeed I'd rather prefer this PostComment being a real class, considering that a NamedTuple normally simply stores values in a named manner, and our PostComment is now more complex than that. Also it would allow us to 'hide' the `context` as a `_context` like it's done in other structures that need to have the context.

instaloader/structures.py
```diff
@@ -50,7 +50,26 @@ class PostComment(NamedTuple):
...
     answers: Iterator[PostCommentAnswer]
+    context: InstaloaderContext
+    shortcode: str
 
+    @property
+    def likes(self) -> Iterator['Profile']:
+        """
+        Iterate over all likes of a comment. A :class:`Profile` instance of each like is yielded.
+
+        .. versionchanged:: 4.10
```
aandergr, line 61:
```suggestion
        .. versionadded:: 4.10
```

instaloader/structures.py
```diff
@@ -50,7 +50,26 @@ class PostComment(NamedTuple):
...
+        if self.likes_count != 0:
+            return NodeIterator(
+                self.context,
+                '5f0b1f6281e72053cbc07909c8d154ae',
+                lambda d: d['data']['comment']['edge_liked_by'],
+                lambda n: Profile(self.context, n),
+                {'comment_id': self.id},
+                'https://www.instagram.com/p/{0}/'.format(self.shortcode),
+            )
+        return iter([])
```
aandergr, line 72: I think it would be easier to make this function return `Iterable['Profile']` instead and then `return []` here.

instaloader/structures.py
M1ndo: I have pushed a commit changing the class, plus other things. It should be good now.

instaloader/structures.py
```diff
@@ -43,18 +43,68 @@ class PostCommentAnswer(NamedTuple):
...
-    owner: 'Profile'
-    likes_count: int
-    answers: Iterator[PostCommentAnswer]
+class PostComment:
+    def __init__(self, node: Dict[str, Any], answers: Iterator['PostCommentAnswer'],
+                 context: 'InstaloaderContext', shortcode: str):
+        self._node = node
+        self._answers = answers
+        self._context = context
+        self.shortcode = shortcode
```
aandergr, line 52: We should pass a `Post` instance rather than just the shortcode, this seems more future-proof in case we once need to access other fields of the Post.

We should also prefix the field with an underscore, to avoid giving the impression that it can be changed from outside.

instaloader/structures.py
```diff
@@ -43,18 +43,68 @@ class PostCommentAnswer(NamedTuple):
...
 
-for field in PostCommentAnswer._fields:
-    getattr(PostComment, field).__doc__ = getattr(PostCommentAnswer, field).__doc__  # pylint: disable=no-member
-PostComment.answers.__doc__ = r"Iterator which yields all :class:`PostCommentAnswer`\ s for the comment."
+    def __repr__(self):
+        return (f'{self.__class__.__name__}(id={self.id!r}, '
+            f'created_at_utc={self.created_at_utc!r}, '
+            f'text={self.text!r}, '
+            f'owner={self.owner.username!r}, '
+            f'likes_count={self.likes_count!r})')
```
aandergr, line 107: Our `__repr__` style is a lot shorter, I'd suggest something like

`<PostComment {id} of {shortcode}>`

instaloader/structures.py
```diff
@@ -43,18 +43,68 @@ class PostCommentAnswer(NamedTuple):
...
-class PostComment(NamedTuple):
-    id: int
-    created_at_utc: datetime
-    text: str
-    owner: 'Profile'
-    likes_count: int
-    answers: Iterator[PostCommentAnswer]
+class PostComment:
+    def __init__(self, node: Dict[str, Any], answers: Iterator['PostCommentAnswer'],
+                 context: 'InstaloaderContext', shortcode: str):
```
aandergr, line 48: The order of the parameters in the constructor for classes like these is always `context`, `node`, and then additional parameters.
```suggestion
    def __init__(self, context: 'InstaloaderContext', node: Dict[str, Any],
                 answers: Iterator['PostCommentAnswer'], shortcode: str):
```

aandergr (commented):
Nice, thanks! Only a few minor suggestions and then I think this can be merged :)

instaloader/structures.py
aandergr: > I'm not very sure what do you mean by passing `Post` instance? we can pass its properties and such but itself?

Yes, we can pass a Post instance.

```python
            return PostComment(context=self._context, node=node,
                               answers=_postcommentanswers(node), post=self)
```

instaloader/structures.py
```diff
@@ -43,19 +43,64 @@ class PostCommentAnswer(NamedTuple):
...
+    def answers(self) -> Iterator['PostCommentAnswer']:
+        """ Iterator which yields all :class:`PostCommentAnswer` for the comment. """
+        return self._answers
+
+    @property
+    def likes(self) -> Iterable['Profile']:
+        """
+        Iterate over all likes of a comment. A :class:`Profile` instance of each like is yielded.
 
+        .. versionadded:: 4.10
```
aandergr, line 89: We released `4.10`, this should go into `4.11` then.

```suggestion
        .. versionadded:: 4.11
```

instaloader/structures.py
M1ndo: Hello, Sorry I missed the release v4.10. i've been quiet offline for the past weeks.
It should be good to merge now.

aandergr:
Thanks!

PR #2012, opened 2023-06-18
Minor documentation updates

swarnatonse:
Minor updates to "What to Download" section of CLI usage


aandergr:
We do not consider this an improvement.

PR #2040, opened 2023-07-23
Feature/load cookies from browser

### LINKED ISSUE #1801: Add Option to Automatically Get Cookies from Browser So Downloading with Instaloader Can Actually Work
As discussed in issue #1787, Instaloader is often or always now completely failing at downloading  unless cookies are loaded from the browser after first logging into Instagram via browser. So my suggestion is adding an option to Instaloader like

`--load-cookies browser`

wherein you can specify that _browser_ string as `chrome`, `firefox`, etc., or as a path to a specific browser cookies file, to automatically load cookies from that browser so that downloading can actually work.

Note that the YouTube downloader [yt-dlp](https://github.com/yt-dlp/yt-dlp) has a working similar option, `--cookies-from-browser browser`, for just this reason. Similar code could be used here.

JunedKhan101:
# Feature get cookies from browser 🚀
  - Fixes #1801 
  - This PR implementes the feature to get cookies from browser for a user as current login function was frequently breaking there will be no need to login in the first place if we can get the cookies from the browser.

- Changes include a function that gets cookie from browser using browser_cookie3 python library, if cookie is loaded succesfully we set the cookie in session cookies and test if we can login using the test login function.

## Test Cases 1:

The testing include me first deleting the session of the test user to demostrate logging in with --load-cookies.
I provide the username with --login option and specify **chrome** browser with --load-cookies.
I get succesfully logged in as I was in the browser.

## Test Cases 2:

I provide the username with --login option and specify **brave** browser with --load-cookies.
I get a message that says "No cookies found for Instagram in brave, Are you logged in succesfully in brave?" because I was not logged into the brave browser.

Attaching video of the testing:

https://github.com/instaloader/instaloader/assets/47941768/ec402892-8046-4a20-8c0a-ac4b80b0ed23

  - Do you consider it ready to be merged or is it a draft? I consider it ready to be merged if it looks good
  - Can we help you at some point? Not at the time


ekalin:
That is a most welcome feature. However, I believe it would be useful to be able to specify the path to the browser profile from which to load the cookies, in case the user has multiple profiles.

JunedKhan101:
@ekalin Do you mean to specify the whole path to the cookie for a browser profile? as it is specified in the `get_cookiefile` function of this script: https://instaloader.github.io/troubleshooting.html

I think it would be much easier for users to specify just the browser name instead of the trouble to find a browser profile path.

ekalin:
If the user has only one profile than just selecting the browser is enough. But a user might have several browser profiles, so an option like `-c` from that script is useful. But it should be optional; if not specified, let the library use the default.

JunedKhan101:
@ekalin By `several browser profiles` do you mean multiple Instagram accounts on a single browser? or one or more Instagram accounts on diffrent browsers?

If there are multiple Instagram accounts on a browser the code just fetches the cookie of the currently active or logged in Instagram user so the Instaloader user has to make sure he is the currently logged in or active user.

I don't know what I am expected to do here would you mind if you can clarify further?

ekalin:
It's multiple profile directories, each used for different users. Or a browser profile just for logging in and then using instaloader, separate from the main browser. In practice the whole directory with user data (cookies, history, settings, and everything else) is changed, and one can have completely separate instances of the browser.

Firefox has this feature built-in (though it's not widely used): https://support.mozilla.org/en-US/kb/profile-manager-create-remove-switch-firefox-profiles#

Chrome also allows this, at least manually: https://www.chromium.org/developers/creating-and-using-profiles/ .

In practice, I imagine what's necessary is a just a command-line option to set the directory of the browser profile, and then passing this directory to the library. I don't really know if the library supports that, but it would be a big limitation if it doesn't.

JunedKhan101:
Yeah, you are right. The library doesn't support browser profiles, it's a limitation.

ekalin:
It does. It does not mention profiles, but one can point to the specific location of the cookies file. Which is basically what a profile is, just a location to store all the settings (including cookies).

This is exactly what's necessary, and is how the snippet in the documentation works. And if the user does not specify the location, then let the library find the default location.

docs/cli-options.rst
```diff
@@ -196,6 +196,10 @@ Instaloader to login.
 
    Login name (profile name) for your Instagram account.
 
+.. option:: --load-cookies BROWSER-NAME, -lc BROWSER-NAME
```
aandergr, line 199: I would not use `-lc` as a option name, we always use single letter option names.

It would also be nice if the profile or the path to the cookie file could be set by the user, similar to the provisional cookie import script mentioned in our documentation, as there are many users who have multiple profiles in use.

instaloader/__main__.py
```diff
@@ -14,6 +14,7 @@
 from .instaloader import (get_default_session_filename, get_default_stamps_filename)
 from .instaloadercontext import default_user_agent
 from .lateststamps import LatestStamps
+import browser_cookie3
```
aandergr, line 17: This would also need to be listed in the package metadata in `setup.py` and in the Pipfile.

It should be an optional dependency, with the feature being disabled if the package is not installed.

JunedKhan101:
Different browsers handle cookies differently, Firefox stores cookies as an sqlite data in which case we need to query the data, this is already implemented in the troubleshooting login error script. In Firefox case there is little role of browser_cookie3 as we need to just query the data using sqlite.

Right now, as of writing this comment browser_cookie3 has stopped working on chrome due to a chrome update which stores cookies encrypted reference: https://github.com/borisbabic/browser_cookie3/issues/180 this has been tested by me it doesn't work on chrome anymore unless the user specifies to store cookies unencrypted.

browser_cookie3 has also stopped working on brave and edge (most likely due to the same chromium update this has not been tested by me and I'm only making assumptions based on issues tab of the library on github)

My question: Should I just focus on using the cookies from firefox? because fetching the encrypted cookies of different browsers, handling cross platform decryption is no easy task. This is probably why the author of previous PR stopped working on this issue.

Also, I'm a bit unclear about using the cookie file path with browser_cookie3 library but I'll try to do more R&D on that.
@ekalin you said the cookie file path can be used with the library would you mind if you can provide an exmaple? that'll be helpful.

ekalin:
The constructors take an optional argument `cookie_file` pointing to the desired `cookies.sqlite` file (or the Chrome equivalent).

I'd create an optional argument that points to the desired file, and pass this to the library. If the argument was not specified, then the default behavior can be used (which looks for the default profile).

JunedKhan101:
@ekalin @aandergr I have implemeted the feature to make use of browser cookie file and I have tested this on FireFox, it will pick the the default profile if cookie file is not specified by the user.

instaloader/__main__.py
aandergr: The package also needs to be added to `Pipfile.lock`. However I just tried doing this myself and run into a problem with resolving the depencencies due to conflicts in sub-depencencies. I will soon try to have a deeper lock into that.

instaloader/__main__.py
```diff
@@ -66,6 +67,45 @@ def filterfunc(item) -> bool:
...
+        return Cookies
+
+def import_session(browser, instaloader, cookiefile):
+    cookie = get_cookies_from_instagram('instagram', browser, cookiefile)
+    if cookie is not None:
+        instaloader.context._session.cookies.update(cookie)
+        username = instaloader.test_login()
+        if not username:
+            raise SystemExit(f"Not logged in. Are you logged in successfully in {browser}?")
+        instaloader.context.username = username
```
aandergr, line 108: This function mostly accesses properties of InstaloaderContext, one of them even a protected property. Maybe we could just move it to being a method of InstaloaderContext?

instaloader/__main__.py
aandergr, line 17: Since this is an optional dependency, it should be wrapped into a `try` block, that disables the new feature if that import is missing, e.g. by setting a variable in that case and then `--load-cookies` showing a message that `browser_cookie3` needs to be installed.

instaloader/__main__.py
```diff
@@ -66,6 +67,45 @@ def filterfunc(item) -> bool:
...
+        "opera": browser_cookie3.opera,
+        "safari": browser_cookie3.safari
+    }
+
+    if browser not in supported_browsers:
+        print("Loading cookies from the specified browser failed. Supported browsers are Chrome, Firefox, Edge, Brave, Opera and Safari")
+        return {}
+
+    Cookies = {}
+    browserCookies = list(supported_browsers[browser](cookie_file=cookiefile))
```
aandergr, line 85: To be consistent with how variables are named, it should rather be `cookies` and `browser_cookies`.

However I do not well understand the purpose of the `Cookies` variable. Is it just to filter the `browserCookies` that have the `domain` attribute set?

docs/cli-options.rst
aandergr: Nice, I think `-b` and `-B` are fine.

instaloader/__main__.py
```diff
@@ -66,6 +67,45 @@ def filterfunc(item) -> bool:
...
+        "chrome": browser_cookie3.chrome,
+        "firefox": browser_cookie3.firefox,
+        "edge": browser_cookie3.edge,
+        "brave": browser_cookie3.brave,
+        "opera": browser_cookie3.opera,
+        "safari": browser_cookie3.safari
+    }
+
+    if browser not in supported_browsers:
+        print("Loading cookies from the specified browser failed. Supported browsers are Chrome, Firefox, Edge, Brave, Opera and Safari")
```
aandergr, line 81: This line is very long, we have set a maximum of 120 characters.

instaloader/__main__.py
JunedKhan101: Yes, `browser_cookies` returns a list of cookies which we need to format in order to set cookies with ease this is done using `instaloader.context._session.cookies.update(cookie)` 

In this case cookie is a key value pair of cookie name and value that was **filtered and formatted** in the `get_cookies_from_instagram` function.

instaloader/__main__.py
JunedKhan101: I have broken down this line.

instaloader/__main__.py
JunedKhan101: I have fixed the naming conventions

instaloader/__main__.py
JunedKhan101: I have made a commit that addresses this.

JunedKhan101: Thanks, I have minimum context on this so help would be much appreciated.

jarellgalura (approved):

instaloader/__main__.py
aandergr: We now managed to update all our dependencies in Pipfile.lock, this allows us to use `pipenv install` again without running into problems.

We did this change on `upcoming/v4.11`  where this MR should also be rebased on, since it adds a new feature.

Once it is rebased, it should be possible to add the dependency with `pipenv install browser_cookie3`. This will cause the package to be installed in the Github Actions jobs that check each change, and in the Windows EXE file which we ship with each release.

It should also be mentioned in `setup.py` as an optional dependency.

The import should be in a `try` block, so that the feature can be disabled if the package is missing.

instaloader/__main__.py
aandergr: The `_session` object is not supposed to be accessed from outside of the InstaloaderContext.

instaloader/__main__.py
JunedKhan101: "The import should be in a try block, so that the feature can be disabled if the package is missing."

This one is done as well.

instaloader/__main__.py
JunedKhan101: Shall I create a method in InstaloaderContext that returns the `_session`? would that be fine?

for example somthing like:

```python
def get_session(self):
        return self._session
```

then maybe in the `import_session` function I can do something like the following to update cookies:

```python
session = instaloader.context.get_session()
session.cookies.update(cookie)
```

In this example the `get_session` method is just the public method (or wrapper one might say) to get the session from InstaloaderContext. This avoid using session directly in InstaloaderContext.

aandergr: The session object should be kept as private as possible to the InstaloaderContext, if we would allow to manipulate it from outside we would invite many strange bug reports. We would really favor to move the part that updates the login cookie to InstaloaderContext.

Pipfile
```diff
@@ -15,6 +15,7 @@ typing-extensions = "*"
 types-requests = "*"
 typed-ast = "*"
 jinja2 = "~=3.0.1"
+browser_cookie3 = "*"
```
aandergr, line 18: Could you rebase this PR to `upcoming/v4.11` and then do `pipenv install browser_cookie3` to properly update the Pipfile and Pipfile.lock?

docs/cli-options.rst
```diff
@@ -196,6 +196,14 @@ Instaloader to login.
 
    Login name (profile name) for your Instagram account.
 
+.. option:: --load-cookies BROWSER-NAME, -b BROWSER-NAME
+
+   Use Instagram cookie in your browser to login (works with --login). Supported browsers: Chrome, Firefox, Edge, Opera, Safari, Brave.
```
aandergr, line 201:
```suggestion
   Use Instagram cookie in your browser to login (works with :option:`--login`). Supported browsers: Chrome, Firefox, Edge, Opera, Safari, Brave.
```

(this causes `--login` to be a link to the definition of the login option)

instaloader/__main__.py
```diff
@@ -14,6 +14,11 @@
 from .instaloader import (get_default_session_filename, get_default_stamps_filename)
 from .instaloadercontext import default_user_agent
 from .lateststamps import LatestStamps
+try:
+    import browser_cookie3
+    is_library_installed = True
```
aandergr, line 19: The variable name is too generic.

instaloader/instaloadercontext.py
```diff
@@ -211,6 +211,9 @@ def save_session(self):
         """Not meant to be used directly, use :meth:`Instaloader.save_session`."""
         return requests.utils.dict_from_cookiejar(self._session.cookies)
 
+    def update_cookie(self, cookie):
```
aandergr, line 214: I would call it `update_cookies` (plural) since it actually updates multiple cookies.

instaloader/__main__.py
```diff
@@ -66,6 +71,46 @@ def filterfunc(item) -> bool:
...
+        "safari": browser_cookie3.safari
+    }
+
+    if browser not in supported_browsers:
+        print("Loading cookies from the specified browser failed")
+        print("Supported browsers are Chrome, Firefox, Edge, Brave, Opera and Safari")
+        return {}
+
+    cookies = {}
+    browser_cookies = list(supported_browsers[browser](cookie_file=cookiefile))
```
aandergr, line 90: It would be nice if the parameter names were more consistent in their style, e.g. `cookie_file` and `cookie_name` rather than mixing `cookieName` and `cookiefile`, `cookie_file`.

instaloader/instaloadercontext.py
JunedKhan101: Updated

instaloader/__main__.py
JunedKhan101: Done.

instaloader/__main__.py
JunedKhan101: Name updated

Pipfile
JunedKhan101: Will do once all the PR reviews are approved.

aandergr: Looks good, I think this is ready to be rebased :)

JunedKhan101: @aandergr Rebased is completed, Pipfile and Pipfile.lock are updated as well.
Let me know if everything looks right :)

aandergr:
Amazing, thanks!

tap0212:
hey @JunedKhan101 I'm using the library 
- I'm creating the session and cookies using this new feature that you've added
- and mounting it using `load_session_from_file` 
- how can I keep updating the session and cookie automatically upon expiration?


Please let me know if there's any existing way to do this!
Thanks

Issue #1737, opened 2022-09-10
JSON Query to accounts/login/: Could not find "window._sharedData" in html response.


axelmiami:
Exactly the same problem:

```
$ instaloader --login=<username>
Session file does not exist yet - Logging in.
Enter Instagram password for p2i.berest:
JSON Query to accounts/login/: Could not find "window._sharedData" in html response. [retrying; skip with ^C]
JSON Query to accounts/login/: Could not find "window._sharedData" in html response. [retrying; skip with ^C]
Fatal error: JSON Query to accounts/login/: Could not find "window._sharedData" in html response.
```
help me please

Moreover, when launched locally from the IDE, everything connects and works, but on the server it gives an error.

All Python packages are the same and updated to the latest versions.
```
$ python3 -m pip install instaloader -U
Requirement already satisfied: instaloader in ./venv/lib/python3.8/site-packages (4.9.3)
Requirement already satisfied: requests>=2.4 in ./venv/lib/python3.8/site-packages (from instaloader) (2.28.1)
Requirement already satisfied: certifi>=2017.4.17 in ./venv/lib/python3.8/site-packages (from requests>=2.4->instaloader) (2022.6.15.1)
Requirement already satisfied: charset-normalizer<3,>=2 in ./venv/lib/python3.8/site-packages (from requests>=2.4->instaloader) (2.1.1)
Requirement already satisfied: idna<4,>=2.5 in ./venv/lib/python3.8/site-packages (from requests>=2.4->instaloader) (3.3)
Requirement already satisfied: urllib3<1.27,>=1.21.1 in ./venv/lib/python3.8/site-packages (from requests>=2.4->instaloader) (1.26.12)`
```

All of the above methods did not help.
Both the call from Python and the call in the console do not work on the server

axelmiami:
I tried following the instructions in #1740, but I still can't log in

aandergr:
Indeed, it seems that the login is currently not working.

Similar errors have been reported in #1635 and #1758.

Until this is fixed, we recommend the workaround described in https://instaloader.github.io/troubleshooting.html#login-error.

anuraag488:
> Indeed, it seems that the login is currently not working.
> 
> Similar errors have been reported in #1635 and #1758.
> 
> Until this is fixed, we recommend the workaround described in https://instaloader.github.io/troubleshooting.html#login-error.

Can you give some option for command line? I want to import cookies in Termux Android.

akafanfan:
[...]

Issue #2086, opened 2023-10-11
Login Error JSON Query to accounts/login/: Could not find "window._sharedData" in html response.


sarvagyaah:
import instaloader
L = instaloader.Instaloader()
L.login('username','password')

Error:
JSON Query to accounts/login/: Could not find "window._sharedData" in html response.[retrying; skip with ^C]
JSON Query to accounts/login/: Could not find "window._sharedData" in html response. [retrying; skip with ^C]

Please fix this issue 
is there any alternative to instaloader liabrary 
please fix i need urgently

EveryDirection:
+1, issue confirmed. Issue #2084 seems related. But I never had this problem before today; Instaloader was PREVIOUSLY working fine until today.

Issue #2084, opened 2023-10-07
I am not able to login with instaloader


nawar-unxd:
**Not able to login**
I am trying to use login(username, password) or interactive_login(username), however both of them are not working

**Here is my log after running these functions**
JSON Query to accounts/login/: Could not find "window._sharedData" in html response. [retrying; skip with ^C]
JSON Query to accounts/login/: Could not find "window._sharedData" in html response. [retrying; skip with ^C]
Traceback (most recent call last):
instagram-scraper/venv/lib/python3.9/site-packages/instaloader/instaloadercontext.py", line 410, in get_json
    raise QueryReturnedNotFoundException("Could not find \"window._sharedData\" in html response.")
instaloader.exceptions.QueryReturnedNotFoundException: Could not find "window._sharedData" in html response.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
instagram-scraper/venv/lib/python3.9/site-packages/instaloader/instaloadercontext.py", line 410, in get_json
    raise QueryReturnedNotFoundException("Could not find \"window._sharedData\" in html response.")
instaloader.exceptions.QueryReturnedNotFoundException: Could not find "window._sharedData" in html response.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):instagram-scraper/venv/lib/python3.9/site-packages/instaloader/instaloadercontext.py", line 410, in get_json
    raise QueryReturnedNotFoundException("Could not find \"window._sharedData\" in html response.")
instaloader.exceptions.QueryReturnedNotFoundException: Could not find "window._sharedData" in html response.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
instagram-scraper/instagram_client.py", line 19, in <module>
    save_session(username, password)
instagram-scraper/instagram_client.py", line 7, in save_session
    L.login(username, password)
instagram-scraper/venv/lib/python3.9/site-packages/instaloader/instaloader.py", line 651, in login
    self.context.login(user, passwd)
instagram-scraper/venv/lib/python3.9/site-packages/instaloader/instaloadercontext.py", line 259, in login
    csrf_json = self.get_json('accounts/login/', {}, session=session)
instagram-scraper/venv/lib/python3.9/site-packages/instaloader/instaloadercontext.py", line 448, in get_json
    return self.get_json(path=path, params=params, host=host, session=sess, _attempt=_attempt + 1,
[...]

Issue #2069, opened 2023-09-11
window._sharedData


Jaudb4:
JSON Query to accounts/login/: Could not find "window._sharedData" in html response. [retrying; skip with ^C]
JSON Query to accounts/login/: Could not find "window._sharedData" in html response. [retrying; skip with ^C]
Traceback (most recent call last):
  File "/data/user/0/ru.iiec.pydroid3/files/aarch64-linux-android/lib/python3.9/site-packages/instaloader/instaloadercontext.py", line 410, in get_json
    raise QueryReturnedNotFoundException("Could not find \"window._sharedData\" in html response.")
instaloader.exceptions.QueryReturnedNotFoundException: Could not find "window._sharedData" in html response.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/data/user/0/ru.iiec.pydroid3/files/aarch64-linux-android/lib/python3.9/site-packages/instaloader/instaloadercontext.py", line 410, in get_json
    raise QueryReturnedNotFoundException("Could not find \"window._sharedData\" in html response.")
instaloader.exceptions.QueryReturnedNotFoundException: Could not find "window._sharedData" in html response.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/data/user/0/ru.iiec.pydroid3/files/aarch64-linux-android/lib/python3.9/site-packages/instaloader/instaloadercontext.py", line 410, in get_json
    raise QueryReturnedNotFoundException("Could not find \"window._sharedData\" in html response.")
instaloader.exceptions.QueryReturnedNotFoundException: Could not find "window._sharedData" in html response.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/data/user/0/ru.iiec.pydroid3/files/accomp_files/iiec_run/iiec_run.py", line 31, in <module>
    start(fakepyfile,mainpyfile)
  File "/data/user/0/ru.iiec.pydroid3/files/accomp_files/iiec_run/iiec_run.py", line 30, in start
    exec(open(mainpyfile).read(),  __main__.__dict__)
  File "<string>", line 7, in <module>
  File "/data/user/0/ru.iiec.pydroid3/files/aarch64-linux-android/lib/python3.9/site-packages/instaloader/instaloader.py", line 651, in login
    self.context.login(user, passwd)
[...]

PR #2088, opened 2023-10-15
Fixed error when trying to log in due to changes by Instagram

LostXOR:
Due to changes by Instagram, the JSON previously found in `window.sharedData` is now a string located deep in some nested JSON tags. This change caused `instaloader.get_json` to throw an error in some circumstances.
Instagram also changed the contents of the JSON, making the `entry_data` key empty. This causes `instaloader.get_json` to throw another error.

The first error occurs when trying to log in, making logging in with a username and password impossible. (Logging in with a session file still works).

This PR fixes both errors by updating `get_json` to get the JSON from its new location, and removing the logic checking the `entry_data` key, which I believe is unnecessary. It fixes issues #1737, #2044, #2059, #2062, #2069, #2084, and #2086 (all related to login).

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -405,20 +405,13 @@ class InstaloaderContext:
                 raise ConnectionException("HTTP error code {}.".format(resp.status_code))
             is_html_query = not is_graphql_query and not "__a" in params and host == "www.instagram.com"
             if is_html_query:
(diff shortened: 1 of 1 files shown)


aandergr:
Awesome, this fixes get_json for the case of is_html_query, which is used for the login. Thanks a lot for your contribution!

PR #2119, opened 2023-11-29
Implemented fix to avoid simuntaneous use of --login and --load-cookies

JunedKhan101:
Related issue: https://github.com/instaloader/instaloader/issues/1801
This fix avoid the simultaneous use of `--login` and `--load-cookies`.
This is necessary to avoid the conflict between username passed by the user and the logged in user account in a browser.

### Test case 1:

Testing instaloader with `--login` and `--load-cookies`
Expected result is to throw an error that says these two option cannot be used together.

Command Ran:
```console
python instaloader.py --login=its_juned42 --load-cookies=firefox
```

Observed output:
```
--load-cookies and --login cannot be used together.
```

### Test case 2:

Loading cookies by specifying a `--cookiefile`
This also works as expected.

Command Ran:
```console
python instaloader.py --load-cookies=firefox --cookiefile=C:\Users\juned_5f6qngh\AppData\Roaming\Mozilla\Firefox\Profiles\en1b55sk.default-release\cookies.sqlite
```

Observed output:
```
Cookies loaded successfully from firefox
Saved session to C:\Users\juned_5f6qngh\AppData\Local\Instaloader\session-its_juned42.
No targets were specified, thus nothing has been downloaded.
```

I deleted session file after this just to make sure it doesn't use it while I'm testing it.

### Test case 3:

Loading cookies normally by specifying a browser name.

Command Ran:
```console
python instaloader.py --load-cookies=firefox
```

Observed output:
```
Cookies loaded successfully from firefox
Saved session to C:\Users\juned_5f6qngh\AppData\Local\Instaloader\session-its_juned42.
No targets were specified, thus nothing has been downloaded.
```

  - Is the documentation updated (if appropriate)? No, not needed
  - Do you consider it ready to be merged or is it a draft? Yes
  - Can we help you at some point? Not at the moment

### DIFF
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -205,7 +205,7 @@ Instaloader to login.
    Supported browsers: Chrome, Firefox, Edge, Opera, Safari, Brave.
 
    After loading the cookies run the :option:`--login` option as it is required to download high quality media
-   and to make full use of instaloader's features.
+   and to make full use of Instaloader's features.
 
    .. versionadded:: 4.1
 
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -199,12 +199,15 @@ Instaloader to login.
 .. option:: --load-cookies BROWSER-NAME, -b BROWSER-NAME
 
    Use Instagram cookie in your browser to login.
+   This feature requires the browser_cookie3 library.
    Compatible with :option:`--cookiefile` if you want to load cookies from browser profiles.
    Incompatible with :option:`--login` due to potential username mismatch between user input and browser login.
    Supported browsers: Chrome, Firefox, Edge, Opera, Safari, Brave.
 
    After loading the cookies run the :option:`--login` option as it is required to download high quality media
    and to make full use of instaloader's features.
+
+   .. versionadded:: 4.1
 
 .. option:: --cookiefile COOKIE-FILE, -B COOKIE-FILE
 
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -207,7 +207,7 @@ Instaloader to login.
    After loading the cookies run the :option:`--login` option as it is required to download high quality media
    and to make full use of Instaloader's features.
 
-   .. versionadded:: 4.1
+   .. versionadded:: 4.11
 
 .. option:: --cookiefile COOKIE-FILE, -B COOKIE-FILE
 
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -198,7 +198,13 @@ Instaloader to login.
 
 .. option:: --load-cookies BROWSER-NAME, -b BROWSER-NAME
 
-   Use Instagram cookie in your browser to login (works with :option:`--login`). Supported browsers: Chrome, Firefox, Edge, Opera, Safari, Brave.
+   Use Instagram cookie in your browser to login.
+   Compatible with :option:`--cookiefile` if you want to load cookies from browser profiles.
+   Incompatible with :option:`--login` due to potential username mismatch between user input and browser login.
+   Supported browsers: Chrome, Firefox, Edge, Opera, Safari, Brave.
+
+   After loading the cookies run the :option:`--login` option as it is required to download high quality media
+   and to make full use of instaloader's features.
 
 .. option:: --cookiefile COOKIE-FILE, -B COOKIE-FILE
 
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -114,8 +114,8 @@ def import_session(browser, instaloader, cookiefile):
         instaloader.context.username = username
         print(f"{username} has been successfully logged in.")
         next_step_text = (f"Next: Run instaloader --login={username} as it is required to download high quality media "
-            "and to make full use of instaloader's features.")
-        print(textwrap.fill(next_step_text, width=80))
+                            "and to make full use of instaloader's features.")
+        print(textwrap.fill(next_step_text))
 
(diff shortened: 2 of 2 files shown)


aandergr:
Thanks, this fixes the behavior!

However, in the documentation there is

```rst
.. option:: --load-cookies BROWSER-NAME, -b BROWSER-NAME

   Use Instagram cookie in your browser to login (works with :option:`--login`). Supported browsers: Chrome, Firefox, Edge, Opera, Safari, Brave.
```

*works with --login* is is now technically wrong, considering that passing both --login and --load-cookies leads to an error message. Could you adapt the description so that it is clear that it is intended that you first load the cookies with --load-cookies and then use them later with --login? It would also be helpful as output on the console, I think, if after the cookie import a hint is displayed on how to use the imported cookies with --login.

JunedKhan101:
@aandergr I have updated docs and added hint message after loading the cookies.

Can you check if it's according to your liking?

Also nested if's were kinda bothering me so I replaced them if and elif.

instaloader/__main__.py
```diff
@@ -95,6 +95,8 @@ def get_cookies_from_instagram(domain, browser, cookie_file='', cookie_name=''):
 
     if cookies:
         print(f"Cookies loaded successfully from {browser}")
+        print("Next: Run instaloader --login=USERNAME as it is required to download high quality media \
+            and to make full use of instaloader's features")
```
aandergr, line 99: This prints

```
Next: Run instaloader --login=USERNAME as it is required to download high quality media             and to make full use of instaloader's features
```

- There is a very big space between the words `media` and `and`
- I think it would be cool if it showed the actual username instead of `USERNAME'.

You could use [textwrap.fill](https://docs.python.org/3/library/textwrap.html#textwrap.fill) to nicely format that long message.

JunedKhan101: Just committed the new changes, Please review.

instaloader/__main__.py
```diff
@@ -111,6 +112,10 @@ def import_session(browser, instaloader, cookiefile):
         if not username:
             raise SystemExit(f"Not logged in. Are you logged in successfully in {browser}?")
         instaloader.context.username = username
+        print(f"{username} has been successfully logged in.")
+        next_step_text = (f"Next: Run instaloader --login={username} as it is required to download high quality media "
+            "and to make full use of instaloader's features.")
+        print(textwrap.fill(next_step_text, width=80))
```
aandergr, line 118: In another place we use `textwrap.fill()` without any explicit `width` parameter, leading to the default of 70. We should be consistent here, I do not see a need to have a wider text at this place.

instaloader/__main__.py
```diff
@@ -111,6 +112,10 @@ def import_session(browser, instaloader, cookiefile):
         if not username:
             raise SystemExit(f"Not logged in. Are you logged in successfully in {browser}?")
         instaloader.context.username = username
+        print(f"{username} has been successfully logged in.")
+        next_step_text = (f"Next: Run instaloader --login={username} as it is required to download high quality media "
+            "and to make full use of instaloader's features.")
```
aandergr, line 117: We always write capitalized `Instaloader`, with a capitalized `I`, unless we refer to the CLI command or the python module.

Also I'd indent the second line further, so that it has the same indentation as the opening parenthesis in the first line.

instaloader/__main__.py
```diff
@@ -140,6 +145,11 @@ def _main(instaloader: Instaloader, targetlist: List[str],
     if latest_stamps_file is not None:
         latest_stamps = LatestStamps(latest_stamps_file)
         instaloader.context.log(f"Using latest stamps from {latest_stamps_file}.")
+    # load cookies if browser is not None
+    if browser and bc3_library:
+        import_session(browser.lower(), instaloader, cookiefile)
+    elif browser and bc3_library is False:
```
aandergr, line 151:
```suggestion
    elif browser and not bc3_library:
```

I think `not ...` is a littler more pythonic and cooler than `... is False`.

docs/cli-options.rst
```diff
@@ -198,7 +198,13 @@ Instaloader to login.
...
 .. option:: --load-cookies BROWSER-NAME, -b BROWSER-NAME
 
-   Use Instagram cookie in your browser to login (works with :option:`--login`). Supported browsers: Chrome, Firefox, Edge, Opera, Safari, Brave.
+   Use Instagram cookie in your browser to login.
+   Compatible with :option:`--cookiefile` if you want to load cookies from browser profiles.
+   Incompatible with :option:`--login` due to potential username mismatch between user input and browser login.
+   Supported browsers: Chrome, Firefox, Edge, Opera, Safari, Brave.
+
+   After loading the cookies run the :option:`--login` option as it is required to download high quality media
+   and to make full use of instaloader's features.
```
aandergr, line 207:
```suggestion
   and to make full use of Instaloader's features.
```

I think here we should also mention that this feature requires `browser_cookie3`.

Oh and could you also add the `.. versionadded:: 4.11` directive here?

instaloader/__main__.py
JunedKhan101: Remove the width parameter.

instaloader/__main__.py
JunedKhan101: Done.

instaloader/__main__.py
JunedKhan101: Done.

docs/cli-options.rst
JunedKhan101: Done.

aandergr:
Looks good, thanks!

PR #2133, opened 2023-12-08
Fix 403 during login due to incorrect login endpoint and CSRF token location

### LINKED ISSUE #2126: Login error: JSON decode fail, 403 - Forbidden
I can't use the instaloader anymore, it's still appear this error: **Login error: JSON decode fail, 403 - Forbidden**. I check the code, but nothing work, Can you help me?

LostXOR:
Instagram made some changes on their end that resulted in an error when logging in.

```
Fatal error: Login error: JSON decode fail, 403 - Forbidden.
```

This is caused by instaloader using the wrong endpoint (`/accounts/login/ajax/` instead of `/api/v1/web/accounts/login/ajax/`). Additionally, the way instaloader gets the CSRF token (from the JSON returned by `/accounts/login/`) no longer works and throws another error.

```
Fatal error: Login error: "fail" status, message "CSRF token missing or incorrect".
```

This PR fixes both errors and makes login work again. It also removes the `is_html_query` part of `instaloadercontext.get_json`, which was only used for login and is no longer necessary.

Fixes #2126.

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -256,18 +256,25 @@ class InstaloaderContext:
         # Override default timeout behavior.
         # Need to silence mypy bug for this. See: https://github.com/python/mypy/issues/2427
         session.request = partial(session.request, timeout=self.request_timeout) # type: ignore
-        csrf_json = self.get_json('accounts/login/', {}, session=session)
-        csrf_token = csrf_json['config']['csrf_token']
+
+        self.do_sleep()
+        # Make a request to Instagram's root URL, which will set the session's csrftoken cookie
+        # Not using self.get_json() here, because we need to access the cookie
+        csrf_request = session.get('https://www.instagram.com/')
+        # Add session's csrftoken cookie to session headers
+        csrf_token = session.cookies.get_dict()['csrftoken']
         session.headers.update({'X-CSRFToken': csrf_token})
-        # Not using self.get_json() here, because we need to access csrftoken cookie
+
         self.do_sleep()
         # Workaround credits to pgrimaud.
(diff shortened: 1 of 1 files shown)


ericdoh0401:
I tried updating instaloder; however, the issues remain.

LostXOR:
> I tried updating instaloder; however, the issues remain.

Did you download it from my fork? The fix is just a pull request for now, and until it gets merged it won't be in the official version of instaloader.

instaloader/instaloadercontext.py
```diff
@@ -256,18 +255,24 @@ def login(self, user, passwd):
         # Override default timeout behavior.
         # Need to silence mypy bug for this. See: https://github.com/python/mypy/issues/2427
         session.request = partial(session.request, timeout=self.request_timeout) # type: ignore
-        csrf_json = self.get_json('accounts/login/', {}, session=session)
-        csrf_token = csrf_json['config']['csrf_token']
+
+        self.do_sleep()
```
aandergr, line 259: At this point `do_sleep()` seems unnecessary to me, no request has been sent here, so the function waits at the start before doing anything.

aandergr (commented):
Amazing, this once again fixes the login. Thanks for sharing this change with us!

I have only one minor comment, then we can release this fix.

instaloader/instaloadercontext.py
LostXOR: In the old code, `get_json` (which has a `do_sleep()` at the start) is called at the start of login. Since my changes made `get_json` not used during login, I moved the `do_sleep()` to before the request. Feel free to remove it if you think it's unnecessary.

aandergr: Oh indeed, thanks for that careful observation. Probably that was to defer the request rather than the processing of data it responds.

PR #2147, opened 2023-12-27
Change pinned post detection, fixes #2142

### LINKED ISSUE #2142: Pinned posts are not correctly detected anymore, causes --latest-stamps to fail
**Describe the bug**
Pinned posts are not detected anymore by `Post.is_pinned`. The "pin" icon is also missing when opening the profile on the browser. This missing pin icon seems to be happening for a long time, but `Post.is_pinned` would correctly return whether the post was pinned, but now `Post.is_pinned` is returning `False` for all posts.

**To Reproduce**
See for example https://www.instagram.com/fefe/. The first three or four posts are older, then the regular chronological sequence starts. There's no real way to see the behavior, but with debugging statements one can see that `Post.is_pinned` returns `False` for those first items.

**Expected behavior**
`Post.is_pinned` should return `True` for pinned posts, so they can be treated specially when downloading only the newer posts.

**Instaloader version**
v4.10.2

**Additional context**
This is breaking `--latest-stamps` (kind of like when pinned posts were introduced): since the first post is older, no new posts are downloaded. It is also setting the date of the last downloaded post to the date of that old post, so when the problem is fixed some posts will be downloaded again.

Hopefully there's still some other flag in the post json structure that indicates whether it's a pinned post.

ekalin:
Fixes #2142.

The property that indicated whether a post is pinned is not available anymore. As a workaround, the first five posts are treated differently (as pinned posts were): if they were already downloaded, they are skipped, but they do not cause the iteration to stop.

*  - Is it just a proof of concept?* No
*  - Is the documentation updated (if appropriate)?* Yes
*  - Do you consider it ready to be merged or is it a draft?* Ready
*  - Can we help you at some point?* I'm open to suggestion for improvement

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -991,7 +991,8 @@ class Instaloader:
                             max_count: Optional[int] = None,
                             total_count: Optional[int] = None,
                             owner_profile: Optional[Profile] = None,
-                            takewhile: Optional[Callable[[Post], bool]] = None) -> None:
+                            takewhile: Optional[Callable[[Post], bool]] = None,
+                            possibly_pinned: int = 0) -> None:
         """
         Download the Posts returned by given Post Iterator.
 
@@ -1003,6 +1004,9 @@ class Instaloader:
         .. versionchanged:: 4.8
            Add `takewhile` parameter.
 
+        .. versionchanged:: 4.10.3
+           Add `possibly_pinned` parameter.
+
         :param posts: Post Iterator to loop through.
         :param target: Target name.
         :param fast_update: :option:`--fast-update`.
@@ -1011,6 +1015,8 @@ class Instaloader:
         :param total_count: Total number of posts returned by given iterator.
         :param owner_profile: Associated profile, if any.
         :param takewhile: Expression evaluated for each post. Once it returns false, downloading stops.
+        :param possibly_pinned: Number of posts that might be pinned. These posts do not cause download
+               to stop even if they've already been downloaded.
         """
         displayed_count = (max_count if total_count is None or max_count is not None and max_count < total_count
                            else total_count)
@@ -1032,7 +1038,7 @@ class Instaloader:
         ) as (is_resuming, start_index):
             for number, post in enumerate(posts, start=start_index + 1):
                 should_stop = not takewhile(post)
-                if should_stop and number <= 5:
+                if should_stop and number <= possibly_pinned:
                     continue
                 if (max_count is not None and number > max_count) or should_stop:
                     break
@@ -1066,7 +1072,7 @@ class Instaloader:
                         except PostChangedException:
                             post_changed = True
                             continue
-                    if fast_update and not downloaded and not post_changed and number > 5:
+                    if fast_update and not downloaded and not post_changed and number > possibly_pinned:
                         # disengage fast_update for first post when resuming
                         if not is_resuming or number > 0:
                             break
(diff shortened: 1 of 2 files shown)


kayiter23:
Tried this branch, it tried to retrieve the already downloaded profiles' metadata (i.e. all posts) even though it was given the "-F"/"--fast-update" command-line argument and all the posts were logged as "exists" and "unchanged".

ekalin:
You're right, there was a wrong condition that was considering more than the desired first five posts. Hopefully it's fixed now.

aandergr:
Nice change and a good idea to simply disregard the fast-update and takewhile stop criterium for the first five posts. Thanks for also updating the docstring of `is_pinned`.

I only have two minor suggestions:

Only up to three posts can be pinned, so it would suffice to limit the check to the first three posts.

> You can pin up to 3 posts you share on Instagram to the top of your profile.

(https://help.instagram.com/318456537074409/?cms_platform=iphone-app)

Also, posts_download_loop is used for every target that yields post, not only for Profiles, whereas only for Profiles Posts can be pinned. I would prefer to have that criterium only for Profiles and for other targets it should still stop as soon as the first already-downloaded post is encountered.

ekalin:
> Only up to three posts can be pinned, so it would suffice to limit the check to the first three posts.

I used five posts instead of three as a safeguard in case IG changed the limit. But it would only work if they changed the limit to four or five, if they decided one can pin six posts, the code would have to be changed anyway. So I guess it's fine to use three.

> Also, posts_download_loop is used for every target that yields post, not only for Profiles, whereas only for Profiles Posts can be pinned. I would prefer to have that criterium only for Profiles and for other targets it should still stop as soon as the first already-downloaded post is encountered.

How would one detect that it's Posts being downloaded? The closest I found was the `owner_profile` argument, but it is passed for posts and also IGTV posts (is this still a thing?).

However the method, one problem is that the conditionals would become even more complex with a check for the iteration type.

aandergr:
> How would one detect that it's Posts being downloaded? The closest I found was the `owner_profile` argument, but it is passed for posts and also IGTV posts (is this still a thing?).

It could be passed as a parameter flag to posts_download_loop.

> However the method, one problem is that the conditionals would become even more complex with a check for the iteration type.

Indeed, it would become a little less readable. However, depending on the use case (e.g. downloading `:saved`) it could then likely save requests when updating an almost-up-to-date collection. I think that's a good trade off.

ekalin:
That function already has a lot of parameters, but I guess that's one solution.

Instead of a flag, I thought of passing a number representing the number of initial posts to not consider for stopping. When downloading posts, I'd pass 3 to this new parameters, in other cases 0 is passed (or rather, nothing is passed, and it receives the value 0 by default). What do you think?

aandergr:
> Instead of a flag, I thought of passing a number representing the number of initial posts to not consider for stopping. When downloading posts, I'd pass 3 to this new parameters, in other cases 0 is passed (or rather, nothing is passed, and it receives the value 0 by default). What do you think?

I like that idea. I doubt that they will allow different numbers of pins for different targets, but it would move the number of maximum pins into the profile download function, and as such semantically closer to where it is relevant.

ekalin:
I've added the parameter, now the extra check only applies to profile posts.

aandergr:
Thanks!

I noticed that you are missing in our AUTHORS.md file where we list the most important contributors. We would be pleased if you'd add yourself in that file.

PR #2166, opened 2024-01-14
sys.exit(1) on abnormal termination

tari3x:
Exit with 1 if there was an error or the user aborted the run.

Motivation: after the download finishes, my bash script separates the videos, images, and metadata into separate folders. However, I only want to do that when the download finished completely, not when it was aborted.

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -230,8 +230,10 @@ def _main(instaloader: Instaloader, targetlist: List[str],
                                                    latest_stamps=latest_stamps)
     except KeyboardInterrupt:
         print("\nInterrupted by user.", file=sys.stderr)
+        sys.exit(1)
     except AbortDownloadException as exc:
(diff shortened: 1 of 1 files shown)


aandergr:
Good idea to use the exit code to inform the user that an irregular termination has occurred, we should have done that from the beginning. Thank you for tackling this! However, the way the change is proposed here, it skips the cleanup i.e. the saving of the session file. This should also happen in the event of an irregular termination.

tari3x:
Thanks! This is my first ever pull request, so I'm not super familiar with the workflow. I think it should be a lot easier for you to make sure this is idiomatic as you like it, compared to trying to explain it to me. But I'm not sure you can edit the pull request - maybe I need to give you write access on my fork for that?

ekalin:
I believe that not only a failure should be indicated, but different status codes could be used for different failures. I've opened an issue to discuss this: #2220.

Issue #2159, opened 2024-01-08
Reopen: Location download 'loop'


tari3x:
Please let me know if there's a better way to do this! 

Reopening https://github.com/instaloader/instaloader/issues/1341 - looks like it was never actually solved but it was closed by the bot, despite multiple other issues being duplicates of it.

I see exactly the same problem as described there.

tari3x:
https://github.com/instaloader/instaloader/issues/1470 seems to have the most information. I don't know much about web tech or instagram APIs, but if someone could consult me on those, I'd be happy to do the coding part.

aandergr:
Alright, we reopened that issue. :+1:

PR #2167, opened 2024-01-14
download all posts for a given location

tari3x:
This resolves https://github.com/instaloader/instaloader/issues/2159.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -1152,11 +1152,13 @@ class Instaloader:
         .. versionchanged:: 4.2.9
            Require being logged in (as required by Instagram)
         """
-        yield from SectionIterator(
+        yield from NodeIterator(
             self.context,
-            lambda d: d["native_location_data"]["recent"],
-            lambda m: Post.from_iphone_struct(self.context, m),
-            f"explore/locations/{location}/",
+            'ac38b90f0f3981c42092016a37c59bf7',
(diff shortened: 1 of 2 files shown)


tari3x:
Update: the change to ignore has_next_page seems necessary for locations, but does break for downloading profiles, so I guess you need an extra argument to the iterator.

aandergr:
We cannot merge a pull request that breaks downloading profiles.

Why does the if condition need to be removed for locations?

tari3x:
See the comment I left in the code: "This value seems to be a lie, at least for some location feeds.". So the feed says there's no next page, even though there are more posts, and it will give them to you if you ask explicitly.

tari3x:
As with the other pull request, this is a sketch that demonstrates the idea, and I'm hoping you'd be ok to apply the polish as you see fit.

aandergr:
Pull requests are the way to incorporate bug fixes or general improvements, also in the form of new features, into Instaloader. Before we can merge one, however, it must meet certain requirements in terms of usefulness and quality of implementation, only then does it make sense for the project to adopt the code. We are happy to offer support, but we simply cannot merge pull requests that disrupt existing functionality.

tari3x:
I am sorry to hear that you misunderstood what I said and opted to reject my contribution towards this functionality to instaloader. The multiple isssues that reference this will remain unresolved.

PR #2170, opened 2024-01-17
More Access to a JsonExportable's Metadata

adamcesco:
- A motivation for this change
  - Fixes the lack of access to post data that is only accessable through the instagram-private-api (e.g. saved_collection_ids).
  - added `get_json_structure(structure: JsonExportable)` function to be an attribute of an `Instaloader` instance. This allows for the access of post data that is not directly accessible through the Instaloader API (e.g. saved_collection_ids).

  - Is it just a proof of concept? `No`
  - Is the documentation updated (if appropriate)? `No`
  - Do you consider it ready to be merged or is it a draft? `Read to be merged`
  - Can we help you at some point? `Sure!`


aandergr:
The implementation does not seem well though-through. The `Post._iphone_struct` method is simply duplicated and private methods are called from outside of the class.

Issue #2172, opened 2024-01-20
How to slow down request via command line?


sgtpepperaut:
Instagram seems to be very strict with rate limits these days. Im using my own account and my own IP. 
Even when trying to only "fastupdate" one profile i am getting kicked out and getting a 401. 

When i return to the browser version i get:

_"We suspect automated behavior on your account
To prevent your account from being temporarily restricted or permanently disabled, ensure that no other users or tools have access to your account and that you're following our Terms of Use. Also consider changing your password to a stronger one to prevent unauthorized access to your account by third parties."

**Is there a way i can slow down the request?**

tari3x:
Here are the changes that work for me. I should maybe submit a pull request.

# in ~/.local/lib/instaloadercontext.py
# in class RateController:
#
# fast_sleeps = 10
#
#    def sleep(self, secs: float):
#        global fast_sleeps
#        if fast_sleeps > 0:
#            fast_sleeps = fast_sleeps - 1
#        else:
#            wait_time = random.uniform(15, 30)
#            print(wait_time)
#            time.sleep(wait_time)
#
# Remove > 0 guards in front of sleep further down

smasonlawrence:
> Here are the changes that work for me. I should maybe submit a pull request.
> 
> ```
> # in ~/.local/lib/instaloadercontext.py
> # in class RateController:
> #
> # fast_sleeps = 10
> #
> #    def sleep(self, secs: float):
> #        global fast_sleeps
> #        if fast_sleeps > 0:
> #            fast_sleeps = fast_sleeps - 1
> #        else:
> #            wait_time = random.uniform(15, 30)
> #            print(wait_time)
> #            time.sleep(wait_time)
> #
> # Remove > 0 guards in front of sleep further down
> ```

Can you please explain where I make this change? I cannot find any of these strings in any of the py files.. sorry, sorta python noob here still figuring things out.

tari3x:
Made a pull request.

tari3x:
Still pertinent.

nothing2obvi:
I am wondering the same thing. I got the warning for suspicious activity...I don't want Instagram to ban my account!

PR #2177, opened 2024-01-25
Random sleep

tari3x:
The current download speed seems too agressive: https://github.com/instaloader/instaloader/issues/2172. 

This change slows it down such that I'm able to download thousands of posts before getting banned.

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -230,8 +230,10 @@ def _main(instaloader: Instaloader, targetlist: List[str],
                                                    latest_stamps=latest_stamps)
     except KeyboardInterrupt:
         print("\nInterrupted by user.", file=sys.stderr)
+        sys.exit(1)
     except AbortDownloadException as exc:
         print("\nDownload aborted: {}.".format(exc), file=sys.stderr)
+        sys.exit(1)
     # Save session if it is useful
     if instaloader.context.is_logged_in:
         instaloader.save_session_to_file(sessionfile)
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -19,7 +19,6 @@ import requests.utils
 
 from .exceptions import *
 
-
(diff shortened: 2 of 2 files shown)


GabeMillikan:
[This](https://github.com/instaloader/instaloader/blob/5509ea7daf7f5e81534d4eb996c8afd478704426/instaloader/instaloadercontext.py#L345-L348l) is the appropriate location to adjust the sleeping logic.

tari3x:
I feel like I tried and it didn't work the expected way - wasn't granular enough. Can you remind me the difference between InstaloaderContext.do_sleep and RateController.sleep?

sapphirepro:
Yes, please accept this merge. Default included "sleep" is useless and not working causing fast downloads and fast ban

sapphirepro:
But also, maybe could add some timings between photos too, not only posts? Cause grabbing all 10 post photos instantly if obvious robot mode, human can not manually look instantly whole 10 photos with no time basically. So that between photos random between 1-5 secs and some 5-15 between posts. Then +/- human like

aandergr:
Does not make any sense. Apart from the many unrelated changes, this essentially destroys the rate controller by not waiting the determined interval, but rather a randomly determined, often shorter!, interval.

tari3x:
The problem with the current rate controller is that it does not work - it gets you banned pretty quickly. As with the other change, I am offering you a piece of code that does work pretty reliably. But since it doesn't make sense to you, I guess there's not much hope.

cyberjock:
I second this.  I'm trying to help a friend retrieve a copy of one of his family member's instagram page, and he keeps getting banned for too many requests.  Can't even get 100 pictures downloaded before instagram catches on.  Being that there's over 20k images, instaloader will not work for this job.  Even waiting 30 seconds between steps would be fine as we're not time limited.

(His family member passed away that owns the account) and his family asked him to try to retrieve all of the pictures)

PR #2195, opened 2024-02-12
use a factory method to instantiate sessions

Henkhogan:
Currently there is no way to customize the instantiation of session objects. I would need to use a dedicated proxy for the connections to insta.

If you'd like me to add some tests or docu, let me know.

Cheers


aandergr:
I like this approach.

This would introduce a way to also cleanly set custom session parameters, like SSL verification (suggested in #2221), or the request timeout (#330, #422), without that we need to pass-through and implement each of those parameters.

With the current implementation suggestion I am wondering how the user would actually override the session factory. Maybe InstaloaderContext and Instaloader should provide a parameter to specify the factory.

The handling of the request timeout might also be refactored to be done in the new session factory.

Henkhogan:
I currently use it like that 

```
def session_factory(self: InstaloaderContext):
    session = Session()
    session.trust_env = False
    return session

def instaloader_factory(max_connection_attempts=10):
    InstaloaderContext.session_factory = session_factory
    return Instaloader(compress_json=False, quiet=True, max_connection_attempts=max_connection_attempts)
```

which is still ugly but does that job

aandergr:
Since this is not a bugfix, could you rebase the change to `upcoming/v4.13`?

Can we somehow help you continuing this feature, do you need any further input or help?

Henkhogan:
Replaced by #2295

Issue #2186, opened 2024-02-04
get all external_url


pooyatfn:
Does it supports getting every single URL in external URLs?

    loader = instaloader.Instaloader()
    loader.login(USER, PASSWORD)

    profile = instaloader.Profile.from_username(loader.context, USERNAME)

    print("External url:", profile.external_url)

this code just give me the first external url in profile. but we can have many external urls in each profile. 
is there anyway to get all website links in profiles?

PR #2197, opened 2024-02-13
Add Profile.bio_links property

SayanthD:
Resolves https://github.com/instaloader/instaloader/issues/2186

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -937,6 +937,11 @@ class Profile:
         return self._metadata('edge_follow', 'count')
 
     @property
+    def bio_links(self) -> List[str]:
+        """list of external links given in the bio"""
+        return [links['url'] for links in self._metadata('bio_links')] if self._metadata('bio_links') else []
+
+    @property
     def external_url(self) -> Optional[str]:
         return self._metadata('external_url')
 


instaloader/structures.py
```diff
@@ -936,6 +936,11 @@ def followers(self) -> int:
     def followees(self) -> int:
         return self._metadata('edge_follow', 'count')
 
+    @property
+    def bio_links(self) -> List[str]:
+        """list of external links given in the bio"""
```
aandergr, line 941: Please also add a `versionadded` directive here, indicating it is new in version 4.12.

instaloader/structures.py
```diff
@@ -936,6 +936,11 @@ def followers(self) -> int:
     def followees(self) -> int:
         return self._metadata('edge_follow', 'count')
 
+    @property
+    def bio_links(self) -> List[str]:
```
aandergr, line 940: It seems the `bio_links` structure provides much more than just the URLs, e.g. a `title` property that might be interesting as well. I would prefer this method to also return the other information.

```json
[
  {
    "title": "",
    "url": "http://taylorswift.com",
    "link_type": "external"
  },
  {
    "title": "THE TORTURED POETS DEPARTMENT 🤍",
    "url": "https://taylor.lnk.to/thetorturedpoetsdepartment",
    "link_type": "external"
  },
  {
    "title": "Taylor Swift | The Eras Tour concert film",
    "url": "https://taylor.lnk.to/TSTheErasTourFilm",
    "link_type": "external"
  }
]
```

instaloader/structures.py
```diff
@@ -936,6 +936,11 @@ def followers(self) -> int:
...
+        """list of external links given in the bio"""
+        return [links['url'] for links in self._metadata('bio_links')] if self._metadata('bio_links') else []
```
aandergr, line 942: I do not understand the purpose of `if self._metadata('bio_links')` here. If it evaluates to false due to an empty list, the left-hand statement would iterate an empty list and return an empty list, without that it needs to be explicitly checked.

Issue #1085, opened 2021-04-09
Instaloader.download_post(): ZeroDivisionError: integer division or modulo by zero


skrull:
version: 4.7.1
command line: instaloader --commit-mode --login=(stripped) --sessionfile session --no-posts --abort-on=302,400,429 --tagged user

[ 30] Traceback (most recent call last):
  File "pvenv/bin/instaloader", line 8, in <module>
    sys.exit(main())
  File "pvenv/lib/python3.8/site-packages/instaloader/__main__.py", line 445, in main
    _main(loader,
  File "pvenv/lib/python3.8/site-packages/instaloader/__main__.py", line 211, in _main
    instaloader.download_profiles(profiles,
  File "pvenv/lib/python3.8/site-packages/instaloader/instaloader.py", line 1274, in download_profiles
    self.download_tagged(profile, fast_update=fast_update, post_filter=post_filter)
  File "pvenv/lib/python3.8/site-packages/instaloader/instaloader.py", line 1119, in download_tagged
    self.posts_download_loop(profile.get_tagged_posts(),
  File "pvenv/lib/python3.8/site-packages/instaloader/instaloader.py", line 922, in posts_download_loop
    downloaded = self.download_post(post, target=target)
  File "pvenv/lib/python3.8/site-packages/instaloader/instaloader.py", line 618, in download_post
    for i in range(self.slide_start % post.mediacount, self.slide_end % post.mediacount + 1)),
ZeroDivisionError: integer division or modulo by zero

Deleora:
Just curious, what's the point of using both `--login` and `--sessionfile`?

skrull:
Nothing really important, but it just asks the password if the session file is missing.

aandergr:
How is this reproducible?

skrull:
There you go.

https://pastebin.com/tf4Hy24b

aandergr:
Thanks. Also reproducible with: `instaloader -- -CNLdFNkBuM1` (downloading the single post that fails)

aandergr:
Mentioned post does not exist anymore. Is there any other post for which this error happens?

skrull:
@aandergr sorry about the delay. Got it again yesterday. 

https://pastebin.com/nn8brAaG

89z:
https://github.com/instaloader/instaloader/issues/1346

QrchackOfficial:
[...]

PR #2214, opened 2024-03-15
Fix for issue #1085

SeamusMullan:
Implement check for post.mediacount before checking if posts have been downloaded which caused crashes when attempting to download posts within sidecars.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -718,36 +718,37 @@ class Instaloader:
         downloaded = True
         if post.typename == 'GraphSidecar':
             if self.download_pictures or self.download_videos:
-                if not _all_already_downloaded(
-                        filename_template, enumerate(
-                            (post.get_is_videos()[i]
-                             for i in range(self.slide_start % post.mediacount, self.slide_end % post.mediacount + 1)),
-                            start=self.slide_start % post.mediacount + 1
-                        )
-                ):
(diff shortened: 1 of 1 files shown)


instaloader/instaloader.py
```diff
@@ -718,36 +718,37 @@ def _all_already_downloaded(path_base, is_videos_enumerated) -> bool:
...
-                        filename_template, enumerate(
-                            (post.get_is_videos()[i]
-                             for i in range(self.slide_start % post.mediacount, self.slide_end % post.mediacount + 1)),
-                            start=self.slide_start % post.mediacount + 1
-                        )
-                ):
-                    for edge_number, sidecar_node in enumerate(
-                            post.get_sidecar_nodes(self.slide_start, self.slide_end),
-                            start=self.slide_start % post.mediacount + 1
+                if not (post.mediacount) == 0: # make sure post hasnt been edited / removed
```
aandergr, line 721: Thanks for fixing this issue!

I think it would be nice to group these two ifs together, to avoid nesting and increase readability, like so

```python
            if (self.download_pictures or self.download_videos) and post.mediacount > 0:
```

SeamusMullan:
bump.

aandergr:
This change now contains three whitespace changes which are unrelated to the change that the PR is supposed to introduce. Once these are removed, we can merge this PR :)

PR #2221, opened 2024-03-23
Preserve session.verify for temporary sessions

Hummer12007:
Prevents SSL verification errors when using proxies with self-signed certificates


aandergr:
Currently the session object cannot be cleanly set from the outside, so this is a fix for a currently-dirty use case.

I am not against allowing to set verify=false or providing a custom way to instantiate and copy the session object, however it should be cleanly implemented to enable us to merge it into the Instaloader source tree.

PR #2243, opened 2024-05-05
Use different exit codes for different failures

### LINKED ISSUE #2220: Different exit codes for different failure types
**Provide us a use case of the feature**
I believe instaloader should exit with a non-zero exit code in case of failure. This has been requested already (#1614) and there is even a PR for this (#2166). However, I believe that not only a non-zero exit code should be used, but different codes should be used for different failures.

**Describe the solution you'd like**
I thought of something like
0 - Finished normally
1 - Initialization failure: invalid command line arguments, --load-cookies without the library, and other failures before anything was attempted
2 - Login failure
3 - Failure while downloading
4 - User aborted (I'm not sure if this is an error - after all, the user requested the action. But it might be useful to distinguish this case)

Maybe there could be a couple more situations. But too much granularity might make this feature less useful.

**Describe alternatives you've considered**
There really aren't many alternatives. One can try parsing the output, but there are too many possible error messages.

**If the feature request is accepted, would you be willing to submit a pull request?**
Yes

ekalin:
Fixes #2220 

- [X] Initialization errors
- [X] Login errors
- [x] Non-fatal error while downloading
- [x] --abort-on
- [x] SIGINT / Control-C
- [x] Documentation

Here's the general idea of how each situation is handled:
* Non fatal error (1): Before exiting, it is checked if something was logged to be repeated at the end (those are the real warning/erros that require attention). If so, the exit code is 1.
* Initialization failure (2): returned by argparse for invalid options and other parsing errors; other errors that instaloader itself detects raise `InvalidArgumentException`, which is caught at `main()`.
  - `InvalidArgumentException` is also used for other errors, that generally can only be triggered when used as a module. I don't think that's a problem, but another exception could be used specifically for those initialization errors.
* Login failure (3): raises the new `LoginException`. Some more specific login exceptions were made into subclasses.
* Download aborted (4) and user aborted (5): these were already caught at `_main`, it just sets a flag with the status to be returned.

Some notes:
* I created a new exit code 99 for any uncaught InstaloaderException (at the very end of `__main__.py`). But maybe that `except` is not necessary: hopefully no exception should reach there, and if it does, maybe letting the program crash (and include the stack trace) is more appropriate.
* I've made the failure when no IG cookies are found an actual failure. It used to just print a message and carry on (anonymously)
* When nothing is done (no targets are specified), I'm returning non-fatal error, but initialization failure could also be appropriate, especially when the user was not logged in. But having two different status codes for two very similar situations is probably too confusing, so I used non-fatal error for both cases.

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -320,7 +320,7 @@ def _main(instaloader: Instaloader, targetlist: List[str],
         else:
             # Instaloader did not do anything
             instaloader.context.log("usage:" + usage_string())
-        exit_code = ExitCode.NON_FATAL_ERROR
+            exit_code = ExitCode.INIT_FAILURE
     return exit_code
 
 
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -11,7 +11,7 @@ from enum import IntEnum
 from typing import List, Optional
 
 from . import (AbortDownloadException, BadCredentialsException, Instaloader, InstaloaderException,
-               InvalidArgumentException, Post, Profile, ProfileNotExistsException, StoryItem,
+               InvalidArgumentException, LoginException, Post, Profile, ProfileNotExistsException, StoryItem,
                TwoFactorAuthRequiredException, __version__, load_structure_from_file)
 from .instaloader import (get_default_session_filename, get_default_stamps_filename)
 from .instaloadercontext import default_user_agent
@@ -106,7 +106,7 @@ def get_cookies_from_instagram(domain, browser, cookie_file='', cookie_name=''):
     if cookies:
         print(f"Cookies loaded successfully from {browser}")
     else:
-        print(f"No cookies found for Instagram in {browser}, Are you logged in succesfully in {browser}?")
+        raise LoginException(f"No cookies found for Instagram in {browser}, Are you logged in succesfully in {browser}?")
 
     if cookie_name:
         return cookies.get(cookie_name, {})
@@ -120,7 +120,7 @@ def import_session(browser, instaloader, cookiefile):
         instaloader.context.update_cookies(cookie)
         username = instaloader.test_login()
         if not username:
-            raise SystemExit(f"Not logged in. Are you logged in successfully in {browser}?")
+            raise LoginException(f"Not logged in. Are you logged in successfully in {browser}?")
         instaloader.context.username = username
         print(f"{username} has been successfully logged in.")
         next_step_text = (f"Next: Run instaloader --login={username} as it is required to download high quality media "
@@ -568,6 +568,9 @@ def main():
     except InvalidArgumentException as err:
         print(err, file=sys.stderr)
         raise SystemExit(ErrorCodes.INIT_FAILURE)
+    except LoginException as err:
+        print(err, file=sys.stderr)
+        raise SystemExit(ErrorCodes.LOGIN_FAILURE)
     except InstaloaderException as err:
         print("Fatal error: %s" % err)
         raise SystemExit(ErrorCodes.UNEXPECTED_ERROR)
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -23,12 +23,13 @@ except ImportError:
     bc3_library = False
 
 
-class ErrorCodes(IntEnum):
+class ExitCodes(IntEnum):
+    SUCESS = 0
     DOWNLOAD_FAILURE = 1
     INIT_FAILURE = 2
     LOGIN_FAILURE = 3
-    HTTP_STATUS_ABORT = 4
-    USER_ABORT = 5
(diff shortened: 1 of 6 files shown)


ekalin:
I believe this is now ready for review.

instaloader/__main__.py
```diff
@@ -22,6 +23,15 @@
     bc3_library = False
 
 
+class ExitCodes(IntEnum):
+    SUCESS = 0
```
aandergr, line 27:
```suggestion
    SUCCESS = 0
```

(typo)

instaloader/__main__.py
```diff
@@ -133,7 +143,7 @@ def _main(instaloader: Instaloader, targetlist: List[str],
           max_count: Optional[int] = None, post_filter_str: Optional[str] = None,
           storyitem_filter_str: Optional[str] = None,
           browser: Optional[str] = None,
-          cookiefile: Optional[str] = None) -> None:
+          cookiefile: Optional[str] = None) -> int:
```
aandergr, line 146: It seems we could be even more explicit regarding the typing here:

```suggestion
          cookiefile: Optional[str] = None) -> ExitCodes:
```

instaloader/__main__.py
```diff
@@ -22,6 +23,15 @@
     bc3_library = False
 
 
+class ExitCodes(IntEnum):
```
aandergr, line 26: It is mainly a matter of preference, and I'm open-minded, but I have the feeling it is a little more common to name enum types in the singular (i.e. ExitCode here) than in the plural. The Python enum examples (https://docs.python.org/3/library/enum.html) also do it this way.

instaloader/instaloader.py
```diff
@@ -1605,3 +1615,9 @@ def interactive_login(self, username: str) -> None:
                 except BadCredentialsException as err:
                     print(err, file=sys.stderr)
                     pass
+
+    def has_stored_errors(self) -> bool:
```
aandergr, line 1619: I think this should be a property (and the method in InstaloaderContext as well)

```suggestion
    @property
    def has_stored_errors(self) -> bool:
```

instaloader/instaloader.py
```diff
@@ -643,11 +643,16 @@ def test_login(self) -> Optional[str]:
     def login(self, user: str, passwd: str) -> None:
         """Log in to instagram with given username and password and internally store session object.
 
-        :raises InvalidArgumentException: If the provided username does not exist.
         :raises BadCredentialsException: If the provided password is wrong.
-        :raises ConnectionException: If connection to Instagram failed.
         :raises TwoFactorAuthRequiredException: First step of 2FA login done, now call
-           :meth:`Instaloader.two_factor_login`."""
+           :meth:`Instaloader.two_factor_login`.
+        :raises LoginException: An error happened during login (for example, and invalid response).
```
aandergr, line 649: I do not understand the phrase `for example, and invalid response`. Probably a typo?

aandergr (commented):
This is an amazing change, it seems well crafted, makes our error handling cleaner and was a pleasure to go through. Thanks! I have only a few minor suggestions/remarks and then this can be merged :)

instaloader/__main__.py
```diff
@@ -307,6 +320,8 @@ def _main(instaloader: Instaloader, targetlist: List[str],
         else:
             # Instaloader did not do anything
             instaloader.context.log("usage:" + usage_string())
+        exit_code = ExitCodes.NON_FATAL_ERROR
```
aandergr, line 323: This leads to a NON_FATAL_ERROR exit code when using Instaloader just for logging in or importing session cookies, which I think is a valid and non-error use case. I think the exit_code should be only set in the `else`, if Instaloader was not used for logging in and the usage was printed.

ekalin: This is one case in which I wasn't sure what was the best way to handle (see the last note in the main text of the PR).

So I guess in the else branch (wasn't logged in) an error should be returned (in this case I think initialization failure is more appropriate), but if it was logged in then success is returned?

instaloader/__main__.py
ekalin: I believe this is also a convention in other languages, and it does read better, especially when it's the return type of a function or when one of the values is used.

instaloader/instaloader.py
ekalin: It was a typo, it should have been 'an invalid responsed'. I've fixed the typo and improved a bit the wording.

instaloader/__main__.py
aandergr: I think calling Instaloader only for creating a session file is a legit non-error use case if successful. When calling Instaloader without a target, only with a —login or —load-cookies option the exit code should indicate success if succesful, and LOGIN_FAILURE otherwise.

In the other case, INIT_FAILURE is probably the best exit code.

PR #2258, opened 2024-05-12
Add support for more browsers

### LINKED ISSUE #2241: vivaldi support cookies
browser_cookie3 support vivaldi cookie stuff add please

ekalin:
Fixes #2241.

Also adds other browsers that the library supports but were not included in instaloader.

I don't have Vivaldi or those browsers so I haven't actually tested that the cookies are loaded, but this would be a test more of the library than instaloader itself, since all the work is done by the library.

  - Do you consider it ready to be merged or is it a draft? Should be ready
  - Can we help you at some point?  It's a simple change and should be ready, but I'm open to any comments

### DIFF
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -202,7 +202,7 @@ Instaloader to login.
    This feature requires the browser_cookie3 library.
    Compatible with :option:`--cookiefile` if you want to load cookies from browser profiles.
    Incompatible with :option:`--login` due to potential username mismatch between user input and browser login.
-   Supported browsers: Chrome, Chromium, Firefox, Edge, Brave, Opera, Opera GX, Safari, Vivaldi and Librewolf.
+   Supported browsers: Brave, Chrome, Chromium, Edge, Firefox, LibreWolf, Opera, Opera_GX, Safari and Vivaldi.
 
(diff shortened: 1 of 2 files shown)


docs/cli-options.rst
```diff
@@ -202,7 +202,7 @@ Instaloader to login.
    This feature requires the browser_cookie3 library.
    Compatible with :option:`--cookiefile` if you want to load cookies from browser profiles.
    Incompatible with :option:`--login` due to potential username mismatch between user input and browser login.
-   Supported browsers: Chrome, Firefox, Edge, Opera, Safari, Brave.
+   Supported browsers: Chrome, Chromium, Firefox, Edge, Brave, Opera, Opera GX, Safari, Vivaldi and Librewolf.
```
aandergr, line 205: I think we should sort the browsers somehow, e.g. alphabetically.

It is probably unclear that for Opera GX, 'opera_gx` must be specified as value.

ekalin: Makes sense. I've ordered the browsers in the `supported_browsers` variable as well for consistency.

I've also changed it to "Opera_GX".

aandergr: Nice, thanks!

PR #2260, opened 2024-05-13
Update actions versions

tinyboxvk:
Fix GitHub Actions Node.js deprecation warnings

`The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, actions/setup-python@v2.`

`Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v2, actions/setup-python@v2.`

This PR bumps actions/checkout to v4, and actions/setup-python to v5.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -8,11 +8,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: "Checkout Instaloader Repository"
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0  # needed for building docs
       - name: "Setup Python"
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
           python-version: 3.12
       - name: "Install Dependencies"
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -15,6 +15,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: 3.12
+          cache: 'pip'
       - name: "Install Dependencies"
         run: |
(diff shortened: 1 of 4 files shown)


aandergr:
Pretty nice that you addressed the deprecation warnings in the actions. I think one should also take the opportunity to adjust the versions in the other actions, not only in the Windows EXE build action.

I have also seen that setup-python provides a cache for the dependencies, which could save a few seconds for the build. Would you perhaps like to take the opportunity to implement this as well?

.github/workflows/documentation.yml
```diff
@@ -8,13 +8,14 @@ jobs:
...
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0  # needed for building docs
       - name: "Setup Python"
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
           python-version: 3.12
+          cache: 'pip'
```
aandergr, line 18:
```suggestion
          cache: 'pipenv'
```

We actually use Pipenv for dependency management, Pip is only used in the Workflows to install Pipenv.

tinyboxvk: Thanks! I wasn't quite sure which one to use. Does that apply to "packages.yml" as well?

aandergr: Thanks! Good point, "packages.yml" does acutally not install the Instaloader dependencies, it just installs the dependencies required for the packaging, and does so with Pip. However, the pip cache mechanism seems to be using the requirements.txt file which we do not have, so for packages.yml the cache should better be disabled.

aandergr: Looks good, thank you very much!

PR #2263, opened 2024-05-14
Download stories when --load-cookies is used

### LINKED ISSUE #2259: Does not download stories when --load-cookies is passed
**Describe the bug**
When `--load-cookies` is passed, stories are not downloaded, even if the loaded cookies provided a valid session (and the user is considered logged in).

**To Reproduce**
Log in using a browser
```bash
instaloder --stories --load-cookies firefox taylorswift
```

**Expected behavior**
Posts and stories should be downloaded.

No error is reported, and posts are downloaded, but there's a warning at the start:
```
--login=USERNAME required to download stories.
```

**Instaloader version**
Using upcoming/4.12 branch.

**Additional context**
`--load-cookies` is an anternative to `--login`, so it should be considered when checking if stories are downloaded.
~There might be other places where a similar problem happens.~

ekalin:
Fixes #2259.

Stories could only be downloaded if `--login` was used. I changed this to either `--login` or `--load-cookies`, which also logs the user in.

Otherwise the user needed to run Instaloader first just to import the cookies, and a second time (using the saved session) to actually download stories.

I've also changed error messages and the documentation in several places to refer to the process of logging in rather than to `--login` directly, but if you think that is too confusing, feel free to discard that commit.

  - Can we help you at some point? There's nothing I need help for, but I'm open to suggestions.

### DIFF
--- a/docs/basic-usage.rst
+++ b/docs/basic-usage.rst
@@ -75,7 +75,7 @@ What to Download
 Instaloader supports the following targets:
 
 - ``profile``
-   Public profile, or private profile with :option:`--login`.
+   Public profile, or private profile with :ref:`login<login>`.
 
    If an already-downloaded profile has been renamed, Instaloader automatically
    finds it by its unique ID and renames the folder accordingly.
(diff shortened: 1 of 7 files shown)


aandergr:
Thanks! 

I also like the changes to the documentation and the wordings in output messages. To me it seems they make the program easier to understand and more user-friendly.

PR #2270, opened 2024-05-20
Add video_play_count property

### LINKED ISSUE #2145: Add a funtion to get the count of plays from Instagram Reels
**use case of the feature**

- **How could the user invoke the new function? ->** Just calling "reels_plays_count" function.
- **Which problem would it solve? ->** With the massive amount of Reels being created in Instagram, a function to get it's plays count (it's different from the current _video_views_count_ function) would be amazing cause it would provide a way to perform analysis on this famous type of content. Your tool would be even more utilized.
- **If new information is obtained, how would it be further processed? ->** Just by running the yet to be created  "reels_plays_count" function

**solution I'd like**
The creation of a function that would allow me to proper get the plays count of a Instagram Reels video.

**Describe alternatives you've considered**
I've considered using the current video_views_count function but it doesn't return the exact same value displayed in instagram insights. Example below:

- **Plays from instagram app:**
 image

- **Count from _video_view_count_ function:**
image

**If the feature request is accepted, would you be willing to submit a pull request?**
Yes. I would need assistance.

**### I looooooooved the instaloader tool, it's amazing and so simple to use. You guys rock!!!**

steven5538:
Hello 👋 
I found this field might be helpful for video metrics, so add it to the `Post`.
It's different than "View" as "View" might have multiple "Play" for video.

I just realized this could also close #2145 .

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -240,6 +240,7 @@ class Post:
             fake_node["video_url"] = media['video_versions'][-1]['url']
             fake_node["video_duration"] = media["video_duration"]
             fake_node["video_view_count"] = media["view_count"]
+            fake_node["video_play_count"] = media["play_count"]
         with suppress(KeyError):
             fake_node["edge_sidecar_to_children"] = {"edges": [{"node": {
                 "display_url": node['image_versions2']['candidates'][0]['url'],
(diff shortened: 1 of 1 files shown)


aandergr:
Hi. It seems that for a video Post constructed with `Post.from_shortcode()` it fails with a `KeyError: 'video_play_count'`. Does it only work for Posts build with from_iphone_struct?

manpreetsingh78:
> Hi. It seems that for a video Post constructed with `Post.from_shortcode()` it fails with a `KeyError: 'video_play_count'`. Does it only work for Posts build with from_iphone_struct?

same issue i'm having even with iphone struct

manpreetsingh78:
How can i use from_iphone_struct(cls, context: InstaloaderContext, media: Dict[str, Any])
its taking _media as a parameter_ from **where to obtain this**
**an example will be very helpful**

Thanks in advance

cnc4ever:
@steven5538 i also get the same error when using it with posts from `Profile.get_posts()`, do you have any idea how to do this?
If this works it would mean so much for me thank you!

PR #2271, opened 2024-05-20
fix: Add max_count to limit profile post downloads

### LINKED ISSUE #2269: get_posts() with limit
Hi. Make the **get_posts()** method accept the limit on the received data as an argument. There are pages with more than 100 thousand posts and it took me about an hour to do this.
I would like to get only the last n posts instead of all the posts by sending an argument. 
`posts = user.get_posts(50); #latest 50 posts only :)`

Mesheo:
## Motivation for this change
  - Fixes #2269 
  
👋 Hi! I made this PR to make it possible to **control the quantity of posts downloaded**. 
The core method `self.posts_download_loop()` already has this functionality, but the external calling functions like `instaloader.download_profiles()` were not adjusted to receive or pass this argument further.

## The completeness of this change
  * is it just a proof of concept? **No**
  * Is the documentation updated (if appropriate)? **Yes**
  * Do you consider it ready to be merged or is it a draft? **Yes**
  * Can we help you at some point? **I dont think so... but its my first PR so let me know anything**

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -8,13 +8,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: "Checkout Instaloader Repository"
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0  # needed for building docs
       - name: "Setup Python"
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
           python-version: 3.12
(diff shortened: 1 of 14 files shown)


aandergr:
This is not a fix, and as such should be based on `upcoming/v4.12`. The docstring comment should then be amended by a `versionchanged` directive documenting that the feature has been added in v4.12. The parameter description for the `max_count` parameter mentions the `--count` option, however, `--count` does not (and must not) affect `max_count` for `download_profiles()`.

PR #2288, opened 2024-06-21
Fix. Return profile_pic_url for non-iphone request

kzuahelp:
There is no profile_pic_url_hd in json response of web interface and no way to get TopSearchResults data in single request with profile picture links.

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -1160,7 +1160,7 @@ class Profile:
         """Return URL of lower-quality profile picture.
 
         .. versionadded:: 4.9.3"""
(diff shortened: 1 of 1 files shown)


aandergr:
We need more information about the motivation of this PR to be able to decide on this. Which problem does it solve? 

> There is no profile_pic_url_hd in json response of web interface

I cannot verify this statement. I just retrieved this field for some profiles and it works, it is included in the JSON. Further, there were no bug reports that the field is missing in the response.

Also, this PR does not replace all accesses of profile_pic_url_hd.

PR #2290, opened 2024-06-23
Fix location endpoint URL

### LINKED ISSUE #2212: Impossible to retrieve Location info | Especially Latitude and Longitude
**Describe the bug**
Impossible to retrieve Location info, especially Latitude and Longitude

**To Reproduce**
```
for post in profile.get_posts():
    post.location
```

**Expected behavior**
I expect a location info object

JSON Query to explore/locations/815623301860707/: HTTP error code 201. [retrying; skip with ^C]
JSON Query to explore/locations/815623301860707/: HTTP error code 201. [retrying; skip with ^C]
JSON Query to explore/locations/815623301860707/: HTTP error code 201.

**Instaloader version**
instaloader==4.10.3

**Additional context**
If i do the following, I get some location info, but no latitude and longitude

```
post._node['location']
{'id': '26058259', 'has_public_page': True, 'name': 'Forestis Dolomites', 'slug': 'forestis-dolomites'}

```

### LINKED ISSUE #2215: Location ID syntax fails with HTTP code 201
**Describe the bug**
The `%location id` syntax appears broken, and results in a 201 response from Instagram

**To Reproduce**

`instaloader --login {username} %109570449061083`

**Expected behavior**
Instaload should run successfully and return posts associated with that location id.

```
Logged in as {username}.
Retrieving pictures for location 109570449061083...
JSON Query to explore/locations/109570449061083/: HTTP error code 201. [retrying; skip with ^C]
JSON Query to explore/locations/109570449061083/: HTTP error code 201. [retrying; skip with ^C]
%109570449061083: JSON Query to explore/locations/109570449061083/: HTTP error code 201.
```

**Instaloader version**
4.10.3

**Additional context**

A get request to `https://www.instagram.com/explore/locations/109570449061083/?__a=1&__d=dis` gives a 201 error, so I would guess that this issue stems from some API change interacting with the query code: 

https://github.com/instaloader/instaloader/blob/5509ea7daf7f5e81534d4eb996c8afd478704426/instaloader/instaloader.py#L1155-L1160

and

https://github.com/instaloader/instaloader/blob/5509ea7daf7f5e81534d4eb996c8afd478704426/instaloader/sectioniterator.py#L29-L33

This is likely the same underlying problem as #2212 , but is accessed in a different way (using the `location id` command-line argument)

This problem also seems to be affecting other projects using the Instagram API https://github.com/subzeroid/instagrapi/issues/1802

felipeadeildo:
The motivation is to solve the problem that ive been received on get post location information. The endpoint apparently has been changed from `explore/locations/<location_id>` to `api/v1/locations/web_info` with the query string `location_id` and some others default.

So, I changed the endpoint and the params, just it.

  - Is it just a proof of concept? Not.
  - Is the documentation updated (if appropriate)? Not.
  - Do you consider it ready to be merged or is it a draft? Yes, i think so, I'm using the version that i chaged locally.
  - Can we help you at some point? I don't think so
  
 Thank you very much!

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -809,9 +809,12 @@ class Post:
             return None
         location_id = int(loc['id'])
         if any(k not in loc for k in ('name', 'slug', 'has_public_page', 'lat', 'lng')):
-            loc.update(self._context.get_iphone_json("api/v1/locations/web_info/,
-                                                params={'location_id': location_id, 'show_nearby': 'false', 'skip_recent_tab': 'false'}
-                                            )['native_location_data']['location_info'])
+            loc.update(self._context.get_iphone_json("api/v1/locations/web_info/",
+                                                params={
+                                                    'location_id': location_id, 
+                                                    'show_nearby': 'false', 
+                                                    'skip_recent_tab': 'false'
+                                                })['native_location_data']['location_info'])
         self._location = PostLocation(location_id, loc['name'], loc['slug'], loc['has_public_page'],
                                       loc.get('lat'), loc.get('lng'))
         return self._location
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -809,8 +809,9 @@ class Post:
             return None
         location_id = int(loc['id'])
         if any(k not in loc for k in ('name', 'slug', 'has_public_page', 'lat', 'lng')):
-            loc.update(self._context.get_json("explore/locations/{0}/".format(location_id),
-                                              params={'__a': 1, '__d': 'dis'})['native_location_data']['location_info'])
+            loc.update(self._context.get_json("api/v1/locations/web_info,
+                                                params={'location_id': location_id, 'show_nearby': 'false', 'skip_recent_tab': 'false'}
+                                            )['native_location_data']['location_info'])
         self._location = PostLocation(location_id, loc['name'], loc['slug'], loc['has_public_page'],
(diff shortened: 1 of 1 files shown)


felipeadeildo (commented):
I forgot to end the string with `"` 😅

aandergr:
The change only fixes the problem in the sense that an error message is no longer thrown, as the previous request that did not work is no longer made. The query was used to retrieve latitude and longitude; the other fields, `name`, `slug`, and `has_public_page`, are already set in `loc`. The new request contains some additional information (which is however not used), but also no latitude and longitude, so unfortunately #2212 is not fixed.

#2215 is not about querying the location of a post, but about querying all posts of a location. The method changed here is not used at all. This PR is not related to this bug.

felipeadeildo:
> The change only fixes the problem in the sense that an error message is no longer thrown, as the previous request that did not work is no longer made. The query was used to retrieve latitude and longitude; the other fields, `name`, `slug`, and `has_public_page`, are already set in `loc`. The new request contains some additional information (which is however not used), but also no latitude and longitude, so unfortunately #2212 is not fixed.
> 
> #2215 is not about querying the location of a post, but about querying all posts of a location. The method changed here is not used at all. This PR is not related to this bug.

OK! Got it, I've removed the mentions of the issues!

aandergr:
The query that is now done does not add any information, this PR just adds an unnecessary query.

Issue #2285, opened 2024-06-15
Unable download Highlights and IGTV with the KeyError "edge_felix_video_timeline"


arthur-lim:
**Describe the bug**
Unable to download highlights and IGTV with the error

**To Reproduce**
Running the command `instaloader username --highlights --igtv --load-cookies=firefox`

**Expected behavior**
Able to download Highlights and IGTV of that profile

**Error messages and tracebacks**

```
Retrieving IGTV videos for profile wanna._b.
Traceback (most recent call last):
  File "C:\Users\ARTHUR\AppData\Roaming\Python\Python312\site-packages\instaloader\structures.py", line 969, in _metadata
    d = d[key]
KeyError: 'edge_felix_video_timeline'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\ARTHUR\AppData\Roaming\Python\Python312\Scripts\instaloader.exe\__main__.py", line 7, in <module>
  File "C:\Users\ARTHUR\AppData\Roaming\Python\Python312\site-packages\instaloader\__main__.py", line 541, in main
    _main(loader,
  File "C:\Users\ARTHUR\AppData\Roaming\Python\Python312\site-packages\instaloader\__main__.py", line 283, in _main
    instaloader.download_profiles(profiles,
  File "C:\Users\ARTHUR\AppData\Roaming\Python\Python312\site-packages\instaloader\instaloader.py", line 1478, in download_profiles
    self.download_igtv(profile, fast_update=fast_update, post_filter=post_filter,
  File "C:\Users\ARTHUR\AppData\Roaming\Python\Python312\site-packages\instaloader\instaloader.py", line 1292, in download_igtv
    igtv_posts = profile.get_igtv_posts()
                 ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ARTHUR\AppData\Roaming\Python\Python312\site-packages\instaloader\structures.py", line 1232, in get_igtv_posts
    self._metadata('edge_felix_video_timeline'),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ARTHUR\AppData\Roaming\Python\Python312\site-packages\instaloader\structures.py", line 975, in _metadata
    d = d[key]
        ~^^^^^
[...]

PR #2294, opened 2024-06-30
add --reels (TERRIBLE SOLUTION)

jpfleischer:
This adds the --reels option but you must have gallery-dl installed with `pip install -U gallery-dl` and ideally use the browser cookies option

WORKING:
`instaloader --reels myprofile --load-cookies firefox`

The documentation is not updated, and it is not a proof of concept because it is working :)

You can help me implement an actual working reels downloader by explaining how I can print the response returned by IG. edge_felix_video_timeline is some metadata that generates a key error https://github.com/instaloader/instaloader/issues/2285 and I have no idea what i should be looking for.

My solution only downloads the reels itself (mp4) and it does not have statistics such as comments or amount of plays.

```bash
git clone https://github.com/jpfleischer/instaloader.git
cd instaloader
pip install -e .
instaloader --reels myprofile --load-cookies firefox
```

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -140,6 +140,7 @@ def _main(instaloader: Instaloader, targetlist: List[str],
           download_highlights: bool = False,
           download_tagged: bool = False,
           download_igtv: bool = False,
+          download_reels: bool = False,
           fast_update: bool = False,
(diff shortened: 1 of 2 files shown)


aandergr:
Reels were implemented with #2355.

PR #2295, opened 2024-06-30
Context factory and session factory

Henkhogan:
Currently there is no way to customize the instantiation of session objects. I would need to use a dedicated proxy for the connections to insta.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -7,6 +7,7 @@ import shutil
 import string
 import sys
 import tempfile
+import warnings
 from contextlib import contextmanager, suppress
 from datetime import datetime, timezone
 from functools import wraps
@@ -192,8 +193,8 @@ def instaloader_context_factory(no_sleep: bool = False,
     return InstaloaderContext(not no_sleep, quiet, user_agent, max_connection_attempts,request_timeout, rate_controller, fatal_status_codes, iphone_support)
 
 
-class Instaloader:
-    """Instaloader Class.
+class InstaloaderBase:
+    """InstaloaderBase Class.
 
     :param dirname_pattern: :option:`--dirname-pattern`, default is ``{target}``
     :param filename_pattern: :option:`--filename-pattern`, default is ``{date_utc}_UTC``
@@ -297,10 +298,7 @@ class Instaloader:
     @contextmanager
     def anonymous_copy(self):
         """Yield an anonymous, otherwise equally-configured copy of an Instaloader instance; Then copy its error log."""
-        new_loader = Instaloader(
-            sleep=self.context.sleep,
-            quiet=self.context.quiet,
-            user_agent=self.context.user_agent,
+        new_loader = InstaloaderBase(
             dirname_pattern=self.dirname_pattern,
             filename_pattern=self.filename_pattern,
             download_pictures=self.download_pictures,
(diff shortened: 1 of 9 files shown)


Henkhogan:
waiting for workflows to be run

Henkhogan:
@aandergr would you please approve the workflow run?

instaloader/instaloader.py
```diff
@@ -223,20 +233,15 @@ def __init__(self,
...
                  resume_prefix: Optional[str] = "iterator",
                  check_resume_bbd: bool = True,
                  slide: Optional[str] = None,
-                 fatal_status_codes: Optional[List[int]] = None,
-                 iphone_support: bool = True,
                  title_pattern: Optional[str] = None,
-                 sanitize_paths: bool = False):
+                 sanitize_paths: bool = False,
+                 context: Optional[InstaloaderContext] = None
+                 ):
```
aandergr, line 242: We cannot remove parameters from the Instaloader constructor, this would potentially break existing code that uses Instaloader.

liponex: If `context` is passed in, we can simply ignore any removed parts of `__init__`.
But since there's a factory there anyway, just use them in there. This will add backward compatibility.

```python
def __init__(self,
             sleep: bool = True,
             quiet: bool = False,
             user_agent: Optional[str] = None,
             dirname_pattern: Optional[str] = None,
             filename_pattern: Optional[str] = None,
             download_pictures=True,
             download_videos: bool = True,
             download_video_thumbnails: bool = True,
             download_geotags: bool = False,
             download_comments: bool = False,
             save_metadata: bool = True,
             compress_json: bool = True,
             post_metadata_txt_pattern: Optional[str] = None,
             storyitem_metadata_txt_pattern: Optional[str] = None,
             max_connection_attempts: int = 3,
             request_timeout: float = 300.0,
             rate_controller: Optional[Callable[[InstaloaderContext], RateController]] = None,
             resume_prefix: Optional[str] = "iterator",
             check_resume_bbd: bool = True,
             slide: Optional[str] = None,
             fatal_status_codes: Optional[List[int]] = None,
             iphone_support: bool = True,
             title_pattern: Optional[str] = None,
             sanitize_paths: bool = False,
             context: Optional[InstaloaderContext] = None
             ):
        self.context = context or instaloader_context_factory(not sleep, quiet, user_agent,
                 max_connection_attempts, request_timeout, rate_controller,
                 fatal_status_codes, iphone_support)
        ...
```

liponex: @Henkhogan, I'm playing with your fork now. Firstly I've thought, that would be easier to pass `Callable` for `session_factory`. But context factory much greater, since it gives rights to rewrite used user-agent, cookies and logic of requests.

Henkhogan: My idea is to create a new base class that Instaloader inherits from. That will allow to not break anything while the new base class can be used by whoever needs more control of the context

Henkhogan:
@aandergr  PR updated to retain backwards compatibility -> please allow workflows to run

instaloader/instaloader.py
aandergr: Another approach that would allow a clean extension of how the sessions are handled would be to refactor the requests.Session handling that we have in instaloadercontext.py into an own session class, extending requests.Session, and having a parameter for the session factory, similar as it is done with the rate controller (https://instaloader.github.io/module/instaloadercontext.html#instaloader.RateController). This provides an easily-overridable default.

aandergr:
We will not mark Instaloader deprecated and switch to a new main class.

The workflows can be run locally, we use Pipenv for dependency pinning so the exact versions that are used in the workflows can be easily installed in a virtual environment. The commands that are executed are described in `.github/workflows`.

instaloader/instaloader.py
```diff
@@ -7,6 +7,7 @@
 import string
 import sys
 import tempfile
+import warnings
```
aandergr, line 10: Seems to be unused.

Henkhogan:
> We will not mark Instaloader deprecated and switch to a new main class.

Please let me know why. 
I see design flaws with this class that I tried to eliminate while keeping backwards compatibility.
If you don't deprecate it, you'll always have to cope with impacts of these flaws -> for long term maintenance of this project, that will cause headaches

aandergr:
It's important to understand that deprecating a core class has significant implications, even if backwards compatibility is maintained. Deprecation signals to users that they should begin transitioning to a new class, which can introduce confusion and a perception that the current design is fundamentally flawed or unstable. This is not a message we want to send unless there is a compelling, unavoidable reason to do so.

In the case of Instaloader, the issues you've identified can be addressed within the existing API design, as you mentioned. This approach allows us to improve the class without causing disruption or requiring users to adapt to a new interface. By enhancing the current design rather than creating a new one, we maintain consistency and avoid the overhead associated with maintaining multiple parallel implementations.

While I agree that forward-thinking design is crucial, it's equally important to balance innovation with stability. For long-term maintenance, we believe it's more efficient to refine and iterate on the existing class, addressing flaws incrementally while preserving the established API. This approach minimizes potential headaches by reducing the complexity that comes with maintaining deprecated code and managing transitions.

Issue #2313, opened 2024-07-20
Contradiction with PEP 561 (PEP 484)


liponex:
By the `py.typed` file, module is marked as PEP 561 compatible,
but the imports in `__init__.py` aren't marked public as required by [PEP 484](https://peps.python.org/pep-0484/#stub-files) for PEP 561 to work properly.
This contradicts with documentation.
E.g.:
• Documentation:
```python
import instaloader

# Get instance
L = instaloader.Instaloader()
```

• Reality:

> "Instaloader" is not exported from module "instaloader"

Meaning `Instaloader` is private and showing up like an error in the IDE.

**To Reproduce**
Can be reproduced by writing first example from documentation in VS Code (or VSCodium) with Pyright installed. 

**Expected behavior**
`Instaloader` and other objects must be marked public/exported as PEP 561 and PEP 484 require.

**Instaloader version**
4.12

**Additional context**
[Pyright checks](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface) for PEP 484 when `py.typed` present as PEP 561 [requires](https://peps.python.org/pep-0561/#specification).
After my analyses I've found out, that it can be fixed by adding `__all__` with list of all exported objects to the `__init__.py`, or by writing e.g. `from .instaloader import Instaloader as Instaloader`

On my opinion adding this:
```python
__all__ = [
    "Instaloader",
    "InstaloaderContext",
    "RateController",
    "LatestStamps",
    "NodeIterator",
    "FrozenNodeIterator",
    "resumable_iteration",
    "Hashtag",
    "Highlight",
    "Post",
    "PostSidecarNode",
    "PostComment",
    "PostCommentAnswer",
    "PostLocation",
    "Profile",
    "Story",
    "StoryItem",
    "TopSearchResults",
    "TitlePic",
    "load_structure_from_file",
    "save_structure_to_file",
    "load_structure",
    "get_json_structure"
]
```
is better, then this:
```python
from .exceptions import *
from .instaloader import Instaloader as Instaloader
from .instaloadercontext import (InstaloaderContext as InstaloaderContext,
                                 RateController as RateController)
from .lateststamps import LatestStamps as LatestStamps
from .nodeiterator import (NodeIterator as NodeIterator,
                           FrozenNodeIterator as FrozenNodeIterator,
                           resumable_iteration as resumable_iteration)
from .structures import (Hashtag as Hashtag,
                         Highlight as Highlight,
                         Post as Post,
[...]

PR #2314, opened 2024-07-23
PEP 561 compliance enhancement for issue #2313

liponex:
PEP 561 compliance enhancement for issue #2313
By the PEP 484 and PEP 561 some of the modules api's aren't exported, so this pull request fixes it, at least partially.

Changed only `__init__.py` file with this pattern
`from .X import A` -> `from .X import A as A`
`from .X import (A, B)` -> `from .X import (A as A, B as B)`
Addition of redundant `as` operator by the [PEP 484](https://peps.python.org/pep-0484/#stub-files) marks elements as exported.

I don't know if these are all the changes needed. This is the only thing that was tearing my work apart.
I think it's ready to merge.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -351,7 +351,7 @@ single-line-if-stmt=no
 allow-any-import-level=
 
 # Allow explicit reexports by alias from a package __init__.
-allow-reexport-from-package=no
+allow-reexport-from-package=yes
 
 # Allow wildcard imports from modules that define __all__.
 allow-wildcard-with-all=no
--- a/instaloader/__init__.py
+++ b/instaloader/__init__.py
@@ -13,11 +13,26 @@ else:
     win_unicode_console.enable()
 
 from .exceptions import *
-from .instaloader import Instaloader
-from .instaloadercontext import InstaloaderContext, RateController
-from .lateststamps import LatestStamps
-from .nodeiterator import NodeIterator, FrozenNodeIterator, resumable_iteration
-from .structures import (Hashtag, Highlight, Post, PostSidecarNode, PostComment, PostCommentAnswer, PostLocation,
-                         Profile, Story, StoryItem, TopSearchResults, TitlePic,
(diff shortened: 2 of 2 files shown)


aandergr:
Indeed, Pyright emits a `reportPrivateImportUsage` message when importing Instaloader the way we intend it to be imported. Thanks for addressing this and improving the developer experience in this regard!

https://microsoft.github.io/pyright/#/typed-libraries gives a brief explanation on the topic and lists ways to mark exports as public for the type checking.

In #2313 you mentioned that the solution setting `__all__` is better than adding redundant symbol aliases, and I agree on that, because I feel like it appears clearer to the reader, however an explaining comment for the purpose would still be beneficial. Still this PR introduces redundant aliases, did you change your opinion on that?

I am also wondering if there is a way to warn on the import style `from instaloader.instaloader import Instaloader`, which is not how we thought to import Instaloader symbols.

liponex:
@aandergr, I've thought a little more about maintaining this code. Since it will probably change, it is better to keep necessary changes together. Some people may forget to change `__all__`, as parameters in docstrings or information in comments.
While changing inputs, redundant symbol aliases will be more visible.
Also, if some lines aren't removed from `__all__` they will be shown in the IDE, so this will clearly sabotage someone's work.

liponex:
> Also, if some lines aren't removed from `__all__` they will be shown in the IDE, so this will clearly sabotage someone's work.

I'm not sure about it enough, but that isn't good anyway, as it'll contain wrong information and wold be potentially misused somehow.

P.S. The variant in this pull request is not as messy as I thought while writing the issue.

aandergr:
Indeed, the redundant-alias reexport seems the cleaner way to do this.

Thanks for your contribution!

fireattack:
I personally would prefer the `__all__` approach. In my experience this approach is more common and looks better. There were some examples of major projects (e.g. https://github.com/aio-libs/aiohttp/pull/7153) getting rid of re-exporting in favor of __all__.

PR #2330, opened 2024-08-04
Use new doc_id based query to get posts

### LINKED ISSUE #2319: Cannot retrieving posts from profile
**Describe the bug**

Cannot retrieving posts from profile when logging in
When I delete --login, retrieving posts from profile only gets the last 4 posts and --fast-update is not working
When I delete --login and --fast-update, retrieving posts from profile stops at 12 posts, responses a 401 error (Please wait a few minutes before you try again.)

**To Reproduce**

instaloader --max-connection-attempts=1 **--login=myaccount** --stories --no-metadata-json --abort-on=302,400,429,401 username --fast-update
instaloader --max-connection-attempts=1 --stories --no-metadata-json --abort-on=302,400,429,401 username --fast-update

**Expected behavior**

retrieve all the posts from profile and download them successfully

with --login:
Loaded session from C:\Users\XXXXXXX\AppData\Local\Instaloader\session-myaccount.
Logged in as myaccount.
[1/1] Downloading profile username
username\2024-06-03_15-56-51_UTC_profile_pic.jpg already exists
Retrieving posts from profile username.
Downloading stories
[1/1] Retrieving stories from profile username.
[  1/  2] username\2024-07-30_10-11-45_UTC.mp4 exists username\2024-07-30_10-11-45_UTC.jpg exists
Saved session to C:\Users\XXXXXXX\AppData\Local\Instaloader\session-myaccount.

without --login:
--login=USERNAME required to download stories.
Hint: Use --login to download higher-quality versions of pictures.
[1/1] Downloading profile username
username\2024-06-03_15-56-51_UTC_profile_pic.jpg already exists
Retrieving posts from profile username.
[ 1/14] username\2024-07-30_08-41-43_UTC.jpg exists [XXXXXXXXX…] unchanged
[ 2/14] username\2024-07-27_11-43-53_UTC.jpg exists [XXXXXXXXX…] unchanged
[ 3/14] username\2024-07-25_08-17-28_UTC.jpg exists [XXXXXXXXX…] unchanged
[ 4/14] username\2024-07-10_08-26-49_UTC.jpg exists [XXXXXXXXX…] unchanged

**Instaloader version**

4.12.1 (also test in 4.10.2 , and it shows the same problem)

ekalin:
Fixes #2319.

This uses the new query to retrieve user posts as discussed in the bug comments.

The returned object is (almost) equal to what was returned by the "iphone queries". (But it `video_versions` is null for images (and maybe sidecars), and `carousel_media` is null for non-sidecars, so I suppressed the `TypeError` that happens when trying to index a `None`.

  - **Is it just a proof of concept?** No, should be complete.
  - **Is the documentation updated (if appropriate)?** Yes. Nothing changed for the user, only the module documentation was updated.
  - **Do you consider it ready to be merged or is it a draft?** Should be now ready.
  - **Can we help you at some point?** Suggestions are welcome.

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -387,7 +387,8 @@ class InstaloaderContext:
 
     def get_json(self, path: str, params: Dict[str, Any], host: str = 'www.instagram.com',
                  session: Optional[requests.Session] = None, _attempt=1,
-                 response_headers: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
+                 response_headers: Optional[Dict[str, Any]] = None,
+                 use_post: bool = False) -> Dict[str, Any]:
         """JSON request to Instagram.
 
         :param path: URL, relative to the given domain which defaults to www.instagram.com/
@@ -414,7 +415,10 @@ class InstaloaderContext:
                 self._rate_controller.wait_before_query('iphone')
             if is_other_query:
                 self._rate_controller.wait_before_query('other')
-            resp = sess.get('https://{0}/{1}'.format(host, path), params=params, allow_redirects=False)
+            if use_post:
+                resp = sess.post('https://{0}/{1}'.format(host, path), data=params, allow_redirects=False)
+            else:
+                resp = sess.get('https://{0}/{1}'.format(host, path), params=params, allow_redirects=False)
             if resp.status_code in self.fatal_status_codes:
                 redirect = " redirect to {}".format(resp.headers['location']) if 'location' in resp.headers else ""
                 body = ""
@@ -528,12 +532,12 @@ class InstaloaderContext:
 
             variables_json = json.dumps(variables, separators=(',', ':'))
 
-            # FIXME: Use POST
             resp_json = self.get_json('graphql/query',
                                       params={'variables': variables_json,
                                               'doc_id': doc_id,
                                               'server_timestamps': 'true'},
-                                      session=tmpsession)
+                                      session=tmpsession,
+                                      use_post=True)
         if 'status' not in resp_json:
             self.error("GraphQL response did not contain a \"status\" field.")
         return resp_json
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -400,6 +400,7 @@ class InstaloaderContext:
         :raises ConnectionException: When query repeatedly failed.
         """
         is_graphql_query = 'query_hash' in params and 'graphql/query' in path
+        is_doc_id_query = 'doc_id' in params and 'graphql/query' in path
         is_iphone_query = host == 'i.instagram.com'
         is_other_query = not is_graphql_query and host == "www.instagram.com"
         sess = session if session else self._session
@@ -407,6 +408,8 @@ class InstaloaderContext:
             self.do_sleep()
             if is_graphql_query:
                 self._rate_controller.wait_before_query(params['query_hash'])
+            if is_doc_id_query:
+                self._rate_controller.wait_before_query(params['doc_id'])
             if is_iphone_query:
                 self._rate_controller.wait_before_query('iphone')
             if is_other_query:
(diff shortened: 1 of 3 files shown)


RazeBerry:
Thank you for the pull request! It seems to work for me at least for images and for videos!

ekalin:
Some comments on the changes:

- I used a fixed page size of 12 to emulate the requests done by a browser. Requesting more entries works, but it seems the limit is 33, even if more are asked.
- Only `after` seems to be really necessary for pagination, but again, I included some extra parameters to more closely emulate a browser's query.
- I removed the code for including the `x-instagram-gis` header in the new queries. I did not see this header (or anything similar) in the queries done by the web interface, and it seems to be working fine without it.
- I have no idea how the rate limiter works regarding these queries. I used a similar strategy (grouping by `doc_id` instead of `query_hash`).

ekalin:
I noticed a small change in behavior regarding posts that are posted by more than one user (I do not know the "official" name - but it's not tagged users, it's when it says "User1 and User2).

Previously `Post.owner_profile` (and therefore `{profile}` in filename templates) would always return the user that is being downloaded, even if it's not the main profile of the post. With this new query it may return a different user than the one being downloaded.

This is a consequence of what's returned by the query. There are two fields (`owner` and `user`), but both return the main owner of the post. (Some other field does list the other users.)

I wouldn't say this new behavior is wrong, just something to be aware of. It is always possible to use `{target}` in the filename pattern to name files based on the profile being downloaded, not where they originally came from.

ekalin:
I have been using this code for a couple days, and other users have also reported testing it, and it seems to work.

The documentation has also been updated, so I believe this should be ready for review.

instaloader/instaloadercontext.py
```diff
@@ -387,31 +387,42 @@ def _response_error(resp: requests.Response) -> str:
...
         :raises QueryReturnedNotFoundException: When the server responds with a 404.
         :raises ConnectionException: When query repeatedly failed.
+
+        .. versionchanged:: 4.13
+           Added `use_post` parameter.
         """
         is_graphql_query = 'query_hash' in params and 'graphql/query' in path
+        is_doc_id_query = 'doc_id' in params and 'graphql/query' in path
         is_iphone_query = host == 'i.instagram.com'
         is_other_query = not is_graphql_query and host == "www.instagram.com"
```
aandergr, line 410:
```suggestion
        is_other_query = not is_graphql_query and not is_doc_id_query and host == "www.instagram.com"
```

Otherwise `is_other_query` is true for doc_id queries, causing unexpected control flow.

instaloader/structures.py
```diff
@@ -1175,13 +1186,17 @@ def get_posts(self) -> NodeIterator[Post]:
         self._obtain_metadata()
         return NodeIterator(
             self._context,
-            '003056d32c2554def87228bc3fd9668a',
-            lambda d: d['data']['user']['edge_owner_to_timeline_media'],
-            lambda n: Post(self._context, n, self),
-            {'id': self.userid},
+            '',
```
aandergr, line 1189: Passing an empty string as query_hash and keeping an empty string in NodeIterator is not so nice I think. I suggest to make the argument an `Optional[str]` in the NodeIterator constructor and passing all parameters by their name in this invocation. Due to the position of the parameter we cannot entirely leave out the query_hash parameter, but I think verbosely passing `query_hash=None` rather than having an empty string at the parameter's position makes this invocation easier to understand.

instaloader/nodeiterator.py
```diff
@@ -77,9 +79,11 @@ def __init__(self,
                  query_variables: Optional[Dict[str, Any]] = None,
                  query_referer: Optional[str] = None,
                  first_data: Optional[Dict[str, Any]] = None,
-                 is_first: Optional[Callable[[T, Optional[T]], bool]] = None):
+                 is_first: Optional[Callable[[T, Optional[T]], bool]] = None,
+                 doc_id: Optional[str] = None):
```
aandergr, line 83: I would appreciate a `versionadded` comment in the docstring that briefly describes this change.

aandergr (commented):
Nice, thanks a lot for figuring this out and sharing such a clean, complete and well thought-through implementation with us! 

We will firstly merge and release #2332, which also seems to fix the issue in a manner that seems safe for the short-term. Thereafter, we will soon merge and release this change, which is probably better in the long run as it is closer to what the browser does, yet less tested in regard of rate controlling, and introduces the new query style which might also be handy in many other places.

ekalin:
I believe everything should be fixed.

I based this on the master branch, but let me know if you want me to rebase it on the upcoming branch.

PR #2333, opened 2024-08-07
add support for verify True/False

dorianmariecom:
I'm using proxies and was getting SSL errors.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -234,12 +234,11 @@ class Instaloader:
                  iphone_support: bool = True,
                  title_pattern: Optional[str] = None,
                  sanitize_paths: bool = False,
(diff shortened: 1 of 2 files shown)


dorianmariecom:
still need to test if that solved the issue, or if env_check is needed as well

aandergr:
To handle this, I would prefer the factory method approach outlined in #2195.

> This would introduce a way to also cleanly set custom session parameters, like SSL verification (suggested in https://github.com/instaloader/instaloader/pull/2221), or the request timeout (https://github.com/instaloader/instaloader/issues/330, https://github.com/instaloader/instaloader/pull/422), without that we need to pass-through and implement each of those parameters.

(https://github.com/instaloader/instaloader/pull/2195#issuecomment-2027015747)

PR #2336, opened 2024-08-13
minor doc update in basic-usage.rst

codeblech:
(almost) nobody says "Id est". Even I, being a native English speaker, had to search for it. It should've been "i.e." in the first place. But, even that is not needed. Let's keep it simple.


aandergr:
Thanks!

PR #2340, opened 2024-08-18
Remove rhx_gis header support

### LINKED ISSUE #2338: JSONDecodeError: Expecting value - Issue with Profile.from_id
Hello,

I have a large number of UserIds for which I need to retrieve profile information. Since the Instaloader library allows retrieval of profile information by username, my goal is to convert these UserIds into usernames. However, I'm encountering an issue when trying to do this using the Profile.from_id method in the Instaloader library. The issue seems to be related to a JSONDecodeError that occurs when attempting to parse the JSON response from the server.

Use the following code snippet to retrieve profile information:

```
import instaloader

L = instaloader.Instaloader()

user_id = 1234567
profile = instaloader.Profile.from_id(L.context, user_id)

print(profile.username)
```

**Expected Behavior:**
The profile information, specifically the username, should be retrieved and printed.

**Actual Behavior:**
The script fails with the following error message:

```
JSON Query to : Expecting value: line 1 column 1 (char 0) [retrying; skip with ^C]
...
instaloader.exceptions.ConnectionException: JSON Query to : Expecting value: line 1 column 1 (char 0)
```
**Additional Information:**

- Python Version: Python 3.12
- Instaloader Version: 4.12.1

I have checked my internet connection, and it seems stable. I have also tried using a VPN, but the issue persists. It seems that the server might be returning an empty or invalid JSON response.

Could you please advise on how to resolve this issue, or if this might be related to changes in Instagram's API?

Thank you for your help.

ekalin:
Fixes #2338, and removes useless code.

The query used to obtain the `rhx_gis` parameter (`root_rhx_gis` function) does not work anymore.

The parameter was only used when obtaining a profile by its id, when not logged in. This query is working fine without the special header, so the whole functionality concerning rhx_gis was removed.

This also fixes this misleading messages logged when trying to obtain a profile anonymously:
```
JSON Query to : Expecting value: line 1 column 1 (char 0)
```

Note that this has changed the signature of the `graphql_query` function. I'm not sure if it's part of the public API, but it's very low level, so I changed it anyway.

  - **Is it just a proof of concept?** No
  - **Is the documentation updated (if appropriate)?** Yes
  - **Do you consider it ready to be merged or is it a draft?** Ready
  - **Can we help you at some point?** No, but feel free to make suggestions.

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -493,7 +493,7 @@ class InstaloaderContext:
         :param referer: HTTP Referer, or None.
         :return: The server's response dictionary.
 
-        .. versionchanged:: 4.13
+        .. versionchanged:: 4.13.1
            Removed the `rhx_gis` parameter.
         """
         with copy_session(self._session, self.request_timeout) as tmpsession:
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -1,4 +1,3 @@
-import hashlib
 import json
 import os
 import pickle
@@ -94,7 +93,6 @@ class InstaloaderContext:
         self.quiet = quiet
         self.max_connection_attempts = max_connection_attempts
         self._graphql_page_length = 50
-        self._root_rhx_gis = None
         self.two_factor_auth_pending = None
         self.iphone_support = iphone_support
         self.iphone_headers = default_iphone_headers()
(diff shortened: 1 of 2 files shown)


instaloader/instaloadercontext.py
```diff
@@ -486,15 +484,17 @@ def get_json(self, path: str, params: Dict[str, Any], host: str = 'www.instagram
...
         """
         Do a GraphQL Query.
 
         :param query_hash: Query identifying hash.
         :param variables: Variables for the Query.
         :param referer: HTTP Referer, or None.
-        :param rhx_gis: 'rhx_gis' variable as somewhere returned by Instagram, needed to 'sign' request
         :return: The server's response dictionary.
+
+        .. versionchanged:: 4.13
```
aandergr, line 496:
```suggestion
        .. versionchanged:: 4.13.1
```

We will merge this into `master` rather than the upcoming minor update.

It does minimally change the API though, however in a way that is required to adapt to Instagram's changes. There is no valid reason anymore to pass `rhx_gis` parameter.

aandergr:
Thanks!

Issue #471, opened 2019-11-30
Documentation unclear on loading single posts.


novakda:
While instaloader can download single posts, how to do so is unclear from the documentation and command line.  Both suggest that "-post" is a literal argument, not indicating that "post" here is placeholder for the shortcode itself.

Please refer to #283 for an example of another user with the same issue.

**Describe the solution you'd like**
Suggest providing an example along with the documentation, and change command line to clearly indicate -post is a placeholder, e.g.

instaloader -- -_YOUR-SHORTCODE_

(similar to --login YOUR-USERNAME, I'm not sure what the convention is for arguments here)

`instaloader -- -B5GWcVnhLWf`

aandergr:
The documentation lists `-post` along with other targets where the target is a placeholder, such as `profile`, `@profile`, `#hashtag`:

> **-post**
> The single **post** with the given shortcode. Must be preceeded by `--` in the argument list to not be mistaken as an option flag.

Also, just as in almost every CLI tool, all our command line flags and options are prefixed with a `--` (two '-').

However, we see that there have been confused users. We are open for suggestions on how to improve the documentation. Feel free to submit a PR to discuss alternative and less confusing ideas on how to introduce this target. The source file for this documentation page is `docs/basic-usage.rst`.

newnewyork:
It is very unclear how to use this option. You could give an example.

aandergr:
Feel free to submit a PR to discuss alternative and less confusing ideas on how to introduce the single post target.

PR #2343, opened 2024-08-19
Clarify that post short codes should be preceded by a hyphen

alexwlchan:
I found the documentation for downloading single posts confusing (and I’m not the only one, e.g. #471).

When I read the example command, I thought that the leading hyphen in `-B_K4CykAOtf` was part of the shortcode – I was trying to download an Instagram post with a hyphen in the shortcode, and [some posts](https://www.instagram.com/p/-9WYAqqvSV/) have a shortcode that starts with a hyphen!

Having the URL there for a point of comparison should make this clearer.


aandergr:
Thanks!

PR #2276, opened 2024-06-03
feat: adding max_count parameter to limit posts on download

Mesheo:
### Motivation for this change
👋 Hi! I made this PR to make it possible to control the quantity of posts downloaded by profile.
The core method `self.posts_download_loop()` already has this functionality, but the external calling function `instaloader.download_profiles()` were not adjusted to receive or pass this argument further.

  - Fixes #2269. (I believe this change will solve this issue) 

### Changes proposed in this pull request
* Added parameter `max_count` to `download_profiles()` method

### The completeness of this change
  - Is it just a proof of concept? **No**
  - Is the documentation updated (if appropriate)? **Yes**
  - Do you consider it ready to be merged or is it a draft? **Ready** 
  - Can we help you at some point? **Feel free to give me some guidelines to better solve this problem, I am fully available to do so.**

Issue #1660, opened 2022-07-26
count for profile downloads


movAX13h:
Is there a technical reason why `--count` is not available for profile downloads?

If a profile has many thousands of posts but I always only want to get the newest 10 posts, currently the only option to reduce workload is to use `--post-filter` with a certain datetime and hope that I get about 10 posts and it also takes very long because instaloader is going through all posts.

It appears to me that the newest posts are always processed first so `--count 10` to stop iteration after 10 posts would really help in this situation.

Any thoughts appreciated; please let me know if I'm missing something.
Thank you!

aandergr:
This would unfortunately change Instaloader's behavior in a backward-incompatible manner, since an invocation with --count and a mix of hashtags/feed/location and a profile would then behave differently in that the max count now also affects the profile and not only hashtags/feed/location.

Also, it is quite trivial for the user to only download a certain count of posts of a profile, e.g. by calling download_post() for each post of an islice()'d profile.get_posts(). There is no need to include that in the CLI.

movAX13h:
I see. Thank you for your response.

movAX13h:
Just noticed that this is a duplicate of #327 and #141 ... Many users seem to ask for this because it is a very common use-case and somehow weird that this is not the default behaviour of instaloader.

vadimkantorov:
Maybe a new command-line option should be introduced for this very frequent usecase?

This is a very useful usecase for small websites: render statically several most recent instagram posts

If instaloader can download most recent 10 instagram posts and save all the info in a directory, or serialize into JSON, it'd be extremely useful

PR #2345, opened 2024-08-20
pass max_count var to func download_profiles to fix PR #2276

### LINKED ISSUE #2339: New "Max Count" Functionality
In the notes for 4.13, one of the items included was an update to make "--count" work for profiles. At least I think that's what it meant. I just tried "--count 10" on a profile and had to manually abort after 11. 

The documentation for CLI options does not have an update to reflect how we are to use this from the command line. 

Can we please get an explanation and the documentation for this be updated please?

mumixam:
Fixes #2339

PR #2276 added max_count var to the ``download_profiles`` function but the ``max_count`` var was not being passed from the function in ``__main__.py`` that calls download_profiles.

This PR corrects that.

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -300,7 +300,8 @@ def _main(instaloader: Instaloader, targetlist: List[str],
         instaloader.download_profiles(profiles,
                                       download_profile_pic, download_posts, download_tagged, download_igtv,
                                       download_highlights, download_stories,
-                                      fast_update, post_filter, storyitem_filter, latest_stamps=latest_stamps)
+                                      fast_update, post_filter, storyitem_filter, latest_stamps=latest_stamps,
+                                      max_count=max_count)
         if anonymous_retry_profiles:
             instaloader.context.log("Downloading anonymously: {}"
                                     .format(' '.join([p.username for p in anonymous_retry_profiles])))


MonkeyKingViper:
I tested this update and it works exactly as expected

MonkeyKingViper:
>Also, it is quite trivial for the user to only download a certain count of posts of a profile, e.g. by calling download_post() for each post of an islice()'d profile.get_posts(). There is no need to include that in the CLI.
Respectfully, is this really trivial? Wouldn't that assume that everyone knows Python? My knowledge of Python is slim to none. And the most attractive thing about this tool was the simple CLI uses that I understand.

For example, I have a .bat file that calls a number of arg files (settings and and profiles) sequentially with a random pause between each (between 4 and 6 minutes). Posts, Stories, Highlights. And it sends the content where I want it based on what I'm downloading through the args for each unique need. I know how to do that in batch, but I have no idea how to run any of this tool in Python. And it seems many others are in a similar state as someone mentioned in #1660 that it looked like a dupe of 3ish requests.

Would it not be possible to work the solution you have into something CLI friendly for those of us that are not Python inclined? Maybe leave --count alone but have a --profile-count that does what this is intending? I'm not a coder I truly have no idea. I can't lend anything but ideas.

aandergr:
Our goal with Instaloader's CLI is to keep it lean and focused on the most common use cases, ensuring that it remains intuitive without overwhelming users with too many options.

While we appreciate the suggestion for --profile-count, adding additional flags like this could lead to feature creep, making the CLI harder to maintain and potentially more confusing for others. For more advanced or customized use cases, Instaloader's Python scripting functionality allows greater flexibility. Although we recognize that Python knowledge is required, the community and Instaloader's documentation provides a lot of helpful resources for getting started with simple scripts.

mumixam:
I didn't actually add any new flags that code already existed.
The only thing the PR did was properly pass a variable to a function.
Seems like this code should be fixed then.
https://github.com/instaloader/instaloader/blame/c5dacb5f6d8bebbfb7ce2a448e32866ecd2d9868/instaloader/instaloader.py#L1020

PR #2393, opened 2024-10-10
Introductory fix to issue #2384

Marethyun:
This is a draft intended to bootstrap the resolution of issue #2384

Fix to reflect changes of the Instagram API (hinted here https://github.com/yt-dlp/yt-dlp/commit/cf85cba5d9496bd2689e1070005b4d1b4cd3dc6d)

I know neither python nor the codebase so I didn't take the time to refactor anything (graphql_query now takes a wrongly named parameter query_hash) and just surrounded the API call with an if/else.

In the current state I only tested it to work with shortcodes (ex. `instaloader -- -DAOdehwP6fC` now works as expected) and nothing else.

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -496,6 +496,7 @@ class InstaloaderContext:
         .. versionchanged:: 4.13.1
            Removed the `rhx_gis` parameter.
         """
+
         with copy_session(self._session, self.request_timeout) as tmpsession:
             tmpsession.headers.update(self._default_http_header(empty_session_only=True))
             del tmpsession.headers['Connection']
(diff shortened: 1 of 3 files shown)


aandergr:
Thanks for addressing this! This seems a valid approach to fix #2384. I rebased this branch to `master`.

With the changes of the typenames it raises many warnings like
```
Warning: <Post ...> has unknown typename: GraphImage
```

Just ping us when this PR is ready to be merged.

instaloader/instaloadercontext.py
```diff
@@ -508,12 +509,21 @@ def graphql_query(self, query_hash: str, variables: Dict[str, Any],
...
+                resp_json = self.get_json('graphql/query',
+                                          params={'doc_id': '8845758582119845',
+                                                  'variables': variables_json},
+                                          session=tmpsession)
+
+            else:
+                resp_json = self.get_json('graphql/query',
+                                          params={'query_hash': query_hash,
+                                                  'variables': variables_json},
+                                          session=tmpsession)
```
lucascwd, line 523: I noticed that the `graphql_query` function has a nearly identical counterpart, `doc_id_graphql_query`, which takes a `doc_id` instead of a `query_hash`. Since the query hash `2b0673e0dc4580674a88d426fe00ea90` is only referenced in two places within the codebase, we can just swap out those `graphql_query` calls with `doc_id_graphql_query`. What do you think?

```py
@classmethod
def from_mediaid(cls, context: InstaloaderContext, mediaid: int):
    """Create a StoryItem object from a given mediaid.

    .. versionadded:: 4.9
    """
    # Before -----
    pic_json = context.graphql_query(
        '2b0673e0dc4580674a88d426fe00ea90',
        {'shortcode': Post.mediaid_to_shortcode(mediaid)}
    )
    # After -----
    pic_json = context.doc_id_graphql_query(
        '8845758582119845',
        {'shortcode': Post.mediaid_to_shortcode(mediaid)}
    )
```

lucascwd (commented):
I'm wondering if it's really necessary to change the type names, or if perhaps mapping the `XDT*` types back to the original names could be a simpler solution. Since the post types seem to be accessed (always?) using the `typename` property, maybe just removing the XDT prefix there would be enough:

https://github.com/instaloader/instaloader/blob/c5dacb5f6d8bebbfb7ce2a448e32866ecd2d9868/instaloader/structures.py#L429-L432

https://github.com/instaloader/instaloader/blob/c5dacb5f6d8bebbfb7ce2a448e32866ecd2d9868/instaloader/structures.py#L1479-L1481

I'm thinking that changing the type names might cause a breaking change, although it's not entirely clear whether Instagram's internal type names are even part of the public interface :thinking:

aandergr:
Obsolete now due to 9a74d9e1092e37d4e1f8b6c6431b18a4793781d7.

PR #2412, opened 2024-10-24
Update windows_exe.yml to address deprecation warnings

tinyboxvk:
https://github.com/actions/create-release and https://github.com/actions/upload-release-asset are no longer maintained. It is suggested in the README switching to a maintained actions like https://github.com/softprops/action-gh-release. 

This PR resolves deprecation warnings like
```
The following actions uses node12 which is deprecated and will be forced to run on node16: actions/create-release@v1
The following actions use a deprecated Node.js version and will be forced to run on node20: actions/create-release@v1
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files.
```


aandergr:
Nice, thanks for addresing this!

PR #2413, opened 2024-10-24
Bump pipenv, remove Python 3.8, add Python 3.13

tinyboxvk:
Python 3.8 is EOL. This PR removes it from lint.yml and add in 3.13.

Bumping pipenv to the latest version to support 3.13.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,10 +18,10 @@ def get_version():
     raise SystemExit("Could not find version string.")
 
 
-if sys.version_info < (3, 8):
-    sys.exit('Instaloader requires Python >= 3.8.')
+if sys.version_info < (3, 9):
+    sys.exit('Instaloader requires Python >= 3.9.')
 
-requirements = ['requests>=2.4']
+requirements = ['requests>=2.32']
 optional_requirements = {
     'browser_cookie3': ['browser_cookie3>=0.19.1'],
 }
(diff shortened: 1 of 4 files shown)


aandergr:
Thanks!

Could you rebase this to `upcoming/v4.14`?

Also, in `setup.py` there is a check that the python version is at least 3.8, a python_requires field, and a list of supported versions as trove classifiers. Could you update them as well?

tinyboxvk:
> Could you rebase this to `upcoming/v4.14`?
> 
> Also, in `setup.py` there is a check that the python version is at least 3.8, a python_requires field, and a list of supported versions as trove classifiers. Could you update them as well?

Done.

setup.py
```diff
@@ -18,10 +18,10 @@ def get_version():
     raise SystemExit("Could not find version string.")
 
 
-if sys.version_info < (3, 8):
-    sys.exit('Instaloader requires Python >= 3.8.')
+if sys.version_info < (3, 9):
+    sys.exit('Instaloader requires Python >= 3.9.')
 
-requirements = ['requests>=2.4']
+requirements = ['requests>=2.32']
```
aandergr, line 24:
```suggestion
requirements = ['requests>=2.25']
```

I suggest to require >= 2.25 since this is safely recent-enough for us, but old enough to be shipped in older, but still-supported debian or ubuntu LTS releases.

aandergr:
Thanks!

PR #2430, opened 2024-11-10
publish wheels

dimbleby:
direct invocation of `setup.py` is considered deprecated these days; and it is desirable to publish wheels as well as sdists.


aandergr:
Thanks! This updates the package building process to match the current recommendations described in the [Python Packaging Guide](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/).

PR #2449, opened 2024-12-03
fixed bug with dowload hastag

GrigoriyKruchinin:
#### Description of the Issue:
When attempting to scrape content for a hashtag using `instaloader`, the following error occurs consistently:
```
JSON Query to explore/tags/{hashtag}/: 404 Not Found [retrying; skip with ^C]
```

This issue arises even for valid and active hashtags, such as `#foodie` or other commonly used ones. It seems that Instagram has changed the endpoint or requires additional handling to correctly access hashtag data.

---

#### Summary of Fix:
This PR introduces a fix that resolves the `404 Not Found` error during hashtag scraping. The key changes include:
1. Updated the `Hashtag` endpoint to use the `web_info` API instead of the outdated `explore/tags` URL.
2. Added a fallback mechanism to handle potential redirects to the login page, ensuring session validity.
3. Optimized error handling for better user feedback when scraping fails due to session or endpoint issues.

---

#### Changes Made:
- **Updated API Endpoint**:
  - Replaced the outdated `/explore/tags/{hashtag}/` endpoint with the newer `/api/v1/tags/web_info/?tag_name={hashtag}`.
  - Included necessary headers and cookies for API requests to pass Instagram's updated security checks.
  
- **Improved Redirect Handling**:
  - Added logic to detect and manage redirection to the login page during hashtag scraping.
  - Enhanced session persistence to ensure users remain authenticated across requests.

- **Error Messaging**:
  - Improved error reporting with actionable messages (e.g., "Login required" or "Endpoint deprecated").

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -1862,7 +1862,7 @@ class Hashtag:
         return self._node["name"].lower()
 
     def _query(self, params):
-        json_response = self._context.get_json("explore/tags/{0}/".format(self.name), params)
+        json_response = self._context.get_json("api/v1/tags/web_info/?tag_name={0}".format(self.name), params)
         return json_response["graphql"]["hashtag"] if "graphql" in json_response else json_response["data"]
 
     def _obtain_metadata(self):
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -8,6 +8,7 @@ import textwrap
 import time
 import urllib.parse
 import uuid
+import re
 from contextlib import contextmanager, suppress
 from datetime import datetime, timedelta
 from functools import partial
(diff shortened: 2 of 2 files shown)


instaloader/structures.py
```diff
@@ -1862,7 +1862,7 @@ def name(self):
         return self._node["name"].lower()
 
     def _query(self, params):
-        json_response = self._context.get_json("explore/tags/{0}/".format(self.name), params)
+        json_response = self._context.get_json("api/v1/tags/web_info/?tag_name={0}".format(self.name), params)
```
aandergr, line 1865: Nice, thanks for figuring this out and proposing this change!

This query type is already implemented as get_iphone_json. Thus, the following will do the request, making the proposed change in instaloadercontext.py unnecessary.

```suggestion
        json_response = self._context.get_iphone_json(
            "api/v1/tags/web_info/", {**params, "tag_name": self.name}
        )
```

Also, since this does no longer work when not logged-in, it would be good to annotate Instaloader.download_hashtag() with `@_requires_login` and to mention that in the --help output and in the docs (`basic-usage.rst`).

aandergr:
@GrigoriyKruchinin Are you still working on this?

sadewa25:
Hope this is can be publish because the hashtags features is error to download

PR #2474, opened 2025-01-09
Fix typos

tinyboxvk:


aandergr:
We typed our code changes quite hastily,  
But spelled “succesfully” so nastily;  
We had “supression” in store,  
“Neccessary,” “preceeding,” and more—  
Thanks for the fix done so gracefully!

PR #2507, opened 2025-02-07
Bump dependencies to support Python 3.13 in CI

tinyboxvk:
PyInstaller added support for Python 3.13 in version 6.10.0.

Other dependencies are also updated using `pipenv lock`.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -14,11 +14,11 @@ jobs:
       - name: "Setup Python"
         uses: actions/setup-python@v5
         with:
-          python-version: 3.12
+          python-version: 3.13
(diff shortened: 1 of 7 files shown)


tinyboxvk:
Looks like the checks are failing because of https://github.com/pylint-dev/pylint/issues/9099 introduced in the newer Pylint version.

aandergr:
Thanks for addressing this! We updated the CI dependencies already in 487d2c366d2a2cdd78a50132d6cc47e053d12059, which is part of the `upcoming/v4.15` branch. There, we addressed the new features of Pylint and Mypy. Could you rebase your change to that branch so that we can merge the update of Pipenv and the Python version used for PyInstaller?

Issue #2536, opened 2025-03-26
"Pinned" reels confuse `--latest-stamps`


vanyasem:
**Describe the bug**
An Instagram profile can choose to "pin" some posts or reels to the top of their profile.
"Pinned" items are not guaranteed to be the latest, in fact most of the time they aren't.
While using `--latest-stamps`, instaloader checks the date of the first post/reel in the user's feed, without taking into account that the first post/reel could be a pinned one. This leads to instaloader incorrectly assuming that there were no new posts/reels since last scraping, and not proceeding with the download.

**To Reproduce**
1) Find a user profile that has a post/reel pinned in their profile, with newer posts/reels having been posted since
2) Run instaloader with the `--latest-stamps` options to generate `stamps.ini`
3) Note that instaloader (incorrectly) wrote _the date of the first (pinned) post_ as the "latest" date
4) This results in instaloader instantly bailing next time it's ran with the `--latest-stamps`, as it will always encounter that first "pinned" post and (incorrectly) assume that it's the latest post, the date of which matches the date saved specified in `stamps.ini`, resulting in all new user media being ignored

**Expected behavior**
instaloader ignores (excludes) pinned posts when calculating dates for `--latest-stamps`, both when reading and writing

**Error messages and tracebacks**
example `stamps.ini`:
```
[REDACTED]
reels-timestamp = 2024-05-29T20:18:55.000000+0000
```

full log (when ran without `--latest-stamps`) - to illustrate that the first 3 posts are pinned, and thus have dates in the past, while the next posts (4-6) are new:
```
Retrieving reels videos for profile REDACTED.
[  1] REDACTED\2024-05-29_20-18-55_UTC.jpg exists [] unchanged REDACTED\2024-05-29_20-18-55_UTC.mp4 exists json
[  2] REDACTED\2024-03-21_19-02-26_UTC.jpg exists [] unchanged REDACTED\2024-03-21_19-02-26_UTC.mp4 exists json
[  3] REDACTED\2023-12-25_17-49-41_UTC.jpg exists [] unchanged REDACTED\2023-12-25_17-49-41_UTC.mp4 exists json
[  4] REDACTED\2025-03-24_23-38-04_UTC.jpg [] REDACTED\2025-03-24_23-38-04_UTC.mp4 json
[  5] REDACTED\2025-03-24_23-37-18_UTC.jpg [] REDACTED\2025-03-24_23-37-18_UTC.mp4 json
[...]

PR #2538, opened 2025-03-26
Check up to 6 pinned posts

ekalin:
As mentioned in #2536, it is now possible to pin up to 6 posts, so this change increases the number of posts that are checked for being possibly pinned when using --latest-stamps.

  - Is the documentation updated (if appropriate)? Not necessary
  - Can we help you at some point? Not necessary.


vanyasem:
Pin limit is still 3, so this PR is indeed not necessary - the issue observed in #2536 happens in a profile that only has 3 reels pinned. There's a separate issue of pinned **reels** not being recognized as pinned at all

ekalin:
Indeed the limit has not changed: https://help.instagram.com/318456537074409/?cms_platform=android-app&helpref=platform_switcher

Issue #2576, opened 2025-07-10
load-cookies KeyError: 'data'


mpalermo73:
**Describe the bug**
When using the `--load-cookies` option, I am getting failure with Traceback

**To Reproduce**
1) Install instaloader (4.14.1?)
2) Successfully login to instagram in your browser to produce the cookie, chrome in my case.
2) `instaloader --load-cookies chrome` 

**Error messages and tracebacks**
```
-> % instaloader --load-cookies chrome
Cookies loaded successfully from chrome
Traceback (most recent call last):
  File "/usr/bin/instaloader", line 11, in <module>
    sys.exit(main())
             ~~~~^^
  File "/usr/lib/python3.12/site-packages/instaloader/__main__.py", line 584, in main
    exit_code = _main(loader,
                      args.profile,
    ...<15 lines>...
                      browser=args.load_cookies,
                      cookiefile=args.cookiefile)
  File "/usr/lib/python3.12/site-packages/instaloader/__main__.py", line 166, in _main
    import_session(browser.lower(), instaloader, cookiefile)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/instaloader/__main__.py", line 127, in import_session
    username = instaloader.test_login()
  File "/usr/lib/python3.12/site-packages/instaloader/instaloader.py", line 641, in test_login
    return self.context.test_login()
           ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.12/site-packages/instaloader/instaloadercontext.py", line 246, in test_login
    return data["data"]["user"]["username"] if data["data"]["user"] is not None else None
                                               ~~~~^^^^^^^^
KeyError: 'data'
```

**Instaloader version**
```
-> % instaloader --version
4.14.1
```
```
-> % pip show browser-cookie3
Name: browser-cookie3
Version: 0.20.1
```

mpalermo73:
That's interesting as I've been using chrome for this for a while without issue.  

I forcibly logged out / in and instaloader is working totally fine with chrome cookies now.  So I guess this is a bogus bug report as I didn't try to relog before posting.  But as the same time, I am completely unable to reproduce this "Chromium-based browsers" not working issue either.  `WORKS_FOR_ME`

```
-> % instaloader --load-cookies chrome
Cookies loaded successfully from chrome
[...]

Issue #2552, opened 2025-04-15
Cookie decrypt error


JoeCool90:
**Describe the bug**
Cookie decryption error

**To Reproduce**
.\instaloader.exe [instagram_profile_name] --stories --highlights --tagged --reels --igtv --comments --fast-update --load-cookies chrome --cookiefile "C:\Users\[profile_name]\AppData\Local\Google\Chrome\User Data\Profile 9\Network\Cookies"

**Expected behavior**
To log in with error

**Error messages and tracebacks**
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (Traceback (most recent call last)::String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
  File "browser_cookie3\__init__.py", line 539, in _decrypt
  File "browser_cookie3\__init__.py", line 530, in _decrypt_windows_chromium
  File "browser_cookie3\__init__.py", line 89, in _crypt_unprotect_data
RuntimeError: Failed to decrypt the cipher text with DPAPI
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "browser_cookie3\__init__.py", line 555, in _decrypt
  File "Cryptodome\Cipher\_mode_gcm.py", line 567, in decrypt_and_verify
  File "Cryptodome\Cipher\_mode_gcm.py", line 508, in verify
ValueError: MAC check failed
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "__main__.py", line 631, in <module>
  File "__main__.py", line 628, in __main
  File "__main__.py", line 584, in main
  File "__main__.py", line 166, in _main
  File "__main__.py", line 124, in import_session
  File "__main__.py", line 105, in get_cookies_from_instagram
  File "browser_cookie3\__init__.py", line 1160, in chrome
  File "browser_cookie3\__init__.py", line 515, in load
  File "browser_cookie3\__init__.py", line 557, in _decrypt
browser_cookie3.BrowserCookieError: Unable to get key for cookie decryption
[42464] Failed to execute script '__main__' due to unhandled exception!

**Instaloader version**
4.14.1

**Additional context**
Windows 10, calling via powershell

Showmaster3000:
[...]

Issue #2550, opened 2025-04-10
--load-cookies doesn't work on Brave browser


Bruce-BC:
**Describe the bug**
the option "--load-cookies" not working on brave,
I tested safari and chrome, and both works.

**To Reproduce**
(e.g. Instaloader command line)
instaloader [private profile] --load-cookies brave # No cookies found for Instagram in brave, Are you logged in succesfully in brave?
instaloader [private profile] --load-cookies safari # works

**Error messages and tracebacks**
No cookies found for Instagram in brave, Are you logged in succesfully in brave?

**Instaloader version**
4.13.2

bamit99:
Use Firefox. It works!

Issue #2510, opened 2025-02-09
Option --cookiefile not work


MGuerrera:
**Describe the bug**
Instaloader give me this error:
JSON Query to graphql/query: 401 Unauthorized

I tried using the --cookiefile option, specifying the location of the Chrome browser’s cookie file where I am logged into Instagram.
But it keeps giving me the same error.

**To Reproduce**

instaloader --cookiefile "/Users/[profile_name]/Library/Application Support/Google/Chrome/Default/Cookies"[instagram_profile_name]
also tried
instaloader --cookiefile "/Users/[profile_name]/Library/Application Support/Google/Chrome/Default/" [instagram_profile_name]

**Expected behavior**
Downloading the specified profile via authentication

**Error messages and tracebacks**
```
Stored ID 1583740558 for profile [instagram_profile_name].
Hint: Login to download higher-quality versions of pictures.
[1/1] Downloading profile [instagram_profile_name]
/Users/mg/Pictures/Instaloader/[instagram_profile_name]/2024-10-22_13-34-17_UTC_profile_pic.jpg
Retrieving posts from profile [instagram_profile_name].
JSON Query to graphql/query: 401 Unauthorized - "fail" status, message "Please wait a few minutes before you try again." when accessing https://www.instagram.com/graphql/query [retrying; skip with ^C]
JSON Query to graphql/query: 401 Unauthorized - "fail" status, message "Please wait a few minutes before you try again." when accessing https://www.instagram.com/graphql/query?variables=%7B%22data%22%3A%7B%22count%22%3A12%2C%22include_relationship_info%22%3Atrue%2C%22latest_besties_reel_media%22%3Atrue%2C%22latest_reel_media%22%3Atrue%7D%2C%22username%22%3A%22[instragram_profile_name]%22%2C%22__relay_internal__pv__PolarisFeedShareMenurelayprovider%22%3Afalse%7D&doc_id=7898261790222653&server_timestamps=true [retrying; skip with ^C]
[...]

PR #2577, opened 2025-07-11
Update session import

denyshon:
Fixes:
- #2510 
- #2550
- #2552
- #2576

Removes Chromium-based browsers from get_cookies_from_instagram (and correspondingly updates the docs).
_Reason: browser_cookie3 has not been updated after Chromium improved cookie encryption._

Adds a default value (None) for cookiefile in import_session (_as expected by browser_cookie3_).
Changes cookie_file default value to None in get_cookies_from_instagram (_as expected by browser_cookie3_).

**Ready to be merged**

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -82,24 +82,16 @@ def filterstr_to_filterfunc(filter_str: str, item_type: type):
     return filterfunc
 
 
-def get_cookies_from_instagram(domain, browser, cookie_file='', cookie_name=''):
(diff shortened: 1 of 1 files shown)


denyshon:
@aandergr There's also an option to truly resolve the issue by providing support for Chromium-based browsers' cookies with a different library. The most reliable solution seems to be importing from an open browser window (as, e.g., playwright does). Are we potentially ok with that (opening the browser to log in and then closing it)?

PR #2578, opened 2025-07-11
Add download_captions parameter

denyshon:
Adds a download_captions parameter (no cmd option though), providing an option to disable post capture download.

As for now, no additional documentation updates seem to be needed.

Ready to be merged (although a cmd option may be added first).


denyshon:
@aandergr Sorry, I haven't noticed that. Is it exposed in the Python module? If not, it may still be a good idea to add an argument for that (changing its implementation to match the command line option though).

PR #2579, opened 2025-07-11
Removed [1/1] for single profile downloading when using download_profiles

denyshon:
Removed [1/1] for single profile downloading when using download_profiles (to allow calling the function for an external profile list, one profile at a time).

No documentation update needed.

**Ready to be merged**


aandergr:
This is a purely cosmetic change in the log output with little (if at all) improvement. Even though the change is small, I prefer to avoid adding special cases and keep the code as simple as possible.

PR #2581, opened 2025-07-14
Add SYSTEMEXITs to get_json for substantial Instagram requirements

denyshon:
When Instagram returns "_feedback_required_" or "_challenge_required_", any further requests will be declined until the required action is performed manually. Those messages mean Instagram suspects us of being a bot, so we must stop producing more requests immediately in order not to increase the suspicions.

The PR adds SYSTEMEXITs to get_json (status_code 400 block) for substantial Instagram requirements (feedback_required | challenge_required), providing detailed explanations to the console.

The documentation does not seem to need to be updated.

**Ready to be merged**

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -395,7 +395,8 @@ class InstaloaderContext:
         :param session: Session to use, or None to use self.session
         :param use_post: Use POST instead of GET to make the request
         :return: Decoded response dictionary
-        :raises QueryReturnedBadRequestException: When the server responds with a 400.
+        :raises AbortDownloadException: When the server responds with "feedback_required"/"challenge_required"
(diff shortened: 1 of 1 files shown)


aandergr:
The idea of explicitly handling feedback_required and challenge_required is good. These responses are indeed meaningful and deserve clearer handling.

However, this PR cannot be merged as is. Raising SystemExit inside library code is not acceptable, since it bypasses all existing error handling and immediately terminates the whole process, including scripts and downstream software that depend on Instaloader as a library.

denyshon:
@aandergr thank you for your reply! You are absolutely right about SystemExit. I've updated the code to use the existing AbortDownloadException and also made it more aligned with the project's style. Could you please take a look at #2612?

PR #2583, opened 2025-07-29
Add filename suffixes for stories

denyshon:
Adds filename suffixes for stories, similar to ones for posts (to give more flexibility for the filename pattern).

No documentation update needed.

**Ready to be merged**


aandergr:
This adds a _count suffix to the story items, unless {filename} is part of the filename pattern, which per default is not. This is a breaking change. Regardless of whether the proposed filenames are better or not, we cannot simply change this.

Issue #2598, opened 2025-09-04
Error doesn't save resume information while user interrupt does


gnestor:
**Describe the bug**
When the process ends due to an error (e.g. 400), the resume information isn't saved, but if it ends due to user interrupt, it does.

**To Reproduce**
```
instaloader --login=USERNAME --dirname-pattern={profile} --filename-pattern={shortcode} --no-captions --no-video-thumbnails --no-compress-json :saved --latest-stamps
Using latest stamps from /Users/USER/.config/instaloader/latest-stamps.ini.
Loaded session from /Users/USER/.config/instaloader/session-USERNAME.
Logged in as USERNAME.
Retrieving saved posts...
Resuming from :saved/iterator_PO_ezfYk.json.xz.
...
[ 7419/10844] IG_HANDLE/SHORTCODE.jpg exists json 
:saved: 400 Bad Request - "fail" status, message "feedback_required" when accessing https://www.instagram.com/graphql/query?query_hash=f883d95537fbcd400f466f63d42bd8a1&variables=%7B%22id%22%3A4391261632%2C%22first%22%3A12%2C%22after%22%3A%22QVFEbjI2V0o5VGZkaV8xSFBHYzF4Tm83cERHaURleTluSFIxa09sckQ1cXU4S1h5NmxSSXdxTlJQTW1xaTJfMVlhY3U3Um5QYlNOMTFURWlqOXVzWE80RQ%3D%3D%22%7D
Saved session to /Users/USER/.config/instaloader/session-USERNAME.

Errors or warnings occurred:
:saved: 400 Bad Request - "fail" status, message "feedback_required" when accessing https://www.instagram.com/graphql/query?query_hash=f883d95537fbcd400f466f63d42bd8a1&variables=%7B%22id%22%3A4391261632%2C%22first%22%3A12%2C%22after%22%3A%22QVFEbjI2V0o5VGZkaV8xSFBHYzF4Tm83cERHaURleTluSFIxa09sckQ1cXU4S1h5NmxSSXdxTlJQTW1xaTJfMVlhY3U3Um5QYlNOMTFURWlqOXVzWE80RQ%3D%3D%22%7D
```

**Expected behavior**
```
Saved resume information to :saved/iterator_PO_ezfYk.json.xz.

Interrupted by user.
Saved session to /Users/USER/.config/instaloader/session-USERNAME.
```

**Error messages and tracebacks**
If applicable, add error messages and tracebacks to help explain your problem.

**Instaloader version**
[...]

PR #2600, opened 2025-09-05
Save resume information on exception

gnestor:
Fixes https://github.com/instaloader/instaloader/issues/2598

### DIFF
--- a/instaloader/nodeiterator.py
+++ b/instaloader/nodeiterator.py
@@ -2,6 +2,7 @@ import base64
 import hashlib
 import json
 import os
+from .exceptions import QueryReturnedBadRequestException
 from contextlib import contextmanager
 from datetime import datetime, timedelta
 from lzma import LZMAError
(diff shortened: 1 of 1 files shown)


instaloader/nodeiterator.py
```diff
@@ -318,12 +319,14 @@ def resumable_iteration(context: InstaloaderContext,
             context.error("Warning: Not resuming from {}: {}".format(resume_file_path, exc))
     try:
         yield is_resuming, start_index
-    except (KeyboardInterrupt, AbortDownloadException):
+    except (KeyboardInterrupt, AbortDownloadException, QueryReturnedBadRequestException):
```
aandergr, line 322: QueryReturnedBadRequestException is too specific, I think it would be beneficial to save the resume information for other errors as well.

I suggest to instead catch InstaloaderException here, to save the resume information always when the loop body is aborted due to an error.

instaloader/nodeiterator.py
```diff
@@ -318,12 +319,14 @@ def resumable_iteration(context: InstaloaderContext,
...
         if os.path.dirname(resume_file_path):
             os.makedirs(os.path.dirname(resume_file_path), exist_ok=True)
         save(iterator.freeze(), resume_file_path)
         context.log("\nSaved resume information to {}.".format(resume_file_path))
         raise
-    if resume_file_exists:
-        os.unlink(resume_file_path)
-        context.log("Iteration complete, deleted resume information file {}.".format(resume_file_path))
+    finally:
```
aandergr, line 328: No, moving this in a finally block is most likely undesired, it causes the resume file to be deleted right after it is updated.

instaloader/nodeiterator.py
```diff
@@ -2,6 +2,7 @@
 import hashlib
 import json
 import os
+from .exceptions import QueryReturnedBadRequestException
```
aandergr, line 5: There is already a `from .exception import` six lines below, where the new import fits better.

instaloader/nodeiterator.py
```diff
@@ -318,12 +319,14 @@ def resumable_iteration(context: InstaloaderContext,
...
         context.log("\nSaved resume information to {}.".format(resume_file_path))
         raise
-    if resume_file_exists:
-        os.unlink(resume_file_path)
-        context.log("Iteration complete, deleted resume information file {}.".format(resume_file_path))
+    finally:
+        if resume_file_exists:
+            os.unlink(resume_file_path)
+            context.log("Iteration complete, deleted resume information file {}.".format(resume_file_path))
+
```
aandergr, line 332: Unnecessary new newline, please remove from the diff.

aandergr (commented):
Thanks for the PR addressing #2598!

Issue #2512, opened 2025-02-11
Profile.get_followers() doesn't seem to care about RateControl


FalxGod:
My code snippet:
```
L = instaloader.Instaloader(
    download_videos=False,
    download_video_thumbnails=False,
    post_metadata_txt_pattern=''  # Disables creation of caption text files.
)

L.load_session("userName", {
    "csrftoken": "token1",
    "sessionid": "token2",
    "ds_user_id": "token3",
    "mid": "token4",
    "ig_did": "token5"
})

L.download_profile(profile_name, profile_pic=True, fast_update=True)

profile_object = instaloader.Profile.from_username(L.context, profile_name)
follower_iterator = profile_object.get_followers()
followers_list = []
for f in follower_iterator:
    print(f"Found follower: {f.username}")
    followers_list.append(f.username)
```

Produces:
`An error occurred while downloading 'profile_name': 400 Bad Request - "fail" status, message "challenge_required" when accessing {url_here}`

After loging into Insta and completing the challange (clicking the button that says the account is probably scraping) you can use instalaoder again.

-----------------------------------------------------------------------------------------------------------------------------------------------

`L.download_profile(profile_name, profile_pic=True, fast_update=True)` 
works fine even when downloading a profile with several thousand pictures for hours.

`profile_object.get_followers()`
on the other hand managed to get ~110k profile names of followers before softlocking the account with the aforementioned error message

Edit 1:
I will add a delay of 30sec after every 1k profile names of followers have been fetched and report back (probably @ ~1800 CET)

Edit 2: forgot to mention this but I use a modified instaloadercontext.py which changes 
`session.cookies.update({'sessionid': '', 'mid': '', 'ig_pr': '1', 
                                'ig_vw': '1920', 'ig_cb': '1', 'csrftoken': '',
                                's_network': '', 'ds_user_id': ''})`
To:
`session.cookies.update({'mid': '', 'ig_pr': '1', 
                                'ig_vw': '1920', 'ig_cb': '1', 'csrftoken': '',
                                's_network': '', 'ds_user_id': ''})`

Edit 3:
Even if using a manual sleep timer i still get the same error message or this error:
`400 Bad Request - "fail" status, message "checkpoint_required" when accessing`

If this second error shows up and you either refresh Instagram in the browser or login. Instagram says:
"Account comprimised. You apperently gave away your password to get more 'Likes'. Please change your password".
After password change instaloader works again but i recommend to NOT use Profile.get_followers() and iterating over the Node.

PR #2612, opened 2025-10-23
Handle substantional Instagram requirements

denyshon:
When Instagram returns "feedback_required", "checkpoint_required" or "challenge_required", any further requests will be declined until the required action is performed manually. Those messages mean Instagram suspects us of being a bot, so we must stop producing more requests immediately in order not to increase the suspicions.

This PR makes get_json (the status_code 400 block) raise AbortDownloadException with a detailed explanation for Instagram "feedback_required"/"checkpoint_required"/"challenge_required" responses.

I've updated get_json's docstring, but the documentation may need to be rerendered.

Generally, the PR's (almost?) ready to be merged.

I would appreciate a review and some help with docs (if there's something else to be done). You should be able to make edits directly to my branch.

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -395,8 +395,10 @@ class InstaloaderContext:
         :param session: Session to use, or None to use self.session
         :param use_post: Use POST instead of GET to make the request
         :return: Decoded response dictionary
-        :raises AbortDownloadException: When the server responds with 'feedback_required'/'checkpoint_required'/'challenge_required'
-        :raises QueryReturnedBadRequestException: When the server responds with a 400 (and not 'feedback_required'/'checkpoint_required'/'challenge_required').
+        :raises AbortDownloadException: When the server responds with
+            'feedback_required'/'checkpoint_required'/'challenge_required'
+        :raises QueryReturnedBadRequestException: When the server responds with a 400 (and not
+            'feedback_required'/'checkpoint_required'/'challenge_required').
         :raises QueryReturnedNotFoundException: When the server responds with a 404.
         :raises ConnectionException: When query repeatedly failed.
 


instaloader/instaloadercontext.py
```diff
@@ -447,6 +448,19 @@ def get_json(self, path: str, params: Dict[str, Any], host: str = 'www.instagram
                 response_headers.clear()
                 response_headers.update(resp.headers)
             if resp.status_code == 400:
+                # Raise AbortDownloadException in case of substantial Instagram requirements to stop producing more requests
+                content_type = resp.headers.get('Content-Type')
+                if 'application/json' in content_type or not content_type:
+                    try:
+                        resp_json = resp.json()
+                    except ValueError:
+                        pass
```
aandergr: The way this try-except is written will just leave resp_json uninitialized when resp.json() fails, continuing with accesses to an uninitialized variable.

instaloader/instaloadercontext.py
```diff
@@ -447,6 +448,19 @@ def get_json(self, path: str, params: Dict[str, Any], host: str = 'www.instagram
                 response_headers.clear()
                 response_headers.update(resp.headers)
             if resp.status_code == 400:
+                # Raise AbortDownloadException in case of substantial Instagram requirements to stop producing more requests
+                content_type = resp.headers.get('Content-Type')
+                if 'application/json' in content_type or not content_type:
+                    try:
+                        resp_json = resp.json()
+                    except ValueError:
```
aandergr: resp.json() would actually raise json.decoder.JSONDecodeError. Though it is a ValueError, catching JSONDecodeError would make that part of the code more explanatory regarding what exactly is the error we're looking out for.

aandergr (commented):
Thanks for the suggestion. However the implementation itself would introduce some issues which I mentioned in inline comments here.

Given that there already is a _response_error() to build an error message, what would you think about the following way to handle the 400 status code instead?

```python
            if resp.status_code == 400:
                with suppress(json.decoder.JSONDecodeError):
                    if resp.json().get("message") in [
                        "feedback_required",
                        "checkpoint_required",
                        "challenge_required",
                    ]:
                        # Raise AbortDownloadException in case of substantial Instagram
                        # requirements to stop producing more requests
                        raise AbortDownloadException(self._response_error(resp))
                raise QueryReturnedBadRequestException(self._response_error(resp))
```

instaloader/instaloadercontext.py
denyshon: Oops, you are right, there should have been `else`

instaloader/instaloadercontext.py
```diff
@@ -447,6 +448,19 @@ def get_json(self, path: str, params: Dict[str, Any], host: str = 'www.instagram
...
+                        pass
+                    if 'feedback_required' in resp_json.get('message'):
```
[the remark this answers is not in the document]
denyshon: Hmm I'm actually not sure about that, so I would better stick to a superstring version

denyshon:
Would you like me to change the PR accordingly, or would you prefer committing by yourself?

instaloader/instaloadercontext.py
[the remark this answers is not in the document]
aandergr: I just searched the issues to check on that (the message string is printed to the log), we have

* feedback_required: https://github.com/instaloader/instaloader/issues/2616
* checkpoint_required: https://github.com/instaloader/instaloader/issues/2512
* challenge_required: https://github.com/instaloader/instaloader/issues/2512

So actually no superstrings, just the plain "feedback_required"/"checkpoint_required"/"challenge_required" strings.

aandergr:
> Would you like me to change the PR accordingly, or would you prefer committing by yourself?

Thanks! I created a commit with the proposed change.

Issue #2621, opened 2025-11-13
fake 1-frame mp4 from insta based on their hex heading must renamed as fake or alternative jpg files


estatistics:
**Describe the bug**
I found out that insta wrap some still images as mp4 but they are in reality just "JFIF" based on the hex editor. this means that insta by some mechanism that was triggered renamed jpg images to mp4. This usually are 1-frame pics that the user may put some reaction or emoticons or whatever.

**To Reproduce**
e.g. Download "2023-02-22_08-32-11_UTC.mp4" from vauntastic main posts.

**Expected behavior**
Maybe these fake mp4 either renamed as "fake" eg. 2023-02-22_08-32-11_UTC_fake.m4 or as 2023-02-22_08-32-11_UTC_mp4.jpg. Note that it may exists 2023-02-22_08-32-11_UTC.jpg already. 

**Error messages and tracebacks**
ffmprobe 
Input #0, jpeg_pipe, from '2023-02-22_08-32-11_UTC.mp4':
  Duration: N/A, bitrate: N/A
  Stream #0:0: Video: mjpeg (Progressive), yuvj420p(pc, bt470bg/unknown/unknown), 612x612 [SAR 1:1 DAR 1:1], 25 fps, 25 tbr, 25 tbn

**Instaloader version**
instaloader --version
4.14.1

PR #2624, opened 2025-11-24
fake 1-frame mp4 from insta based on their hex heading is renamed as …

sudo-arash:
…fake or alternative jpg files

This is the pull request to fix the issue #2621.

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -8,7 +8,7 @@ from itertools import islice
 from pathlib import Path
 from typing import Any, Callable, Dict, Iterable, Iterator, List, NamedTuple, Optional, Tuple, Union
 from unicodedata import normalize
-
+import requests
 from . import __version__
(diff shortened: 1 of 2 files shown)


instaloader/structures.py
```diff
@@ -160,7 +160,19 @@ def _optional_normalize(string: Optional[str]) -> Optional[str]:
...
     else:
         return None
-
+def is_fake_mp4(url: str) -> bool:
+    """
+    Detect if a video URL is actually a 1-frame JPEG.
+    """
+    try:
+        headers = {"Range": "bytes=0-1023"}  # only first KB
+        r = requests.get(url, headers=headers, timeout=5)
```
aandergr: This introduces a request that is very specific, likely never sent by any browser and would make it trivial to detect Instaloader. This feature must be implemented ideally without a extra request.

PR #2632, opened 2025-12-03
Fix: --count option for profile targets

dubbydubzz:
# Fix --count option for profile targets

## Summary

This PR fixes the `--count` option to work with profile targets (e.g., `instaloader --count 1 profile`). Previously, the `--count` option only worked for hashtags, locations, feeds, and saved posts, but not for regular profiles.

## Problem

The `--count` CLI option should limit the number of posts downloaded, but it wasn't working for profile targets. Running `instaloader --count 1 instagram` would download all available posts instead of stopping at 1.

```bash
instaloader --count 1 instagram
...
Retrieving posts from profile instagram.
[   1/8256]
```

## Solution

Added the missing `max_count=max_count` parameter to `download_profiles()` in `__main__.py`. The `download_profiles()` function already supported this parameter, but it wasn't being passed from the CLI.

```bash
instaloader --count 1 instagram
...
Retrieving posts from profile instagram.
[1/1]
```

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -311,6 +311,7 @@ def _main(instaloader: Instaloader, targetlist: List[str],
             post_filter,
             storyitem_filter,
             latest_stamps=latest_stamps,
+            max_count=max_count,
             reels=download_reels,
         )
         if anonymous_retry_profiles:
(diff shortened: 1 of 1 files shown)


vadimkantorov:
Related:
- https://github.com/instaloader/instaloader/issues/2639

Maybe then worth introducing a new option with a new name which would apply to profiles?

It's very useful for SSG scenarios (e.g. in combination with Astro), where downloading with CLI-only is useful to not add extra custom Python scripts to the mostly JavaScript flow.

vadimkantorov:
@aandergr is there any instaloader's CLI alternative for downloading only most recent several posts (and stories) for a given profile?

PR #2642, opened 2026-01-09
Added --better-output flag for rich CLI experience and progress bars

lukasbecvar:
## Description
This PR introduces a new flag `--better-output` that enhances the CLI experience by utilizing the [rich](https://github.com/Textualize/rich) library. It provides a modern, colorful interface with progress bars, improved logging, and a statistical overview panel for profiles, making the download process visually appealing and easier to monitor.

## Changes
- **New Flag:** Added `--better-output` to enable the rich interface.
- **Rich Integration:**
  - Implemented `InstaloaderContext.progress_context` for handling progress bars.
  - Added smart log formatting with icons (📸, 📹, ✅) and an Instagram-inspired color theme.
  - Integrated a **Profile Stats Panel** that displays follower counts, bio, and post statistics before downloading.
- **Dependency Management:** Added `rich` as an optional dependency in `setup.py` and `Pipfile`. The code gracefully falls back to standard output if `rich` is not installed.
- **Documentation:** Updated `docs/cli-options.rst` with the new option.

## Motivation and Context
The default output, while functional, can be verbose and hard to parse visually during long download sessions. This feature aims to improve the user experience by providing immediate visual feedback on progress and better structuring of information.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -20,3 +20,4 @@ pefile = "*"          # Needed for pyinstaller on Windows
 [packages]
 requests = "*"
 browser-cookie3 = "*"
+rich = "*"
--- a/docs/cli-options.rst
+++ b/docs/cli-options.rst
@@ -355,6 +355,12 @@ Miscellaneous Options
    if login credentials are needed but not given.
    This is handy for running :ref:`instaloader-as-cronjob`.
 
+.. option:: --better-output
+
+   Use rich output (colors, progress bars). Requires "rich" library installed.
+
+   .. versionadded:: 4.15
+
 .. option:: +args.txt
 
    Read arguments from file `args.txt`, a shortcut to provide arguments from
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -521,6 +521,8 @@ def main():
                         help='Disable user interaction, i.e. do not print messages (except errors) and fail '
                              'if login credentials are needed but not given. This makes Instaloader suitable as a '
                              'cron job.')
+    g_misc.add_argument('--better-output', action='store_true',
+                        help='Use rich output (colors, progress bars). Requires "rich" library installed.')
     g_misc.add_argument('-h', '--help', action='help', help='Show this help message and exit.')
     g_misc.add_argument('--version', action='version', help='Show version number and exit.',
                         version=__version__)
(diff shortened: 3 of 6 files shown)


aandergr:
Wow this is cool! This gives Instaloader finally a modern console output.

I also like the Profile Stats Panel and the progress bar. The time remaining will cause confusion, since Instaloader sometimes interrupts the download waiting for several minutes, messing up with the remaining time estimation, but I still tend to keep it, since in most cases this provides a good estimation.

I suggest that instead of enabling it only with `--better-output`, the better output shall be enabled if stdout is a TTY ([`isatty()`](https://docs.python.org/3/library/io.html#io.IOBase.isatty)) and rich is installed (and of course, `--quiet` is not applied).  It seems, this can be already handled by rich itself: [`is_terminal`](https://rich.readthedocs.io/en/latest/reference/console.html#rich.console.Console.is_terminal). Users then still have the option of disabling the better output by piping it through `cat`.

I would also suggest to go the other way around: That the "old" output logging is replaced by the better output and only certain elements are removed if stdout is not a TTY, such as the colors, the spinners, the progress bars – the kind of output that does not work properly when piped into a logfile. This would simplify maintaining the console output in the long run. This also implies that the dependency is non-optional. Additionally, it would most likely remove the need of `_style_log_message()`, in favor of a more rich-native approach.

The change shouuld be based on the branch `upcoming/v4.16`, since it is a new feature.

xurizaemon:
It looks great! May I suggest a small change, reconsidering the flag / switch name? I think it can be clearer and more specific.

IMO `--better-output` is _relative_ (comparing current output with the changed version) and _not specific_ (eg for non-TTY / piped output, while this would still be fancy, pretty, or rich, it would not be better).

Referring to existing [CLI options](https://instaloader.github.io/cli-options.html), the comparable misc flag is `--quiet`, which suppresses output.

As clearer flag, I'd suggest:

- `--pretty` is well-established CLI parameter in other tools
- `--ansi` is common too
- `--output=<pretty|plain|json>` style allows lots of room for expansion ...
- `--rich` is nice because it references Rich 🎨 

(Of course Instaloader could support multiple forms, eg allowing for `--pretty`, `--quiet` as well as `--output=<pretty|quiet|text|...>`.)

I see @aandergr has above suggested making this default behaviour. For folks executing from cron or parsing the existing output, then perhaps `--no-pretty` or `--output=plain` would make sense when disabling this.

Issue #2655, opened 2026-02-16
Persistent 429 Too Many Requests error


huntermilo1:
**Description**
I am consistently encountering a 429 Too Many Requests error when attempting to access api/v1/users/web_profile_info/. This issue has persisted for 3–4 days. I am not a developer, but I have attempted several standard troubleshooting steps without success.

**Error Log**
JSON Query to api/v1/users/web_profile_info/?username=profile: 429 Too Many Requests
Instagram responded with HTTP error "429 - Too Many Requests".
The request will be retried in 30 minutes.

Steps I have already taken:

- Switched networks to obtain a new IP address.
- Completely uninstalled and reinstalled Instaloader.
- Ensured no Instagram apps or browsers were running in the background.
- Stayed logged out/inactive on Instagram for over 60 minutes before retrying.
- Verified that no parallel instances of Instaloader are running.

**Environment:**
Tool: Instaloader
Occurrence: Persistent for 72+ hours.

**Additional Context:**
Even with a fresh IP and significant cooldown periods, the error triggers immediately upon the first request. It appears the web_profile_info endpoint is being aggressively throttled or requires a specific session handling update.

rohal12:
same!

stxmp:
This is a duplicate of this issue: https://github.com/instaloader/instaloader/issues/2651

A fix is posted in this pr: https://github.com/instaloader/instaloader/pull/2652

huntermilo1:
> This is a duplicate of this issue: [#2651](https://github.com/instaloader/instaloader/issues/2651)
> 
> A fix is posted in this pr: [#2652](https://github.com/instaloader/instaloader/pull/2652)

unfortunately this has not fixed the issue for me, no matter what i do, I'm still getting that error.

puer89:
Broken for me too and this does not fix

> This is a duplicate of this issue: [#2651](https://github.com/instaloader/instaloader/issues/2651)
> 
> A fix is posted in this pr: [#2652](https://github.com/instaloader/instaloader/pull/2652)

yuyoiuyoiu:
> > This is a duplicate of this issue: [#2651](https://github.com/instaloader/instaloader/issues/2651)
> > A fix is posted in this pr: [#2652](https://github.com/instaloader/instaloader/pull/2652)
> 
> unfortunately this has not fixed the issue for me, no matter what i do, I'm still getting that error.

[...]

PR #2652, opened 2026-02-03
Fix Profile API by migrating to GraphQL endpoints

cxyfer:
## Summary

Fixes #2651 - Resolves persistent 429 errors and "useragent mismatch" issues when fetching profile data by migrating from the deprecated `web_profile_info` API to GraphQL endpoints.

## Problem

The `web_profile_info` API endpoint has become unreliable, frequently returning:
- 429 Too Many Requests
- "useragent mismatch" errors
- Inconsistent behavior even with valid sessions

This affects `Profile.from_username()` and `Profile._obtain_metadata()`, blocking basic profile operations.

## Solution

- **`Profile.from_username()`**: Now uses `TopSearchResults` to fetch profile data, with fallback to GraphQL doc_id `34579740524958711` (PolarisProfilePostsQuery) for short usernames that may not appear in search results
- **`Profile._obtain_metadata()`**: Now uses GraphQL doc_id `25980296051578533` (PolarisProfilePageContentQuery) to fetch profile metadata
- Added `_normalize_profile_data()` to map new API response structure to existing Profile fields
- Maintains backward compatibility with existing code

## Testing

⚠️ **This fix requires authenticated sessions to work reliably.**

Tested with authenticated session:
- Public profile post downloads
- Public profile story downloads

Additional testing on edge cases (private profiles, profiles without posts, etc.) would be appreciated.

## Breaking Changes

None. The changes are internal API migrations that maintain the existing public interface.

Issue #2651, opened 2026-02-02
Title: Persistent 429 Error and "Useragent Mismatch"


Squidiis:
I'm experiencing a very inconsistent behavior with Instaloader that I can't seem to pin down. I'm frequently hitting 429 Too Many Requests along with the useragent mismatch error.

The Issue: The script fails intermittently. Sometimes it works perfectly, and minutes later—without any changes to the code or environment—it gets blocked.

Error Message:

```
JSON Query to api/v1/users/web_profile_info/?username=...: 429 Too Many Requests
{"message":"useragent mismatch","status":"fail"}
```

Environments Tested (All show the same inconsistent behavior):
- Local Machine: Private IP, real browser user-agents.
- VPS: Data-center IP (blocked more frequently, but works sometimes).
- Proxies: Residential proxies used to bypass IP bans.
- Authentication: Tested both with a valid session.cos and without login.

Key Observation: The most frustrating part is the inconsistency. One request might pass, and the very next one (or one five minutes later) fails with a useragent mismatch. This happens even when the User-Agent string is kept constant and matches the session.

```py
import instaloader

def get_latest_posts(username, count=5):
    loader = instaloader.Instaloader()

    try:
        profile = instaloader.Profile.from_username(loader.context, username)
        
        print(f"--- Latest {count} posts from {username} ---")
        for i, post in enumerate(profile.get_posts()):
            if i >= count:
                break
            print(f"Post #{i+1} | Link: https://www.instagram.com/p/{post.shortcode}/")

    except Exception as e:
        print(f"Scraping error: {e}")

if __name__ == "__main__":
    get_latest_posts("", count=5)
```

**Observations:**
- Mismatch Error: The useragent mismatch suggests that even when a session is loaded, the headers sent by Instaloader might not align with what the web_profile_info endpoint currently expects.
- Timing: Since it works "sometimes," it feels like Instagram's server-side protection is using a very aggressive reputation/scoring system or a sliding window for requests that don't look 100% like a real browser.
- App-ID: I suspect the missing x-ig-app-id or incorrect referer headers might be the trigger.
**Questions:**
- Is there a way to manually inject mandatory headers (like x-ig-app-id: 936619743392459) into the Profile.from_username request?
- Why does it work intermittently on the same IP without any changes?
- Is the web_profile_info endpoint still viable for headless/VPS scraping without a full browser emulation?

yuyoiuyoiu:
same here : (

[...]

Issue #2649, opened 2026-01-28
error / 401 / "FAIL"


kevingan10:
Regards, I'm having trouble downloading someone's Instagram follow list.

JSON Query to graphql/query: 401 Unauthorized - "fail" status, message "Please wait a few minutes before you try again." when accessing https://www.instagram.com/graphql/query?query_hash=58712303d941c6855d4e888c5f0cd22f&variables=%7B%22id%22%3A%222378304718%22%2C%22first%22%3A12%7D [retrying; skip with ^C]
JSON Query to graphql/query: 401 Unauthorized - "fail" status, message "Please wait a few minutes before you try again." when accessing https://www.instagram.com/graphql/query?query_hash=58712303d941c6855d4e888c5f0cd22f&variables=%7B%22id%22%3A%222378304718%22%2C%22first%22%3A12%7D [retrying; skip with ^C]
Traceback (most recent call last):
  File "C:\Users\Human\AppData\Local\Programs\Python\Python314\Lib\site-packages\instaloader\instaloadercontext.py", line 468, in get_json
    raise ConnectionException(self._response_error(resp))
instaloader.exceptions.ConnectionException: 401 Unauthorized - "fail" status, message "Please wait a few minutes before you try again." when accessing https://www.instagram.com/graphql/query?query_hash=58712303d941c6855d4e888c5f0cd22f&variables=%7B%22id%22%3A%222378304718%22%2C%22first%22%3A12%7D

I would like to ask for your feedback and solutions. Thanks.

tomballgithub:
Looks to me like your account is flagged. Have you tried logging in with it to see if there is a captcha or other warning message about being flagged as an automated tool? @kevingan10

ScribbleGhost:
I also get this when using a cookie file.

eholzbach:
Same. I can login from a browser without issue.

kjcole:
I just tried `instaloader` for the first time after an `apt install`. (`pip` is becoming more annoying to use these days.):

```
$ dpkg -l instaloader 
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version      Architecture Description
+++-==============-============-============-====================================
ii  instaloader    4.10.3+ds-1  all          Instagram automatic photo downloader

[...]

PR #2663, opened 2026-02-26
Fix get_posts() — replace revoked doc_id 7898261790222653 with 34579740524958711

dandaka:
## Problem

\`Profile.get_posts()\` is broken for all users globally. The \`doc_id=7898261790222653\` used for the logged-in path returns \`401 Unauthorized - "Please wait a few minutes before you try again."\` regardless of session validity, IP address, or account.

Related issues: #2637, #2649, #2651, #2655

## Root cause

Instagram revoked \`doc_id=7898261790222653\`. The endpoint still accepts the request but rejects the specific doc_id with a 401.

## Fix

Replace with \`doc_id=34579740524958711\` — the same endpoint (\`xdt_api__v1__feed__user_timeline_graphql_connection\`) already used by the fix in PR #2652 for \`from_username()\`. Nodes are returned in iphone_struct format, compatible with the existing \`Post.from_iphone_struct()\` wrapper.

This also simplifies the logged-in/anonymous branch: the new doc_id uses username-based variables (same as the timeline feed query), so the anonymous fallback path is no longer needed — Instagram no longer embeds post edges in the initial profile response anyway.

## Testing

Verified against \`@miamipadelcenter\` and \`@natgeo\` with authenticated cookie session:
- Profile fetched successfully
- \`get_posts()\` returns posts without 401
- \`download_post()\` saves \`.jpg\` and \`.txt\` files correctly

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -1206,33 +1206,26 @@ class Profile:
 
         :rtype: NodeIterator[Post]"""
         self._obtain_metadata()
-        logged_in = self._context.is_logged_in
+        # doc_id 7898261790222653 was revoked by Instagram (returns 401 for all users).
+        # Use 34579740524958711 instead — same fix approach as PR #2652.
         return NodeIterator(
             context=self._context,
-            edge_extractor=(
-                (lambda d: d["data"]["xdt_api__v1__feed__user_timeline_graphql_connection"])
-                if logged_in
-                else (lambda d: d["data"]["user"]["edge_owner_to_timeline_media"])
-            ),
-            node_wrapper=(
-                (lambda n: Post.from_iphone_struct(self._context, n))
-                if logged_in
-                else (lambda n: Post(self._context, n, self))
-            ),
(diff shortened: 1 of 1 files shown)


Talleyrand-34:
I tried this version and is also broken for me

aandergr:
Changing the query causes some drawbacks, one of them is that all frozen get_posts() node iterators (resume information) would become invalid, and it introduces additional uncertainty/risk. This would of course be obvious if the current method no longer worked, but to me it seems that the current get_posts() with doc_id=7898261790222653 still works, and I do not see any other credible reports or indications that it is revoked.

The linked issues describe observations regarding get_stories() and Profile.from_username(); they do not appear to be related to the get_posts() method that is changed here.

So, as of now, I do not see the need to merge this PR, but I am very open and curious about any rationale that I may be missing. Regardless, it is a nice and valuable finding that the get_posts() query can also be done with an alternative doc_id, and indeed the get_posts() proposed here seems to work fine. Thanks for sharing that with us!

PR #2676, opened 2026-03-21
Fix persistent 429/401 errors via HTTP/2 adapter and updated doc_id

irishpadres:
## Summary

**Note**: PR #2677 is an alternative implementation of the same fix using https://github.com/jawah/niquests instead of a custom httpx adapter — cleaner if the maintainers are open to swapping the requests dependency.

Fixes the persistent 429 Too Many Requests errors reported in #2655 and the 401 errors from the revoked `doc_id` in `get_posts()`.

- **HTTP/2 adapter**: Adds an `Http2Adapter` class (backed by `httpx`) that upgrades all Instagram requests to HTTP/2. Instagram's `web_profile_info` and related endpoints return 429 on some hosts/networks (NAS devices, containers, certain ISPs) under HTTP/1.1 but succeed reliably with HTTP/2. Confirmed working where PR #2652 alone was insufficient.
- **Empty-cookie stripping**: The adapter filters out empty-value placeholder cookies before sending — these confuse Instagram's CSRF/session checks and can cause spurious 401s.
- **Updated `doc_id` in `get_posts()`**: `7898261790222653` was revoked by Instagram and returns 401 for all users. Replaced with `34579740524958711`. Removed the now-broken anonymous fallback path since this endpoint requires a logged-in session.
- **Resilient `test_login()`**: Returns the stored username on connection error instead of `None`, so a transient rate-limit failure doesn't trigger an interactive re-login prompt for a valid session.
- **New dependency**: `httpx[http2]>=0.24` added to `install_requires`.

## Test plan

- [ ] `pip install git+https://github.com/irishpadres/instaloader.git` and run `instaloader` against a profile that previously returned 429 immediately
- [ ] Verify `get_posts()` works for a logged-in session (no 401 on first page fetch)
- [ ] Verify loading a saved session file still works and HTTP/2 adapter is active
- [ ] Verify that a connection error during `test_login()` does not force re-login when a valid session is loaded

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -1266,8 +1266,8 @@ class Profile:
                 "data": {
                     "count": 12,
                     "include_relationship_info": True,
-                    "latest_besties_reel_media": False,
-                    "latest_reel_media": False,
+                    "latest_besties_reel_media": True,
+                    "latest_reel_media": True,
                 },
                 "username": self.username,
             },
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -1258,33 +1258,25 @@ class Profile:
 
         :rtype: NodeIterator[Post]"""
         self._obtain_metadata()
-        logged_in = self._context.is_logged_in
         return NodeIterator(
             context=self._context,
-            edge_extractor=(
-                (lambda d: d["data"]["xdt_api__v1__feed__user_timeline_graphql_connection"])
-                if logged_in
-                else (lambda d: d["data"]["user"]["edge_owner_to_timeline_media"])
-            ),
-            node_wrapper=(
-                (lambda n: Post.from_iphone_struct(self._context, n))
-                if logged_in
-                else (lambda n: Post(self._context, n, self))
-            ),
+            edge_extractor=lambda d: d["data"]["xdt_api__v1__feed__user_timeline_graphql_connection"],
+            node_wrapper=lambda n: Post.from_iphone_struct(self._context, n),
             query_variables={
                 "data": {
                     "count": 12,
                     "include_relationship_info": True,
-                    "latest_besties_reel_media": True,
-                    "latest_reel_media": True,
+                    "latest_besties_reel_media": False,
+                    "latest_reel_media": False,
                 },
-                **({"username": self.username} if logged_in else {"id": self.userid}),
+                "username": self.username,
             },
             query_referer="https://www.instagram.com/{0}/".format(self.username),
             is_first=Profile._make_is_newest_checker(),
-            doc_id="7898261790222653" if logged_in else "7950326061742207",
+            # doc_id 7898261790222653 was revoked by Instagram (returns 401).
+            # 34579740524958711 is the current replacement.
+            doc_id="34579740524958711",
             query_hash=None,
-            first_data=(None if logged_in else self._metadata("edge_owner_to_timeline_media")),
         )
 
     def get_saved_posts(self) -> NodeIterator[Post]:
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -355,6 +355,7 @@ class InstaloaderContext:
         # pylint:disable=protected-access
         http.client._MAXHEADERS = 200
         session = requests.Session()
+        session.mount("https://", Http2Adapter())
         session.cookies.update({'sessionid': '', 'mid': '', 'ig_pr': '1',
                                 'ig_vw': '1920', 'ig_cb': '1', 'csrftoken': '',
                                 's_network': '', 'ds_user_id': ''})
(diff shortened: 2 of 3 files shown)


irishpadres:
Rebased onto v4.15.1 (which merges #2652). The doc_id fix and HTTP/2 adapter patches still apply — upstream 4.15.1 does not address the revoked `doc_id=7898261790222653` in `get_posts()`.

instaloader/structures.py
```diff
@@ -1258,33 +1258,25 @@ def get_posts(self) -> NodeIterator[Post]:
...
                 },
-                **({"username": self.username} if logged_in else {"id": self.userid}),
+                "username": self.username,
             },
             query_referer="https://www.instagram.com/{0}/".format(self.username),
             is_first=Profile._make_is_newest_checker(),
-            doc_id="7898261790222653" if logged_in else "7950326061742207",
+            # doc_id 7898261790222653 was revoked by Instagram (returns 401).
+            # 34579740524958711 is the current replacement.
+            doc_id="34579740524958711",
```
Set27, line 1278: just interested, how do u get this number?

irishpadres: It came from PR #2663

irishpadres: And #2652 which is already merged

aandergr:
Thank you for the suggestions. This PR bundles four independent concerns into a single change, which unfortunately makes it hard to review, validate, or selectively merge each one on its own. 

**1. HTTP/2 adapter (429 errors on specific network configurations)**
This may address a real problem for a subset of users. It introduces `httpx[http2]` as a mandatory new dependency for everyone, and the implementation heavily changes the transport layer. This deserves its own focused discussion before being accepted.

**2. Empty-cookie stripping (spurious 401s)**
There is a long-standing design quirk where domain-less placeholder cookies are inserted into the session jar at login. However, this code has existed for over 8 years without being a reported cause of 401s, and it is unclear whether it is actually the cause for the observed errors. This deserves a dedicated investigation.

**3. `doc_id` update in `get_posts()`**
The claim that `doc_id=7898261790222653` is revoked is not true: `get_posts()` works correctly with it for logged-in sessions. Additionally, issues #2655/#2651 referenced here have already been fixed in a recent commit.

**4. Resilient `test_login()` on connection errors**
The intent is valid, a connection error should not ultimately be interpreted as not being logged-in. However, the proposed fix is wrong in the opposite direction. It returns the stored username without having verified the session, which defeats the purpose of `test_login()`. This needs a more careful solution, also to be addressed separately.

irishpadres:
@aandergr Thanks, that feedback makes sense. I agree this PR bundles too many independent concerns, so I’m fine abandoning #2676 rather than trying to keep expanding it.

One clarification on the motivation: #2655 was closed as a duplicate of #2651, and #2652 addressed that class of failures by moving profile metadata away from `web_profile_info`. However, the persistent 429 behavior reported in #2655 was still reproducible for some users (including me) after that work, which is what led to these follow-up PRs.

For the HTTP/2 piece, I opened #2677 as a narrower alternative that avoids the custom `httpx` adapter and uses `niquests` instead, based on the suggestion in #2655. I’m happy to continue that discussion there if maintainers are open to evaluating a transport-layer change separately.

Before making further changes, I’ll re-test latest `upstream/master` without either PR’s changes, especially logged-in `Profile.get_posts()` with the existing `doc_id=7898261790222653`, so we can avoid carrying fixes that are no longer needed or not well supported by current behavior.

irishpadres:
Thanks @aandergr — closing this. Responding to each point, with results from re-testing on master:

  1. HTTP/2 adapter. Withdrawing. I re-tested from the same home server where HTTP/1.1 was previously blocked on the
  first request regardless of cooldown or IP rotation. With the current doc_ids, plain HTTP/1.1 now completes the full
  profile-metadata + get_posts workflow with no 429 — identical to HTTP/2. The host-level block that originally forced
  HTTP/2 appears to have lifted (Instagram's flagging was evidently temporal), so I can no longer reproduce the
  condition that motivated it. If it recurs, an opt-in flag would be the right shape rather than a forced transport
  change — but I'm not pushing it now.

  2. Empty-cookie stripping. Dropping it — I can't substantiate that the placeholder cookies cause 401s.

  3. doc_id in get_posts(). You're right. I re-tested logged-in get_posts() with the existing doc_id=7898261790222653
  and it works correctly, returning the same posts as the replacement value. Withdrawing.

  4. Resilient test_login(). Agreed the implementation is backwards — returning the stored username without verifying
  defeats the purpose. The correct fix is to distinguish a genuine auth failure from a transient ConnectionException and
  only re-raise/retry on the latter. Leaving that to a separate, careful PR.

  Net: the only durable issue here was doc_id rotation. The current live one is profile metadata in _obtain_metadata()
  (#2695), cleanly fixed by the single-concern #2696

PR #2677, opened 2026-03-21
Alternative HTTP/2 fix: replace requests with niquests

irishpadres:
## Summary

This is an alternative to PR #2676 for fixing the persistent 429 errors reported in #2655.

Rather than adding a custom `Http2Adapter` shim on top of `requests`, this swaps the `requests` dependency for [`niquests`](https://github.com/jawah/niquests) — a drop-in fork that enables HTTP/2 (and HTTP/3) by default with no additional configuration.

**Changes:**
- `import niquests as requests` in `instaloadercontext.py` and `instaloader.py`
- Remove the `Http2Adapter`, `_Http2HeaderMsg`, `_Http2OriginalResponse`, `_Http2RawResponse` classes (~80 lines)
- Remove `session.mount()` calls (HTTP/2 is automatic)
- Replace `httpx[http2]>=0.24` with `niquests>=3.9` in `install_requires`

## Relationship to PR #2676

PR #2676 solves the same problem with a more conservative approach (adds `httpx` as a second dependency, keeps `requests`). This PR is the cleaner alternative — fewer lines, one dependency instead of two — but it does replace a well-established dependency with a fork, which is a bigger ask.

Both include the same `doc_id` fix for `get_posts()` and the `test_login()` resilience improvement.

Leaving it to the maintainers to decide which approach they prefer.

## Test plan

- [ ] Load existing session file and verify login
- [ ] `Profile.from_username()` resolves without 429
- [ ] `profile.get_posts()` returns posts without 401
- [ ] Verify HTTP/2 is being used (e.g. via `response.http_version`)

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -15,7 +15,7 @@ from pathlib import Path
 from typing import Any, Callable, IO, Iterator, List, Optional, Set, Union, cast
 from urllib.parse import urlparse
 
-import requests
+import niquests as requests
 import urllib3  # type: ignore
 
 from .exceptions import *
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -355,6 +355,7 @@ class InstaloaderContext:
         # pylint:disable=protected-access
         http.client._MAXHEADERS = 200
         session = requests.Session()
+        session.mount("https://", Http2Adapter())
         session.cookies.update({'sessionid': '', 'mid': '', 'ig_pr': '1',
                                 'ig_vw': '1920', 'ig_cb': '1', 'csrftoken': '',
                                 's_network': '', 'ds_user_id': ''})
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -13,105 +13,19 @@ from datetime import datetime, timedelta
 from functools import partial
 from typing import Any, Callable, Dict, Iterator, List, Optional, Union
 
-import httpx
-import requests
-import requests.cookies
-import requests.utils
-
-from .exceptions import *
-
-
-class _Http2HeaderMsg:
-    """Mimics the httplib.HTTPMessage interface that requests' MockResponse.info() expects."""
-    def __init__(self, httpx_headers):
-        self._h = httpx_headers
-
-    def get_all(self, name, default=None):
-        values = self._h.get_list(name.lower())
-        return values if values else default
-
-    def getheaders(self, name):
-        return self._h.get_list(name.lower())
-
-
-class _Http2OriginalResponse:
-    """Mimics urllib3.HTTPResponse so that requests' extract_cookies_to_jar guard
-    (checks for _original_response.msg) passes and can read Set-Cookie headers."""
-    def __init__(self, httpx_headers):
-        self.msg = _Http2HeaderMsg(httpx_headers)
-
-
-class _Http2RawResponse:
-    """Wraps an httpx response so that requests' Session.send() can extract
-    Set-Cookie headers from it into the session's cookie jar via
-    extract_cookies_to_jar(self.cookies, request, r.raw)."""
-    def __init__(self, httpx_headers):
-        self._original_response = _Http2OriginalResponse(httpx_headers)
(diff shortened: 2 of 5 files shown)


tomballgithub:
Can this be used with today's release?

irishpadres:
Rebased onto v4.15.1 (which merges #2652). The doc_id fix and niquests swap still apply — upstream 4.15.1 does not address the revoked `doc_id=7898261790222653` in `get_posts()`.

tomballgithub:
There is an issue with this PR. The instaloadercontext.py file had the following line removed via the PR. The line was left there on the other PR #2676:
`from .exceptions import *`

Which causes an exception when an error occurs:
```
  File "C:\Python\Lib\site-packages\instaloader\instaloadercontext.py", line 437, in get_json
    raise AbortDownloadException("Query to https://{}/{} responded with \"{} {}\"{}{}".format(
          ^^^^^^^^^^^^^^^^^^^^^^
NameError: name 'AbortDownloadException' is not defined

During handling of the above exception, another exception occurred:

[redacted]

  File "C:\Python\Lib\site-packages\instaloader\instaloadercontext.py", line 560, in doc_id_graphql_query
    resp_json = self.get_json('graphql/query',
                              params={'variables': variables_json,
    ...<2 lines>...
                              session=tmpsession,
                              use_post=True)
  File "C:\Python\Lib\site-packages\instaloader\instaloadercontext.py", line 479, in get_json
    except (ConnectionException, json.decoder.JSONDecodeError, requests.exceptions.RequestException) as err:
            ^^^^^^^^^^^^^^^^^^^
NameError: name 'ConnectionException' is not defined
```

Debug message:
`Query to https://www.instagram.com/graphql/query responded with "400 Bad Request": {"message":"checkpoint_required","checkpoint_url":"https://www.instagram.com/challenge/?next=/graphql/query","lock":false,"flow_render_type":0,"status":"fail"}`

irishpadres:
Good catch — `from .exceptions import *` was accidentally dropped when replacing the `requests` imports. Fixed in the latest push.

javierdiaz94:
I was having 403 json issues... I apply this fix... and now 

[CTX] @instagramaccount
  Proxy: http://8.211.200.183:1080
JSON Query to graphql/query: 403 Forbidden when accessing https://www.instagram.com/graphql/query [retrying; skip with ^C]
  [ERROR] @isntagramaccount: Profile alcguanipa does not exist.
^CTraceback (most recent call last):
  File "/home/xxxxx", line 244, in instagram_Scrapper
    profile = instaloader.Profile.from_username(L.context, username)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/axxxpython3.12/site-packages/instaloader/structures.py", line 919, in from_username
    raise ProfileNotExistsException("Profile {} does not exist.".format(username))
instaloader.exceptions.ProfileNotExistsException: Profile alcguanipa does not exist.

Now doesn't detect de user... what could I do

irishpadres:
@javierdiaz94   This looks like it may be unrelated to the HTTP/2 change — the 403 is coming from a proxy (http://8.211.200.183:1080) which Instagram appears to be blocking. That IP resolves to Alibaba Cloud (AliCloud). Instagram may be blocking proxy servers from datacenters or cloud providers.

A few things worth trying:
1. Test without the proxy
2. Check if the proxy supports HTTP/2 — niquests uses HTTP/2 by default, and some proxies don't handle it correctly. You can try forcing HTTP/1.1 by setting NIQUESTS_DISABLE_HTTP2=1 in your environment as a quick test
3. Try a residential proxy — datacenter IPs are much more likely to be blocked by Instagram

tomballgithub:
I've been running this PR since inception with no issues except 403's only at initial login.

irishpadres:
Withdrawing this alongside #2676. I re-tested from the home server that originally couldn't get past HTTP/1.1 at all: with current doc_ids, plain HTTP/1.1 now completes the full workflow with no 429, identical to the niquests/HTTP-2 build. The host-level block that motivated this has lifted, so the transport swap isn't justified. The real fix throughout was the doc_id updates (current live one: #2696). Thanks.

PR #2687, opened 2026-04-16
Modernize type hinting and implement CI/CD infrastructure

Jorgeotero1998:
This PR introduces a significant architectural modernization to improve code maintainability and developer experience. The focus was on implementing strict type safety in core modules and establishing a robust automated quality gate.

🛠 Engineering Changes
Structural Type Hinting: Added comprehensive type annotations to critical methods in instaloader/nodeiterator.py and instaloader/structures.py, specifically targeting __iter__, __next__, and count methods.

CI/CD Pipeline: Integrated a new GitHub Actions workflow (ci.yml) that automatically triggers on every push/PR to run linting and type checking across Python 3.11+.

Automated Quality Control: - Configured Ruff for high-performance linting and automated code formatting.

Set up Mypy with strict local configurations to ensure type consistency.

Code Standards: Performed a global cleanup of imports and legacy formatting patterns to align with modern PEP 8 and PEP 484 standards.

📈 Impact
Reduced Technical Debt: Modern type hints allow for better IDE introspection and early bug detection.

Infrastructure: The new CI pipeline ensures that future contributions maintain the project's quality standards without manual oversight.

✅ Checklist
[x] Type hints are compatible with Python 3.9+ standards.

[x] Ruff linting and formatting pass successfully.

[x] Mypy static analysis completed with zero errors in core modules.

[x] GitHub Actions workflow verified and functional.

PR #2699, opened 2026-06-11
Derive title pic file extension from Content-Type instead of hardcoding .jpg

### LINKED ISSUE #2686: Instaloader mislabeling new webp profile pictures as jpg
**Describe the bug**
Instaloader is mislabeling new webp profile pictures as jpg

**To Reproduce**
Scrape any profile to see that the webp images are being saved with jpg file extensions

**Expected behavior**
Instaloader should save the image in its native format of webp

N/A

**Instaloader version**
commit 66165a9

**Additional context**

ennsharma:
Fixes #2686.

Instagram now serves some profile pictures as webp, but `download_title_pic()` unconditionally appends `.jpg`, so webp profile pictures (and highlight covers / hashtag pictures, which share this code path) get saved with the wrong extension.

This change derives the extension the same way `download_pic()` already does for post media:

1. from the response's `Content-Type` header (normalizing `jpeg` → `jpg`),
2. falling back to the URL filename's extension,
3. falling back to `jpg` (previous behavior) if neither is available.

Existing jpg downloads are unaffected — `image/jpeg` responses still produce `.jpg` filenames. A previously-downloaded picture whose served format changed from jpg to webp will be re-downloaded once under its correct extension, which mirrors how `download_pic()` behaves.

I verified the three derivation paths (webp Content-Type, jpeg Content-Type, missing Content-Type with/without URL extension) against a mocked response; `pylint` (10.00) and `mypy` are clean on the touched file.

Note: #2687 previously bundled a fix for this issue into a larger type-hinting/CI PR that was closed — this PR is intentionally scoped to just the extension bug.

### DIFF
--- a/instaloader/instaloader.py
+++ b/instaloader/instaloader.py
@@ -521,7 +521,11 @@ class Instaloader:
         """Downloads and saves a picture that does not have an association with a Post or StoryItem, such as a
         Profile picture or a Highlight cover picture. Modification time is taken from the HTTP response headers.
 
-        .. versionadded:: 4.3"""
+        .. versionadded:: 4.3
+
+        .. versionchanged:: 4.16
+           The file extension is derived from the response's Content-Type header (falling back to the URL),
+           rather than always being ``.jpg``."""
 
         http_response = self.context.get_raw(url)
         date_object: Optional[datetime] = None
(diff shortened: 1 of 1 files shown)


aandergr:
This may fix the observation, but in a way that appears quite brittle. I suggest to use the standard library's [guess_extension()](https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_extension) preferably, and ".jpg" as fallback. What do you think?

aandergr:
The topic of file extension handling needs a bit more than this change to be properly resolved.

This PR only addresses download_title_pic(). However, there is also download_pic(), which derives the extension from the remote filename. And there is the _already_downloaded() logic in download_post() and download_storyitem(), which has ".jpg" hardcoded.

PR #2696, opened 2026-05-22
Fix profile metadata GraphQL query flags

npiriou:
Fixes the profile metadata request after Instagram-side changes.

Changes:
- switch to GraphQL doc id 27937681195819736
- add __relay_internal__pv__PolarisWebSchoolsEnabledrelayprovider = False
- add enable_integrity_filters = True

Credit: this patch is based on the workaround shared by [@VupVup in issue #2695](https://github.com/instaloader/instaloader/issues/2695#issuecomment-4495553287).

PR #2701, opened 2026-06-16
Fix profile resolution after Instagram retired its GraphQL profile queries

### LINKED ISSUE #2682: 403 Forbidden on graphql/query leading to incorrect "Profile does not exist" error
**Describe the bug**

After updating to Instaloader 4.15.1, calling `Profile.from_username()` frequently results in a `403 Forbidden` error on the `graphql/query` endpoint, which then incorrectly raises `ProfileNotExistsException` even for well-known public profiles.

This issue is much more frequent than in previous versions

**To Reproduce**

```python
import instaloader

L = instaloader.Instaloader()

# Test with a very popular public profile
profile = instaloader.Profile.from_username(L.context, "zuck")

for post in profile.get_posts():
    print(post.shortcode)
```

**Expected behavior**

The profile should load successfully and `get_posts()` should work without triggering 403 errors so often, and definitely without throwing "Profile does not exist" for existing public accounts.

```
JSON Query to graphql/query: 403 Forbidden when accessing 
https://www.instagram.com/graphql/query [retrying; skip with ^C]

  File "...", line 4, in <module>
    profile = instaloader.Profile.from_username(L.context, "zuck")
  File "/.../instaloader/structures.py", line 919, in from_username
    raise ProfileNotExistsException("Profile {} does not exist.".format(username))
instaloader.exceptions.ProfileNotExistsException: Profile zuck does not exist.
```

**Instaloader version**

```bash
instaloader --version
→ 4.15.1
```

**Additional context**

- Using anonymous sessions
- The same code was noticeably more stable before updating to 4.15.1.
- The error occurs both during profile loading and during `get_posts()` pagination.
- Even high-profile accounts like `@zuck` (Mark Zuckerberg) are affected.
- This behavior started immediately after the update to 4.15.1.

It appears the new GraphQL query handling or doc_id changes in this version are being blocked more aggressively by Instagram.

### LINKED ISSUE #2688: Workaround: fall back to web_profile_info when the fbsearch doc_id query returns no data
Related to #2682, but framing this as a workaround proposal rather than another bug report.

### What I'm seeing

`Profile.from_username` hits the `doc_id=26347858941511777` GraphQL query and never finds the user in the response, so every call raises `ProfileNotExistsException`, even for obviously-existing profiles like `natgeo` or my own account.

Depending on the session state, the underlying response differs:

- **Anonymous**: `403 Forbidden` on `graphql/query` (this is what #2682 describes).
- **Logged in**: `200 OK` with this payload:
  ```json
  {
    "status": "ok",
    "errors": [{"message": "execution error", "path": ["xdt_api__v1__fbsearch__non_profiled_serp"], "severity": "CRITICAL"}],
    "data": null
  }
  ```

Same end result either way. `data["xdt_api__v1__fbsearch__non_profiled_serp"]` doesn't exist, so the loop in `from_username` never matches and `ProfileNotExistsException` fires. From the caller's side it looks like the profile genuinely doesn't exist, which is misleading.

### Workaround

The older `api/v1/users/web_profile_info/?username=<name>` endpoint still works reliably if you send the headers the current web client sends. The JSON it returns is shape-compatible with `Profile.__init__`, so you can just wrap the user dict.

Minimal example that works on 4.15.1 today:

```python
import instaloader

HEADERS = {
    "X-IG-App-ID": "936619743392459",
    "X-ASBD-ID": "198387",
    "User-Agent": (
        "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) "
        "AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.0 Safari/605.1.15"
    ),
}

def fetch_profile(L: instaloader.Instaloader, username: str) -> instaloader.Profile:
    try:
        return instaloader.Profile.from_username(L.context, username)
    except instaloader.exceptions.ProfileNotExistsException:
        pass
    url = f"https://www.instagram.com/api/v1/users/web_profile_info/?username={username}"
    headers = {**HEADERS, "Referer": f"https://www.instagram.com/{username}/"}
    resp = L.context._session.get(url, headers=headers, timeout=L.context.request_timeout)
    if resp.status_code == 404:
        raise instaloader.exceptions.ProfileNotExistsException(f"Profile {username} does not exist.")
    resp.raise_for_status()
    user = (resp.json().get("data") or {}).get("user")
    if not user:
        raise instaloader.exceptions.ProfileNotExistsException(f"Profile {username} does not exist.")
    return instaloader.Profile(L.context, user)
```

Tested with a logged-in session across a handful of profiles (mine, a mid-size creator, `natgeo`). Returns the expected `mediacount` and `full_name` in all cases.

### Proposal

Either:
1. Switch `Profile.from_username` to use `web_profile_info` as its primary path (it's what the IG web app itself uses), or
2. Keep the current `doc_id` call as the fast path and fall back to `web_profile_info` when `data` is null, the errors array is non-empty, or the response is 4xx.

Option 2 is the safer landing. If Meta fixes the `doc_id` endpoint later, nothing regresses.

Happy to open a PR if a maintainer can confirm which direction is preferred. Not married to the header set above either. It's just what the web client sends today and what works for me.

### Version

`instaloader==4.15.1`, Python 3.12, macOS.

### LINKED ISSUE #2695: 400 Bad Request on graphql/query: invalid request, execution error, severity: CRITICAL
### Update: This issue has been successfully solved https://github.com/instaloader/instaloader/issues/2695#issuecomment-4495553287

---
**Describe the bug**
cannot download profile from username
ig might changed something on graphql/query

**To Reproduce**
instaloader --user-agent "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:139.0) Gecko/20100101 Firefox/139.0" --max-connection-attempts=10 --stories --no-metadata-json --abort-on=302,400,401 XXXXXXXXX --fast-update --login=myaccount

Loaded session from C:\Users\XXXXXXX\AppData\Local\Instaloader\session-myaccount.
Logged in as myaccount.

Download aborted: Query to https://www.instagram.com/graphql/query responded with "400 Bad Request": {"message":"invalid request","errors":[{"message":"execution error","severity":"CRITICAL"}],"status":"fail"}.
Saved session to C:\Users\XXXXXXX\AppData\Local\Instaloader\session-myaccount.

**Instaloader version**
4.15.1

**Additional context**
:stories command (download stories from following list) still works fine

alimony:
## Fix profile resolution after Instagram retired its GraphQL profile queries

Fixes #2695

### Problem

Instagram stopped serving the GraphQL endpoints Instaloader relies on for profile resolution, so every profile download now fails before any media is fetched. Three separate queries are affected:

| Method | Old query | Symptom |
| --- | --- | --- |
| `Profile.from_username` | `fbsearch/non_profiled_serp` (`doc_id 26347858941511777`) | `400 Bad Request` on `api/graphql` |
| `Profile.from_id` | `graphql_query 7c16654f22c819fb63d1183034a5162f` | `400 Bad Request` on `graphql/query` |
| `Profile._obtain_metadata` | `doc_id 25980296051578533` | `400 Bad Request` on `graphql/query` |

The existing community workaround ([#2696](https://github.com/instaloader/instaloader/pull/2696)) only updates the `_obtain_metadata` `doc_id`. That fixes the *logged-in* metadata fetch but leaves `from_username` and `from_id` broken, and does nothing for anonymous use – which is why several people in #2695 reported it "works logged in, but profile-not-found when logged out."

### Fix

Resolve profiles through Instagram's web API endpoints, which still work **both anonymously and when logged in** and return the profile node in the legacy format the rest of `Profile` already expects:

- **New `InstaloaderContext.web_profile_info(username)`** → `api/v1/users/web_profile_info/`. Returns the complete profile node, including the first page of timeline posts.
- **New `InstaloaderContext.web_user_info(user_id)`** → `api/v1/users/{id}/info/`. Resolves the current username for a given user id.

Both require the `www.instagram.com` web app id (`x-ig-app-id: 936619743392459`); the default app id is rejected with HTTP 400.

Wiring:

- `from_username` now calls `web_profile_info` and marks the node as fully populated, so no redundant follow-up query is made.
- `from_id` resolves the id → current username via `web_user_info`, then delegates to `from_username` (so renamed accounts are still handled).
- `_obtain_metadata` uses `web_profile_info` when anonymous. The logged-in branch keeps the `doc_id` query but adopts the values from #2696 (`doc_id 27937681195819736` plus `__relay_internal__pv__PolarisWebSchoolsEnabledrelayprovider=False` and `enable_integrity_filters=True`), so it still serves as a fallback for partially constructed profiles. Credit to @VupVup / #2696 for those values.

### Testing

Verified against a real account both ways:

- **Anonymous:** `from_username`, full metadata property access (no extra request), `get_posts` pagination across the first page boundary, `from_id`, and `ProfileNotExistsException` for a nonexistent username and id.
- **Logged in** (session imported from a browser): `from_username`, the logged-in `_obtain_metadata` `doc_id` branch, `from_id`, `get_posts` (iphone/xdt path), and a CLI profile download that fetched media successfully.

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -191,7 +191,8 @@ class InstaloaderContext:
                   'Referer': 'https://www.instagram.com/',
                   'User-Agent': self.user_agent,
                   'X-Instagram-AJAX': '1',
-                  'X-Requested-With': 'XMLHttpRequest'}
+                  'X-Requested-With': 'XMLHttpRequest',
+                  'x-ig-app-id': '936619743392459'}
         if empty_session_only:
             del header['Host']
             del header['Origin']
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -912,11 +912,20 @@ class Profile:
         :param username: Username
         :raises: :class:`ProfileNotExistsException`
         """
-        data = context.doc_id_graphql_query("26347858941511777", {"hasQuery": True, "query": username})["data"]
-        if data:
-            for user in data["xdt_api__v1__fbsearch__non_profiled_serp"]["users"]:
-                if user["username"].lower() == username.lower():
-                    return cls(context, user)
+        # Resolve the profile through the web_profile_info endpoint, which works both
+        # anonymously and when logged in and returns the complete profile node
+        # (including the first page of posts). The GraphQL fbsearch query previously
+        # used here started responding with HTTP 400.
+        try:
+            data = context.get_json(
+                "api/v1/users/web_profile_info/", params={"username": username.lower()}
+            ).get("data")
+        except QueryReturnedNotFoundException:
+            data = None
+        if data and data.get("user"):
+            profile = cls(context, data["user"])
+            profile._has_full_metadata = True
+            return profile
 
         raise ProfileNotExistsException("Profile {} does not exist.".format(username))
 
@@ -931,15 +940,16 @@ class Profile:
         """
         if profile_id in context.profile_id_cache:
             return context.profile_id_cache[profile_id]
-        data = context.graphql_query('7c16654f22c819fb63d1183034a5162f',
-                                     {'user_id': str(profile_id),
-                                      'include_chaining': False,
-                                      'include_reel': True,
-                                      'include_suggested_users': False,
-                                      'include_logged_out_extras': False,
-                                      'include_highlight_reels': False})['data']['user']
-        if data:
-            profile = cls(context, data['reel']['owner'])
+        # Resolve the current username from the user id, then load the full profile.
+        # The GraphQL user query previously used here started responding with HTTP 400.
+        try:
+            user = context.get_json(
+                "api/v1/users/{0}/info/".format(profile_id), params={}
+            ).get('user')
+        except QueryReturnedNotFoundException:
+            user = None
+        if user and user.get('username'):
+            profile = cls.from_username(context, user['username'])
         else:
             raise ProfileNotExistsException("No profile found, the user may have blocked you (ID: " +
                                             str(profile_id) + ").")
@@ -985,6 +995,24 @@ class Profile:
     def _obtain_metadata(self):
         try:
             if not self._has_full_metadata:
+                if not self._context.is_logged_in:
+                    # Anonymous access: the web_profile_info endpoint returns the full
+                    # node in the legacy format and still works, unlike the GraphQL
+                    # profile query.
+                    try:
+                        data = self._context.get_json(
+                            "api/v1/users/web_profile_info/",
+                            params={"username": self.username},
+                        ).get('data')
+                    except QueryReturnedNotFoundException as err:
+                        raise ProfileNotExistsException(
+                            'Profile {} does not exist.'.format(self.username)) from err
+                    user_data = data.get('user') if data else None
+                    if user_data is None:
+                        raise ProfileNotExistsException('Profile {} does not exist.'.format(self.username))
+                    self._node = user_data
+                    self._has_full_metadata = True
+                    return
                 user_id = self._node.get('id') or self._node.get('pk')
                 variables = {
                     "id": str(user_id),


KoalaBear84:
I did the pip command from https://github.com/instaloader/instaloader/issues/2695#issuecomment-4721229688

I'm getting these errors. Strange, I've only did this, not used the app or website, so I don't extra HTTP 429.

```
instaloader --fast-update --login thesessionname instausername

Loaded session from C:\Users\Username\AppData\Local\Instaloader\session-thesessionname.
Logged in as thesessionname.
JSON Query to api/v1/users/web_profile_info/: 429 Too Many Requests when accessing https://www.instagram.com/api/v1/users/web_profile_info/?username=instausername [retrying; skip with ^C]
   d6f4427fbe92d846298cf93df0b937d3:    1    1    1    1    1    1
 *                            other:    1    1    1    1    1    1
is running.
The request will be retried in 666 seconds, at 09:58.
JSON Query to api/v1/users/web_profile_info/: 429 Too Many Requests when accessing https://www.instagram.com/api/v1/users/web_profile_info/?username=instausername [retrying; skip with ^C]
   d6f4427fbe92d846298cf93df0b937d3:    0    0    1    1    1    1
 *                            other:    1    1    2    2    2    2
is running.
The request will be retried in 666 seconds, at 10:09.
instausername: JSON Query to api/v1/users/web_profile_info/: 429 Too Many Requests when accessing https://www.instagram.com/api/v1/users/web_profile_info/?username=instausername
Saved session to C:\Users\Username\AppData\Local\Instaloader\session-thesessionname.

Errors or warnings occurred:
instausername: JSON Query to api/v1/users/web_profile_info/: 429 Too Many Requests when accessing https://www.instagram.com/api/v1/users/web_profile_info/?username=instausername
```

instaloader/instaloadercontext.py
```diff
@@ -562,6 +562,56 @@ def doc_id_graphql_query(self, doc_id: str, variables: Dict[str, Any],
...
+        with copy_session(self._session, self.request_timeout) as tmpsession:
+            tmpsession.headers.update(self._default_http_header(empty_session_only=True))
+            del tmpsession.headers['Connection']
+            del tmpsession.headers['Content-Length']
+            tmpsession.headers['x-ig-app-id'] = '936619743392459'
+            tmpsession.headers['accept'] = '*/*'
+            return self.get_json('api/v1/users/{0}/info/'.format(user_id),
+                                 params={},
+                                 session=tmpsession)
+
```
aandergr: I think web_profile_info() and web_user_info() are not actually needed; we can add the x-ig-app-id to the _default_http_header() definition and then call get_json() directly, i.e.

```diff
diff --git a/instaloader/instaloadercontext.py b/instaloader/instaloadercontext.py
index ddb9844..becb8c8 100644
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -191,7 +191,8 @@ class InstaloaderContext:
                   'Referer': 'https://www.instagram.com/',
                   'User-Agent': self.user_agent,
                   'X-Instagram-AJAX': '1',
-                  'X-Requested-With': 'XMLHttpRequest'}
+                  'X-Requested-With': 'XMLHttpRequest',
+                  'x-ig-app-id': '936619743392459'}
         if empty_session_only:
             del header['Host']
             del header['Origin']
diff --git a/instaloader/structures.py b/instaloader/structures.py
index 93ee08d..d099ea1 100644
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -915,7 +915,9 @@ class Profile:
         # (including the first page of posts). The GraphQL fbsearch query previously
         # used here started responding with HTTP 400.
         try:
-            data = context.web_profile_info(username.lower()).get("data")
+            data = context.get_json(
+                "api/v1/users/web_profile_info/", params={"username": username.lower()}
+            ).get("data")
         except QueryReturnedNotFoundException:
             data = None
         if data and data.get("user"):
@@ -939,7 +941,9 @@ class Profile:
         # Resolve the current username from the user id, then load the full profile.
         # The GraphQL user query previously used here started responding with HTTP 400.
         try:
-            user = context.web_user_info(profile_id).get('user')
+            user = context.get_json(
+                "api/v1/users/{0}/info/".format(profile_id), params={}
+            ).get("user")
         except QueryReturnedNotFoundException:
             user = None
         if user and user.get('username'):
@@ -994,7 +998,10 @@ class Profile:
                     # node in the legacy format and still works, unlike the GraphQL
                     # profile query.
                     try:
-                        data = self._context.web_profile_info(self.username).get('data')
+                        data = self._context.get_json(
+                            "api/v1/users/web_profile_info/",
+                            params={"username": self.username},
+                        ).get("data")
                     except QueryReturnedNotFoundException as err:
                         raise ProfileNotExistsException(
                             'Profile {} does not exist.'.format(self.username)) from err
```

(and then remove the newly added web_profile_info() and web_user_info())

I tested that and it seems to work.

instaloader/structures.py
```diff
@@ -983,15 +989,32 @@ def _asdict(self):
...
                     "id": str(user_id),
                     "render_surface": "PROFILE",
                     "__relay_internal__pv__PolarisCannesGuardianExperienceEnabledrelayprovider": True,
                     "__relay_internal__pv__PolarisCASB976ProfileEnabledrelayprovider": False,
                     "__relay_internal__pv__PolarisRepostsConsumptionEnabledrelayprovider": False,
+                    "__relay_internal__pv__PolarisWebSchoolsEnabledrelayprovider": False,
+                    "enable_integrity_filters": True,
                 }
-                data = self._context.doc_id_graphql_query('25980296051578533', variables)
+                data = self._context.doc_id_graphql_query('27937681195819736', variables)
```
aandergr, line 1026: This change for the logged-in Profile._obtain_metadata() seems unnecessary now after rebasing, since this was addressed in b79a0a1e8b08adf9fd03376f7ef532ca101d83b3.

aandergr (commented):
Thanks for this PR! It does fix the Profile access for non-logged-in sessions.

I have two small remarks on it, once they're addressed, we can merge it.

masioware (approved):

alimony:
@KoalaBear84 @aandergr Updated and simplified.

aandergr:
Thanks again for sharing this fix! We merged it and it will be included in the next release.

Issue #2698, opened 2026-06-10
Instaloader not worked error 400. Please help!!!!


MrAleksandrN:
Defaulting to user installation because normal site-packages is not writeable
Requirement already satisfied: instaloader in C:\Users\Aleksandr\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\site-packages (4.15.1)
Requirement already satisfied: requests>=2.25 in C:\Users\Aleksandr\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\site-packages (from instaloader) (2.32.5)
Requirement already satisfied: charset_normalizer<4,>=2 in C:\Users\Aleksandr\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\site-packages (from requests>=2.25->instaloader) (3.4.3)
Requirement already satisfied: idna<4,>=2.5 in C:\Users\Aleksandr\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\site-packages (from requests>=2.25->instaloader) (3.10)
Requirement already satisfied: urllib3<3,>=1.21.1 in C:\Users\Aleksandr\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\site-packages (from requests>=2.25->instaloader) (2.5.0)
Requirement already satisfied: certifi>=2017.4.17 in C:\Users\Aleksandr\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\site-packages (from requests>=2.25->instaloader) (2025.11.12)
Using latest stamps from O:\фото\secret\Session-mraleksandr\latest-stamps.ini.
Cookies loaded successfully from firefox
murger95 has been successfully logged in.
Next time use --login=murger95 to reuse the same session.
JSON Query to graphql/query: 403 Forbidden when accessing https://www.instagram.com/graphql/query [retrying; skip with ^C]
JSON Query to graphql/query: 403 Forbidden when accessing https://www.instagram.com/graphql/query [retrying; skip with ^C]
[...]

PR #2703, opened 2026-06-24
Add web_profile_info fallback for Profile metadata; report real HTTP status on non-JSON errors

mholzinger:
## Motivation

Two related failures when downloading profiles while logged in:

1. **Profile metadata fetch returns `400 Bad Request – "fail" status, message "invalid request"`** (#2695, #2698). The `doc_id 25980296051578533` (`PolarisProfilePageContentQuery`) used by `Profile._obtain_metadata` is now rejected by Instagram, so every profile download fails right after `Stored ID …`.

2. **Failed CDN/media downloads report a misleading `Expecting value: line 1 column 1 (char 0)` instead of the real HTTP status.** This is the masking behind #2682, where a `403 Forbidden` surfaces as an incorrect *"Profile does not exist"*. When `simplejson` is installed, `requests`' `resp.json()` raises `requests.exceptions.JSONDecodeError`, which is **not** a subclass of the stdlib `json.decoder.JSONDecodeError`. `_response_error()` only suppresses the latter, so a non-JSON error body escapes `_response_error`, propagates out of `get_raw`, and is re-reported by the `_retry_on_connection_error` decorator as the cryptic JSON message — hiding the true status code.

## Changes

1. **`Profile._obtain_metadata`** – fetch metadata via the `web_profile_info` endpoint (which returns the node already in the legacy format, so no normalization is needed) and fall back to the existing `doc_id` query otherwise. Having two independent sources makes metadata fetching resilient to Instagram rotating either endpoint. (See #2688, which proposed exactly this fallback.)

2. **`_response_error`** – suppress the common `ValueError` ancestor so both `json.decoder.JSONDecodeError` and the `simplejson`-backed `requests.exceptions.JSONDecodeError` are handled, and the actual HTTP status is reported.

## Relationship to #2701 / #2696

The metadata fix (commit 1) overlaps with **#2701**, which takes the same web-API approach and additionally migrates `from_username` and `from_id` — a broader fix. I arrived at the same approach independently, which at least corroborates the diagnosis. **If #2701 is preferred, commit 1 here can be dropped.**

**Commit 2 (the `_response_error` fix) is independent and not addressed by #2701, #2696, or any other open PR.** It resolves the masked-error behavior reported in #2682 regardless of which metadata fix lands, so it stands on its own.

## Completeness

Tested against a live logged-in account: a full 374-post profile download completes with **no** `invalid request` errors, and transient CDN failures now report their real status (`403 Forbidden when accessing …`) instead of the JSON-decode message. Ready to merge — happy to split commit 2 into its own PR if you'd rather take the metadata fix via #2701.

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -370,7 +370,10 @@ class InstaloaderContext:
     @staticmethod
     def _response_error(resp: requests.Response) -> str:
         extra_from_json: Optional[str] = None
-        with suppress(json.decoder.JSONDecodeError):
+        # resp.json() may raise either the stdlib json.JSONDecodeError or, when simplejson is
+        # installed, requests' own JSONDecodeError, which is not a subclass of the former. Suppress
+        # their common ValueError ancestor so a non-JSON error body does not mask the real status.
+        with suppress(ValueError):
             resp_json = resp.json()
             if "status" in resp_json:
                 extra_from_json = (
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -1,6 +1,7 @@
 import json
 import lzma
 import re
+import time
 from base64 import b64decode, b64encode
 from contextlib import suppress
 from datetime import datetime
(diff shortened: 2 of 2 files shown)


aandergr:
The profile metadata access had already been fixed in the last release for logged-in instances; the anonymous case is not fixed by this PR either.

Commit 2 does not effectively handle the simplejson decoding error exception, since it only changes the exception clause for one of multiple cases where this error is catched. Also, catching ValueError is too broad: with catching JSONDecodeError the code is explicit about what kind of error is catched there, making the code more readable. We generally discourage having simplejson installed in the same venv as Instaloader.

PR #2713, opened 2026-07-12
chore: add monthly Dependabot and harden checkout

andife:
## What this changes
This PR adds Dependabot configuration and hardens GitHub Actions checkout behavior.

## Dependabot
- adds .github/dependabot.yml
- enables grouped dependency updates
- schedules updates monthly
- configures a 7-day cooldown to avoid rapid follow-up bumps
- includes GitHub Actions updates

Configured update scopes:
- pip in /
- pip in /docs
- github-actions in /

## GitHub Actions hardening
Sets persist-credentials: false for ctions/checkout@v4 in:
- .github/workflows/windows_exe.yml
- .github/workflows/packages.yml
- .github/workflows/lint.yml
- .github/workflows/documentation.yml

## Why
- grouped updates reduce PR noise
- monthly cadence is lower maintenance
- cooldown helps avoid update churn
- disabling persisted checkout credentials follows least-privilege CI practices

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,10 +6,6 @@ updates:
       interval: monthly
     cooldown:
       default-days: 7
-    groups:
-      python-dependencies:
-        patterns:
-          - "*"
 
   - package-ecosystem: pip
     directory: "/docs"
@@ -17,10 +13,6 @@ updates:
       interval: monthly
     cooldown:
       default-days: 7
-    groups:
-      docs-python-dependencies:
-        patterns:
-          - "*"
 
   - package-ecosystem: github-actions
     directory: "/"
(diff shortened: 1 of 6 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
...
+    cooldown:
+      default-days: 7
+
+  - package-ecosystem: pip
+    directory: "/docs"
+    schedule:
+      interval: monthly
+    cooldown:
+      default-days: 7
+
```
aandergr, line 16: This needs to be removed; the `/docs` dependencies are actually included in the top-level Pipfile.

I noticed that there is a `/docs/requirements.txt`, however, this is an accidental leftover that we need to remove (and update the `/docs/README.md` accordingly).

aandergr (commented):
Thanks for addressing this!

I also like the Dependabot configuration that you chose, the values seem fine for me to start with. I have a minor comment that the /docs setting for Dependabot is superfluous.

PR #2706, opened 2026-06-29
Fix Post metadata fetching broken by Instagram deprecating doc_id

overskye:
Fixes #2704.

## Problem

Around June, 2026, Instagram deprecated the GraphQL `doc_id` with `8845758582119845` (`xdt_shortcode_media`), causing `_obtain_metadata()` to receive `null` for the `xdt_shortcode_media` field and crash with:

    TypeError: 'NoneType' object is not subscriptable

## Changes

**`instaloader/structures.py` — `_obtain_metadata()`**

Migrates to the new working endpoint:
- doc_id: `27128499623469141` (`PolarisPostRootQuery`)
- response field: `xdt_api__v1__media__shortcode__web_info.items[0]`

The new endpoint returns data in v1/iPhone API format (`media_type` integer, `code`, `pk`, `taken_at`, `image_versions2`, etc.) instead of the legacy GraphQL format. The method now converts this to the legacy field names that the rest of the codebase expects (`__typename`, `shortcode`, `id`, `taken_at_timestamp`, `display_url`, `edge_media_to_caption`, etc.).

**`instaloader/instaloadercontext.py` — `doc_id_graphql_query()`**

The new endpoint requires an `X-CSRFToken` request header (cookie alone is rejected with 403). The method now fetches a `csrftoken` from Instagram if the session doesn't have one yet (anonymous use), then passes it as the `x-csrftoken` header.

## Note for issue subscribers

Hope this solution helps you 🙏

PR #2717, opened 2026-07-17
Degrade gracefully when a single media's info endpoint answers feedback_required

mholzinger:
## Problem

Instagram returns `400 Bad Request` with `feedback_required` for `api/v1/media/<id>/info/` on certain media (observed repeatedly with reels carrying licensed audio), while the same call succeeds for neighboring media in the same session moments later. Because `feedback_required` raises `AbortDownloadException` (which intentionally bypasses the per-post error catchers), one such media kills the entire download pass — and since it is deterministic for that media, every retry aborts on the same first item. The profile can never be downloaded past it:

```
[  1] [Wow — a snippet of our gig at…]
Download aborted: 400 Bad Request - "fail" status, message "feedback_required" when accessing https://i.instagram.com/api/v1/media/3711864895911901816/info/.
```

The account itself is fine: downloading a different profile immediately afterwards works.

## Fix

The media-info lookup in `Post._iphone_struct` only *upgrades* image/video quality — both call sites already degrade gracefully on other errors. Catch `AbortDownloadException` there, log a warning, and fall back to standard quality for that one media (cached, so it is not re-requested).

Account-level protection is preserved: if several distinct media (5) hit `feedback_required` in one run, the pattern does look like an account flag and the exception is re-raised as before.

## Testing

- A restricted reel that previously aborted the whole `--reels` pass now downloads in standard quality with a warning, and the pass continues through the rest of the backlog.
- Simulated repeated failures across 5 distinct media re-raise `AbortDownloadException`.
- Media where the info endpoint succeeds are unaffected.

### DIFF
--- a/instaloader/structures.py
+++ b/instaloader/structures.py
@@ -354,9 +354,27 @@ class Post:
             raise IPhoneSupportDisabledException("iPhone support is disabled.")
         if not self._context.is_logged_in:
             raise LoginRequiredException("Login required to access iPhone media info endpoint.")
-        if not self._iphone_struct_:
-            data = self._context.get_iphone_json(path='api/v1/media/{}/info/'.format(self.mediaid), params={})
-            self._iphone_struct_ = data['items'][0]
+        if self._iphone_struct_ is None:
+            try:
+                data = self._context.get_iphone_json(
+                    path='api/v1/media/{}/info/'.format(self.mediaid), params={})
+                self._iphone_struct_ = data['items'][0]
+            except AbortDownloadException as err:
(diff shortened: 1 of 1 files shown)


aandergr:
> ```
> [ 1] [Wow — a snippet of our gig at…]
> Download aborted: 400 Bad Request - "fail" status, message "feedback_required" when accessing https://i.instagram.com/api/v1/media/3711864895911901816/info/.
> ```

We cannot reproduce this. Logged-in downloading this Post by shortcode, or by its media id works with the branch discussed in #2706, a candidate for soon being merged to fix the Post metadata query.

Is this an observation that can be otherwise reproduced?

PR #2718, opened 2026-07-18
Reduce get_reels() request volume by reusing the clips payload (on top of #2706)

theojcw-source:
## Update

This PR originally added a mobile-metadata fallback for Reels. As @aandergr noted, that became obsolete once #2706 landed a proper fix for the underlying issue (Instagram deprecating the `xdt_shortcode_media` doc_id). Repurposing this PR rather than opening a new one to keep the discussion in one place.

The branch is now rebased on #2706 (including the `media_type` type-guard fix suggested in that thread) and adds one more improvement on top:

## Problem

`get_reels()` builds each `Post` via `Post.from_shortcode()`, which throws away the media payload the clips connection already returned and issues one additional metadata request per Reel. For an account with 80 Reels that's 80 extra requests just for the reels tab, on top of pagination — which is a meaningful contributor to rate limiting/checkpoints when scraping Reels at any volume.

## Change

`Profile._reel_from_node()` now builds the `Post` directly from the clips payload via `Post._normalize_post_data()` (the same normalizer #2706 introduced for `_obtain_metadata()`), instead of re-fetching. `Post._field()` still lazily falls back to a full fetch for any field this payload doesn't cover, and `_reel_from_node()` falls back to the old `from_shortcode()` behavior entirely if the payload is missing something `_normalize_post_data()` needs (e.g. an unrecognized `media_type`).

## Validation

- `mypy` / `pylint`: clean
- New unit test (`TestReelsFromClipsPayload`) covering both the direct-build path and the fallback, same style as the test from this PR's previous version
- Live-tested against 3 real public Instagram accounts via an authenticated session (80, 65, and 80 reels respectively — 225 Reels total), 0 rate-limit errors, 100% with caption/timestamp/views/likes/comments populated, including from GitHub-Actions-hosted runners (previously a source of aggressive 429s when `get_reels()` made this many requests per profile)

## Note

This is stacked on #2706 — until that merges, this PR's diff shows both. Happy to rebase and trim to just the `get_reels()` commits once #2706 lands, or to fold review of both into one pass, whichever the maintainers prefer.

### DIFF
--- a/instaloader/instaloadercontext.py
+++ b/instaloader/instaloadercontext.py
@@ -540,6 +540,13 @@ class InstaloaderContext:
         :param referer: HTTP Referer, or None.
         :return: The server's response dictionary.
         """
+        csrf = next((c.value for c in self._session.cookies
+                     if c.name == 'csrftoken' and c.value), None)
+        if not csrf:
+            self._session.get('https://www.instagram.com/', timeout=self.request_timeout)
+            csrf = next((c.value for c in self._session.cookies
+                         if c.name == 'csrftoken' and c.value), '')
+
         with copy_session(self._session, self.request_timeout) as tmpsession:
             tmpsession.headers.update(self._default_http_header(empty_session_only=True))
             del tmpsession.headers['Connection']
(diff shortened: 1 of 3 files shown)


aandergr:
Hi! Thanks for this suggestion! However, this seems obsolete with #2706, which also seems to fix the reels downloading.

The unit test is cool, though!

Issue #2724, opened 2026-07-29
400 Bad Request when downloading posts


amchap:
When I want to download images associated with a public user profile (I'm the admin)  I get the following error 
```
posts = instaloader.Profile.from_username(L.context, PROFILE).get_posts()
File "/Library/Python/3.9/lib/python/site-packages/instaloader/structures.py", line 1007, in from_username
    data = context.get_json(
File "/Library/Python/3.9/lib/python/site-packages/instaloader/instaloadercontext.py", line 463, in get_json
    raise QueryReturnedBadRequestException(self._response_error(resp))
instaloader.exceptions.QueryReturnedBadRequestException: 400 Bad Request - "fail" status, message "Asset asset://laser.provider/ig_business_category_subvertical has been deleted. You cannot use this schema" when accessing https://www.instagram.com/api/v1/users/web_profile_info/?username=###

```
It looks like the IG API expects a different schema than what Instaloader is asking

I'm running the python script and the latest version of Instaloader, v4.15.3.

I thought it might be permissions, but I get the same error when I add the password function

vadimkantorov:
Also getting 400 on local machine:
```
instaloader.exceptions.QueryReturnedBadRequestException: 400 Bad Request - "fail" status, message "Asset asset://laser.provider/ig_business_category_subvertical has been deleted. You cannot use this schema" when accessing https://www.instagram.com/api/v1/users/web_profile_info/?username=russie.libertes
```

---

And getting 429 when running to scrape top 5 instagram post on GitHub CI as part of SSG site generation:

```
JSON Query to api/v1/users/web_profile_info/: 429 Too Many Requests when accessing https://www.instagram.com/api/v1/users/web_profile_info/?username=russie.libertes [retrying; skip with ^C]
Number of requests within last 10/11/20/22/30/60 minutes grouped by type:
 *                            other:    1    1    1    1    1    1
 *

```

Are there any suggestions on configuring Instaloader for running at GitHub CI machines? I'm interested to scrape recent 5 posts only

roehrs:
Confirming this on the current PyPI release, **instaloader 4.15.3** (macOS, Python 3.9.6).

`Profile.from_username()` fails for every public profile, authenticated (valid saved session) or not:

```
QueryReturnedBadRequestException: 400 Bad Request - "fail" status,
message "Asset asset://laser.provider/ig_business_category_subvertical has been deleted.
You cannot use this schema"
[...]

PR #2730, opened 2026-08-15
Send Instagram API requests over HTTP/2

### LINKED ISSUE #2726: 429 Too Many Requests on web_profile_info
**Describe the bug**
`Profile.from_username()` returns a 429 Too Many Requests on the very first request to `api/v1/users/web_profile_info/`, with no prior requests made in the session. This is reproducible:
- On a completely fresh, valid session cookie (not previously rate-limited)
- Across three distinct networks: mobile data, home wifi, and a VPN
- When making the request fully anonymously (no session/login at all), against a public account (`@instagram`)
- On the latest released version (4.15.3)

This suggests either a broad throttle on the `web_profile_info` endpoint independent of account/IP/session, or a change in Instagram's expected request shape/headers that current Instaloader isn't matching.

**To Reproduce**
Run the following minimal script with no session attached:
```python
import instaloader

L = instaloader.Instaloader(
    download_pictures=False,
    download_videos=False,
    download_comments=False,
    download_geotags=False,
    save_metadata=False,
    compress_json=False,
    max_connection_attempts=1,
    request_timeout=30,
)

profile = instaloader.Profile.from_username(L.context, "instagram")
print(profile.followers)
```
2. Observe 429 on the first and only request made in the session.
3. Repeated on a separate run with a valid `sessionid` cookie attached (own account, freshly obtained), same result.
4. Repeated across mobile data, home wifi, and VPN; same result on all three.

**Expected behavior**
A single, first request in a fresh session (with or without authentication) should not be rate-limited expected either a successful response or, if Instagram is now blocking anonymous/first-touch requests to this endpoint outright, a clearer error than a generic 429.

```
JSON Query to api/v1/users/web_profile_info/: 429 Too Many Requests when accessing https://www.instagram.com/api/v1/users/web_profile_info/?username=instagram [retrying; skip with ^C]
 *                            other:    1    1    1    1    1    1
is running.
```
Note: this was the first and only request made not a result of request volume.

**Instaloader version**
4.15.3

**Additional context**
- Running inside Termux on Android 14 (aarch64), Python 3.14.6.
- Isolated the cause methodically: ruled out IP/network (three different networks tested), ruled out session/account (reproduced fully anonymously), ruled out outdated client (confirmed on latest version).
- Tried to inspect the underlying HTTP response for rate-limit headers (e.g. `Retry-After`) via the `ConnectionException`, but `response`/`headers` were not reachable from the exception object (came back `null`) so no header-level detail available from the client side, in case that's itself informative about how the exception is structured.

e3rd:
# Send Instagram API requests over HTTP/2

Fixes #2726.

## The cause

Instagram has started answering HTTP/1.1 requests to its web API with `429 Too
Many Requests` and an empty body. The identical request over HTTP/2 succeeds.
This is not a rate limit: it triggers on the very first request of a fresh
session, and is independent of account, session and IP address.

The rollout appears to be partial — many users are evidently still unaffected,
which is presumably why there is only one report of it so far — so this may not
reproduce everywhere yet. Both machines I tested from are in the same country,
so I cannot say how it behaves in other regions.

Reproduced on two unrelated networks and machines, with the same client, same
headers, same second:

```console
$ curl -s -o /dev/null -w '%{http_code} http/%{http_version}\n' --http2 \
    -H 'X-IG-App-ID: 936619743392459' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36' \
    'https://www.instagram.com/api/v1/users/web_profile_info/?username=instagram'
200 http/2

$ curl -s -o /dev/null -w '%{http_code} http/%{http_version}\n' --http1.1 \
    -H 'X-IG-App-ID: 936619743392459' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36' \
    'https://www.instagram.com/api/v1/users/web_profile_info/?username=instagram'
429 http/1.1
```

The TLS fingerprint is not the discriminator: HTTPX with its stock (non-browser)
fingerprint gets `200` over HTTP/2 and `429` over HTTP/1.1. Only the protocol
version matters.

Requests is built on urllib3, which speaks HTTP/1.1 only, so on an affected
network Instaloader cannot reach the API at all — which is what #2726 reports.
Browsers, whose behavior Instaloader mimics, do not use HTTP/1.1 here.

## The fix

A `requests.adapters.HTTPAdapter` subclass (`instaloader/http2.py`) that performs
the request with HTTPX and converts the reply back into a `requests.Response`.
It is mounted on `www.instagram.com` and `i.instagram.com` only.

This keeps the change small and leaves Instaloader's Requests-based code
untouched — cookie handling, redirects, rate controlling, session save/load and
the login flow all work exactly as before. Media downloads go to the CDN, which
is not affected by the 429 and stays on the default transport, so the streaming
download path is unchanged.

## Testing

Two tests are added in `TestHTTP2Transport`: one asserts the adapter is mounted
for the Instagram hosts but not for the CDN, the other is a direct regression
test for #2726 (`web_profile_info` must return `200`).

Verified manually against Instagram:

- anonymous `Profile.from_username()` on a public profile — the exact call from
  the issue
- logged-in access to a private profile, including GraphQL paging over posts
- media + metadata download of a post (CDN path)
- `save_session_to_file()`, which harvests `csrftoken` from `Set-Cookie` and so
  exercises the adapter's cookie conversion

`pylint instaloader` reports 10.00/10 and `mypy -m instaloader` is clean.

## Completeness

I consider this ready to be merged rather than a proof of concept, but two
decisions are yours:

- It adds `httpx[http2]` as a **required** dependency, since on an affected
  network the library cannot function without it. If you would rather not take
  one on, the adapter can be made optional, falling back to Requests when HTTPX
  is absent — happy to rework it.
- It is based on `master`, treating this as a bug fix. If you consider swapping
  the transport extensive enough to warrant an `upcoming/v4.X` cycle, I will
  gladly rebase.

No documentation change seemed necessary, as the transport is invisible to
users, but I am glad to add a note to the troubleshooting page.

---

Disclosure: this change was written with the help of Claude (Anthropic), which
is also recorded as a `Co-Authored-By` trailer on the commit. I have reviewed
and tested it myself.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -19,4 +19,5 @@ pefile = "*"          # Needed for pyinstaller on Windows
 
 [packages]
 requests = "*"
+httpx = {extras = ["http2"], version = "*"}
 browser-cookie3 = "*"
--- a/instaloader/http2.py
+++ b/instaloader/http2.py
@@ -90,14 +90,18 @@ class HTTP2Adapter(requests.adapters.HTTPAdapter):
         # Instagram's API hosts, whose replies are small JSON documents. Media
         # files are downloaded from the CDN, which the adapter is not mounted on.
         body = response.content
-        headers = urllib3.HTTPHeaderDict()
-        for name, value in response.headers.multi_items():
-            if name.lower() not in _SKIPPED_RESPONSE_HEADERS:
-                headers.add(name, value)
-        headers['Content-Length'] = str(len(body))
+        # Passing the headers as a list of pairs rather than building a
+        # urllib3.HTTPHeaderDict keeps duplicate headers (Set-Cookie) intact
+        # while working on both urllib3 1.x and 2.x, which export the class
+        # from different places.
+        headers = [(name, value) for name, value in response.headers.multi_items()
+                   if name.lower() not in _SKIPPED_RESPONSE_HEADERS]
+        headers.append(('Content-Length', str(len(body))))
         return urllib3.HTTPResponse(
             body=io.BytesIO(body),
-            headers=headers,
+            # urllib3 accepts an iterable of pairs here, but is annotated for
+            # mappings only, which cannot carry a repeated Set-Cookie.
+            headers=headers,  # type: ignore[arg-type]
             status=response.status_code,
             reason=response.reason_phrase,
             version=20 if response.http_version == 'HTTP/2' else 11,
--- a/instaloader/http2.py
+++ b/instaloader/http2.py
@@ -1,0 +1,110 @@
+"""HTTP/2 transport for Instagram requests.
+
+Instagram has started answering HTTP/1.1 requests to its web API with ``429 Too
+Many Requests`` and an empty body. It is not a rate limit: it hits the very first
+request of a fresh session, and neither the session, the account nor the IP
+address makes a difference. The same request over HTTP/2 succeeds, so Instaloader
+has to speak HTTP/2 -- which Requests, being built on urllib3, cannot do.
+
+This module bridges the gap with a :class:`requests.adapters.HTTPAdapter` that
+hands the request over to HTTPX (which does speak HTTP/2) and converts the reply
+back into a :class:`requests.Response`. Mounting the adapter keeps cookie
+handling, redirects and the rest of Instaloader's Requests-based code untouched.
+"""
+
+import http.client
+import io
+from typing import Optional
+
+import httpx
+import requests
+import requests.adapters
+import urllib3
+
+# Set by the transport itself; forwarding them to HTTPX would conflict with the
+# HTTP/2 pseudo-headers or with the body HTTPX re-encodes.
+_SKIPPED_REQUEST_HEADERS = frozenset({'host', 'connection', 'transfer-encoding', 'content-length'})
+
+# HTTPX transparently decompresses the body, so passing these on would make
+# urllib3 attempt to decode the payload a second time.
+_SKIPPED_RESPONSE_HEADERS = frozenset({'content-encoding', 'content-length', 'transfer-encoding'})
+
+
+class _OriginalResponse:
+    """The bit of :mod:`http.client` API that Requests needs to extract cookies."""
+
+    def __init__(self, headers: httpx.Headers):
+        self.msg = http.client.HTTPMessage()
+        for name, value in headers.multi_items():
+            self.msg.add_header(name, value)
+
+    def isclosed(self) -> bool:
+        return True
+
+    def close(self) -> None:
+        pass
+
+
+class HTTP2Adapter(requests.adapters.HTTPAdapter):
+    """Requests transport adapter that performs the request over HTTP/2.
+
+    .. versionadded:: 4.15.4"""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._client = httpx.Client(http2=True, follow_redirects=False, timeout=None)
+
+    def send(self, request, stream=False, timeout=None, verify=True, cert=None, proxies=None):
+        headers = {name: value for name, value in request.headers.items()
+                   if name.lower() not in _SKIPPED_REQUEST_HEADERS}
+        try:
+            response = self._client.request(
+                request.method,
+                request.url,
+                headers=headers,
+                content=request.body,
+                timeout=self._httpx_timeout(timeout),
+            )
+        except httpx.TimeoutException as err:
+            raise requests.exceptions.Timeout(err, request=request) from err
+        except httpx.HTTPError as err:
+            raise requests.exceptions.ConnectionError(err, request=request) from err
+        return self.build_response(request, self._to_urllib3_response(response))
+
+    def close(self) -> None:
+        self._client.close()
+        super().close()
+
+    @staticmethod
+    def _httpx_timeout(timeout) -> Optional[httpx.Timeout]:
+        if timeout is None:
+            return None
+        if isinstance(timeout, tuple):
+            connect, read = timeout
+            return httpx.Timeout(read, connect=connect)
+        return httpx.Timeout(timeout)
+
+    @staticmethod
+    def _to_urllib3_response(response: httpx.Response) -> urllib3.HTTPResponse:
+        # Reading the body eagerly is fine here: the adapter only serves
+        # Instagram's API hosts, whose replies are small JSON documents. Media
+        # files are downloaded from the CDN, which the adapter is not mounted on.
+        body = response.content
+        headers = urllib3.HTTPHeaderDict()
+        for name, value in response.headers.multi_items():
+            if name.lower() not in _SKIPPED_RESPONSE_HEADERS:
+                headers.add(name, value)
+        headers['Content-Length'] = str(len(body))
+        return urllib3.HTTPResponse(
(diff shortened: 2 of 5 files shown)


tomballgithub:
Both of your curl commands worked for me just now in the USA

davispuh:
I can confirm this issue.
When using latest instaloader `master` branch
```
JSON Query to api/v1/users/web_profile_info/: Expecting value: line 1 column 1 (char 0) [retrying; skip with ^C]
JSON Query to api/v1/users/web_profile_info/: Expecting value: line 1 column 1 (char 0) [retrying; skip with ^C]
```

```
$ curl -s -o /dev/null -w '%{http_code} http/%{http_version}\n' --http2 \
                                                           -H 'X-IG-App-ID: 936619743392459' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36' \
                                                           'https://www.instagram.com/api/v1/users/web_profile_info/?username=instagram'
200 http/2
```

```
$ curl -s -o /dev/null -w '%{http_code} http/%{http_version}\n' --http1.1 \
                                                       -H 'X-IG-App-ID: 936619743392459' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36' \
                                                       'https://www.instagram.com/api/v1/users/web_profile_info/?username=instagram'
429 http/1.1
```

But this PR doesn't work, fails with:
```py
  File "instaloader.py", line 6, in <module>
    main()
    ~~~~^^
  File "instaloader/__main__.py", line 581, in main
    exit_code = _main(loader,
                      args.profile,
    ...<15 lines>...
                      browser=args.load_cookies,
                      cookiefile=args.cookiefile)
  File "instaloader/__main__.py", line 237, in _main
    profile = Profile.from_username(instaloader.context, target[1:])
  File "instaloader/structures.py", line 1007, in from_username
    data = context.get_json(
           ~~~~~~~~~~~~~~~~^
        "api/v1/users/web_profile_info/", params={"username": username.lower()}
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ).get("data")
    ^
  File "instaloader/instaloadercontext.py", line 443, in get_json
    resp = sess.get('https://{0}/{1}'.format(host, path), params=params, allow_redirects=False)
  File "/usr/lib/python3.14/site-packages/requests/sessions.py", line 600, in get
    return self.request("GET", url, **kwargs)
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/requests/sessions.py", line 587, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python3.14/site-packages/requests/sessions.py", line 701, in send
    r = adapter.send(request, **kwargs)
  File "instaloader/http2.py", line 72, in send
    return self.build_response(request, self._to_urllib3_response(response))
                                        ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "instaloader/http2.py", line 93, in _to_urllib3_response
    headers = urllib3.HTTPHeaderDict()
              ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'urllib3' has no attribute 'HTTPHeaderDict'
```
(using urllib3 2.7.0)

This fixes it:
```diff
diff --git a/instaloader/http2.py b/instaloader/http2.py
index e54f13e..ae4f655 100644
--- a/instaloader/http2.py
+++ b/instaloader/http2.py
@@ -90,7 +90,7 @@ class HTTP2Adapter(requests.adapters.HTTPAdapter):
         # Instagram's API hosts, whose replies are small JSON documents. Media
         # files are downloaded from the CDN, which the adapter is not mounted on.
         body = response.content
-        headers = urllib3.HTTPHeaderDict()
+        headers = urllib3._collections.HTTPHeaderDict()
         for name, value in response.headers.multi_items():
             if name.lower() not in _SKIPPED_RESPONSE_HEADERS:
                 headers.add(name, value)
```

But even with this, this API `/api/v1/users/web_profile_info/` is unusable, because:
```
{"message":"Asset asset://laser.provider/ig_business_category_subvertical has been deleted. You cannot use this schema","status":"fail"}
```
Looks like #2724 #2731

e3rd:
Thanks, that is a real bug — and thank you for the diff, it pointed straight at the cause!

HTTPHeaderDict is only re-exported at urllib3's top level in urllib3 2.x; in 1.x it lives in urllib3._collections. Requests still allows urllib3>=1.21.1,<3, so I should never have relied on the top-level name. Your traceback confirms which one you are actually running: urllib3.HTTPResponse resolved fine on the line above, and that attribute exists in both — only HTTPHeaderDict is missing, which is the 1.x layout. So despite pip reporting 2.7.0, the interpreter that ran this picked up a 1.x install (a second environment, or system site-packages).

Rather than reaching into urllib3._collections — that is private API, and in 2.x the module is only a legacy alias — I pushed a fix that hands the headers to urllib3 as a list of (name, value) pairs and lets it build its own HTTPHeaderDict. That works identically on 1.x and 2.x and still preserves repeated Set-Cookie headers, which a plain mapping would collapse.

I also added an offline regression test that runs a fake HTTPX reply through the adapter and into a requests.Session, asserting the body and both cookies survive. The conversion path had no test that ran without network access, which is exactly why this slipped through. It fails on urllib3 1.26.20 without the fix and passes on both 1.26.20 and 2.7.0 with it.

The Asset asset://laser.provider/ig_business_category_subvertical has been deleted reply is unrelated to this PR — from here web_profile_info returns normal profile data over HTTP/2 right now, so that looks like the server-side issue tracked in #2724 / #2731, hitting a subset of requests. This PR only changes which protocol version the request goes out on; it does not touch what the API answers.

okkesd:
As of today, it works for me, thanks !!!

e3rd:
<3

KertLynx:
I confirm too. This MR worked for me.

van-sprundel:
You've added it to the pipfile, not setup.py, so this wouldn't fix the issue. add `httpx[http2]` to setup.py.

But the idea is correct. `python-requests` does not support HTTP2 yet (urllib, the underlying library, has been working [on support for 2 years now](https://urllib3.readthedocs.io/en/latest/changelog.html#id28)). I think it's a fine replacement until requests can support HTTP/2.

Injecting httpx in all instagram reqs sounds fine for now, though

e3rd:
> You've added it to the pipfile, not setup.py

I've put it into setup.py too: `requirements = ['requests>=2.25', 'httpx[http2]>=0.27']`

Or am I wrong? :o

apastel:
Sweet, thanks for this PR. Adding another "works for me!" comment, because it does ;)

marty90:
it works for me. I got rid of the 429 at every request

PR #2732, opened 2026-08-20
Make --count apply to profile downloads (posts and reels)

mholzinger:
## Problem

`download_profiles()` accepts `max_count` and honors it for the posts loop, but `__main__` never forwards `--count` to it — so `instaloader --count 100 <profile>` silently downloads the full history. Reels have no cap at all (`download_reels()` lacks the parameter).

## Changes

- Forward `--count` from the CLI to `download_profiles()` (both the logged-in and anonymous-retry paths)
- Add `max_count` to `download_reels()` and pass it through to `posts_download_loop` (which already implements the limit)
- Update the `--count` help text to mention profiles

## Motivation

A gentle first clone of a large account: fetch the most recent N posts now instead of the entire history in one burst (the kind of burst most likely to trip Instagram's automation heuristics), and backfill later. With `--latest-stamps`, the stamp records the newest post either way, so subsequent runs stay incremental.

## Testing

- `--count 3` against a profile with thousands of posts downloads exactly 3 and stops (verified live)
- Without `--count`, behavior is unchanged
- `--count` with `--reels` caps the reels pass the same way

### DIFF
--- a/instaloader/__main__.py
+++ b/instaloader/__main__.py
@@ -309,6 +309,7 @@ def _main(instaloader: Instaloader, targetlist: List[str],
             storyitem_filter,
             latest_stamps=latest_stamps,
             reels=download_reels,
+            max_count=max_count,
         )
         if anonymous_retry_profiles:
             instaloader.context.log("Downloading anonymously: {}"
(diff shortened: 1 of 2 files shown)


aandergr:
> This would unfortunately change Instaloader's behavior in a backward-incompatible manner, since an invocation with --count and a mix of hashtags/feed/location and a profile would then behave differently in that the max count now also affects the profile and not only hashtags/feed/location.
> 
> Also, it is quite trivial for the user to only download a certain count of posts of a profile, e.g. by calling download_post() for each post of an islice()'d profile.get_posts(). There is no need to include that in the CLI.

https://github.com/instaloader/instaloader/issues/1660#issuecomment-1200263791

PR #2740, opened 2026-09-10
Vendor browser_cookie3 to fix Firefox cookie path detection

agustux:
## Motivation

browser_cookie3, used by `--load-cookies`, has been unmaintained upstream
for over two years (last commit ~2 years ago), blocking fixes for
outdated Firefox profile-path detection, most recently affected by
Firefox 147's XDG Base Directory support (#2727).

Rather than continue patching an external dependency we can't get merged
upstream, this PR vendors browser_cookie3 directly into Instaloader so
these kinds of fixes can be maintained in-tree going forward.

## Changes

- Vendor browser_cookie3 (LGPL-3.0) into `instaloader/_vendor/browser_cookie3/`,
  with attribution and a `README.md` documenting the source commit and
  applied patches.
- Update `instaloader/__main__.py` to import from the vendored package.
- Remove `browser_cookie3` as an external dependency in `setup.py` and
  `Pipfile`; add its actual runtime dependencies (`lz4`, `pycryptodomex`,
  `dbus-python` on Linux) as a new optional extra, `browser-cookie-import`.
- Add `THIRD_PARTY_LICENSES.md` documenting the vendored LGPL-3.0 code,
  kept separate from Instaloader's MIT license.
- Update `docs/cli-options.rst` to reflect that browser_cookie3 is now
  bundled rather than a separate install.
- Update `docs/contributing.rst` with a note on the vendored-code policy
  and why it was adopted, plus a reminder to install optional extras
  before linting.
- Exclude `instaloader/_vendor/` from pylint via `.pylintrc`
  (`ignore-paths`), since vendored code isn't held to project style.

## Completeness

- Manually tested `--load-cookies firefox` against the new XDG (`~/.config/mozilla`) profile layouts, in
  a clean virtualenv install (`pip install -e .[browser-cookie-import]`).
- `pylint` and `mypy` pass cleanly on non-vendored code.
- Documentation updated everywhere browser_cookie3 was previously
  referenced (as far as I know)
- I think this is ready for review and open to feedback on the vendoring approach in case Alex has other thoughts.


agustux:
Sorry about all those commits trying to fix Python 3.9 tests

PR #2743, opened 2026-09-16
Add authenticated profile post iteration without profile metadata

okkesd:
This change was motivated by an issue in an application that monitors
recent posts from Instagram profiles. The `web_profile_info` request
returned HTTP 400 with `feedback_required`, while the authenticated
timeline query successfully retrieved posts using the username directly.

The application worked around this by constructing the timeline
iterator without resolving a Profile first. This PR exposes that
approach through a supported library method, so applications that only
need posts do not have to duplicate the query setup.

This addresses the unnecessary dependency on profile metadata; it does
not claim to resolve all `feedback_required` or HTTP 429 responses.
Existing rate-limit and authentication handling remain in place.

This PR adds `Instaloader.get_profile_posts(username)`, which returns
a resumable `NodeIterator[Post]` without requesting profile metadata.

The method requires authentication, normalizes and validates the
username, and propagates request errors. An empty timeline does not
establish whether the profile exists or is accessible.

The authenticated iterator construction is shared with
`Profile.get_posts()`. Its existing metadata lookup and anonymous
pagination behavior are retained.

Validation:
- 10 offline tests pass.
- Coverage includes authentication, username validation, pagination,
  empty/missing timeline responses, challenge propagation, freeze/thaw,
  and existing authenticated and anonymous Profile.get_posts() paths.
- Tests block real HTTP requests.

Test command:
`python3 -m unittest discover -s test -p 'test_profile_posts.py' -v`

This is a proposed public API addition; feedback on the method's name
and placement is welcome. The new method has not yet been live-tested
against Instagram. But the application used the same method and worked as expected


kkraiser:
Tested this PR by okkesd, commit ba8386ac57c20604e0abfd836420b314e4b1ad69) against live Instagram on Instaloader 4.15.3.

The new Instaloader.get_profile_posts(username) method successfully returned posts for multiple public profiles using an authenticated account. I also tested it with a private profile that the authenticated account legitimately follows, and posts from that profile were returned successfully.

I then passed the returned Post objects directly to Instaloader.download_post(). Downloads worked normally, including a mixed-media carousel containing multiple JPGs and an MP4, along with caption/JSON metadata.

I also tested using get_profile_posts() as a workaround for the currently failing profile-metadata path while retaining an existing --latest-stamps workflow. For one existing profile:

Existing post-timestamp: 2026-08-19 02:36:00-07:00
get_profile_posts() returned one newer post dated 2026-09-01 09:10:40-07:00
Iteration then reached the existing Aug. 19 checkpoint exactly
The newer post downloaded successfully
I updated the checkpoint with LatestStamps.set_last_post_timestamp()
A subsequent run correctly found zero posts newer than the new Sept. 1 checkpoint

One authentication observation: an older saved Instaloader session began returning HTTP 401 / "Please wait a few minutes before you try again." from both the login check and the new timeline query. Refreshing the session from current Firefox Instagram cookies immediately restored operation. This appeared to be a stale/restricted session rather than a problem with get_profile_posts().

So far, the new API has worked for public profiles, an authenticated private profile, normal Post downloading, mixed-media/carousel posts, and incremental downloading based on existing timestamps. I'm now testing it across a larger batch of existing and newly archived profiles.

